PR #25, opened 2014-07-01
Fixed the Linear stage when the output size is 1.

jonathantompson:
addmm wont work when the output dimension is 1 which causes Linear to fail.  At the cost of another lua branch conditional, this fixes the resultant error.  I've also added a test for this special case.  Soumith, please have a look over the code and make sure there's nothing stupid here (I wouldn't want to break Linear).

### DIFF
--- a/Linear.lua
+++ b/Linear.lua
@@ -40,8 +40,14 @@ function Linear:updateOutput(input)
       local nunit = self.bias:size(1)
 
       self.output:resize(nframe, nunit)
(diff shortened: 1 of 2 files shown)


szagoruyko:
@jonathantompson what was the issue? seems that it is no longer there, but there is a small numerical difference between the special cases

jonathantompson:
I can't comment on the numerical difference, but I believe it was fixed.  The issue was a hard error on the addmm call.

PR #118, opened 2014-12-06
Complex Convolutions (no CPU support, only CUDA).

jonathantompson:
I've taken the CUDA implementation of SpatialConvolutionMM and created a complex convolution module out of it.  There will be accompanying commits in cutorch and cunn to follow.

The cunn commit won't pass unit-tests (it will fail to compile) without the cutorch commit.  I'm not sure what to do about this...  If only torch was in a single repo like the good old days :-)


jonathantompson:
OK, I'm finally getting around to looking at this.  Adding CPU support is definitely possible but more work than I was anticipating.  The GPU version was a quick change because CudaBlas has support for cgemm, but the SpatialConvolutionMM code uses an addmm call which does not have complex support.  The alternative is to call addmm 4 times but this is super, super ugly.

Soumith, I vaguely remember you saying facebook might add complex number support to torch soon.  Is this still in the pipeline?  I'd rather not hack something together if there is a much more polished version coming soon.

soumith:
complex number support is in the pipeline. you can probably wait on it, and keep this in your jtools or jtorch repo, or cunnx

soumith:
let's close this. :-/

PR #127, opened 2015-01-02
Euclidean batch support

nicholas-leonard:
Includes unit tests. Optimized for the new non-contiguous copy kernels (https://github.com/torch/cutorch/pull/93).


nicholas-leonard:
The first commit shouldn't be there... Let me create a new PR.

PR #132, opened 2015-01-03
refactoring all the common container code into nn.Container

soumith:
nn.Container helps simplify the code for all of the container modules like nn.Parallel, nn.Concat, nn.Sequential etc.

This PR introduces nn.Container, and makes some of the existing container modules make it as the base class.
This change is fully backward-compatible in terms of serialization.
Reviewed by @andresy, authored by Keith Adams

### DIFF
--- a/Concat.lua
+++ b/Concat.lua
@@ -1,19 +1,9 @@
-local Concat, parent = torch.class('nn.Concat', 'nn.Module')
+local Concat, parent = torch.class('nn.Concat', 'nn.Container')
 
 function Concat:__init(dimension)
-   parent.__init(self)
-   self.modules = {}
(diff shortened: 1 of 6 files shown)


nicholas-leonard:
We need to add documentation for this new class.

soumith:
it is an abstract class, so i did not add any docs. hmmm, wonder what docs we should add, as "module" docs cover what container docs would be...

nicholas-leonard:
I can a small section to document nn.Container at the beginning of container.md. No biggy. 

Also I think this PR was great. I was myself thinking about doing this for a while. Thanks.

soumith:
i think that makes sense.

PR #128, opened 2015-01-02
Euclidean batch support

nicholas-leonard:
Includes unit tests. Optimized for the new non-contiguous copy kernels (https://github.com/torch/cutorch/pull/93).

PR #133, opened 2015-01-05
WeightedEuclidean optimizations

nicholas-leonard:
Some better unit tests and optimizations. Similar to #128 .

### DIFF
--- a/WeightedEuclidean.lua
+++ b/WeightedEuclidean.lua
@@ -3,20 +3,12 @@ local WeightedEuclidean, parent = torch.class('nn.WeightedEuclidean', 'nn.Module
 function WeightedEuclidean:__init(inputSize,outputSize)
    parent.__init(self)
 
-   self.templates = torch.Tensor(inputSize,outputSize)
-   self.gradTemplates = torch.Tensor(inputSize,outputSize)
+   self.weight = torch.Tensor(inputSize,outputSize)
+   self.gradWeight = torch.Tensor(inputSize,outputSize)
 
    -- each template (output dim) has its own diagonal covariance matrix
    self.diagCov = torch.Tensor(inputSize,outputSize)
    self.gradDiagCov = torch.Tensor(inputSize,outputSize)
-   
-   -- for compat with Torch's modules (it's bad we have to do that)
-   do
-      self.weight = self.templates
-      self.gradWeight = self.gradTemplates
-      self.bias = self.diagCov
-      self.gradBias = self.gradDiagCov
-   end
 
    self:reset()
 end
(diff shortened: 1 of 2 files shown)


nicholas-leonard:
These are ready to be merged.

WeightedEuclidean.lua
```diff
@@ -25,55 +17,63 @@ function WeightedEuclidean:reset(stdv)
...
    if input:dim() == 1 then
-      self.output:resize(self.templates:size(2))
-      for outIdx = 1,self.templates:size(2) do
-         self._temp:copy(input):add(-1,self.templates:select(2,outIdx))
-         self._temp:cmul(self._temp)
-         local diagCov = self.diagCov:select(2,outIdx)
-         self._temp:cmul(diagCov):cmul(diagCov)
-         self.output[outIdx] = math.sqrt(self._temp:sum())
-      end
+      self._input:reshape(input, inputSize, 1)
```
soumith: same comments as from the Euclidean PR. we should try to avoid reshape (and resize when you want to retain the memory, resize can be destructive as needed, does not guarantee that the memory stays the same, unless the underlying storage is not being resized. in this case, the storage is not being changed, but a good practice to use :view)

nicholas-leonard:
Done.

nicholas-leonard:
Now accepts non-contiguous input and gradOutput.

WeightedEuclidean.lua
```diff
@@ -25,55 +17,71 @@ function WeightedEuclidean:reset(stdv)
...
-      for i=1,self.templates:size(2) do
-         self.templates:select(2, i):apply(function()
-            return torch.uniform(-stdv, stdv)
-         end)
-      end
+   self.weight:uniform(-stdv, stdv)
+   self.diagCov:fill(1)
+end
+
+function view(res, src, ...)
```
soumith: this should be a local function! you are setting it to be a global function

soumith: another note on varargs (i.e.  ...) in terms of performance. In constructors etc they are okay to use, but it is not good to use them in high-performance functions like :forward and :backward phases, as they are not JIT-compiled. i.e. if the JIT compiler encounters a varargs function, it halts and goes to interpreter mode.

nicholas-leonard: it halts for the function only?

soumith: it halts for the sequence of instructions so far that i was planning to JIT. Example: if it wanted to inline the function as part of a larger block, it wont be able to.

nicholas-leonard:
done.

soumith:
thanks!

PR #135, opened 2015-01-06
Parallel, Container & cie

nicholas-leonard:
This PR adds a couple of things:
 * fixes the Parallel:updateOutput where self.module[i]:updateOutput(input) was being called twice...
 * added unit tests for Parallel and ParallelTable
 * ParalleTable and ConcatTable now inherit nn.Container
 * added nn.Container doc
 * the containers now call torch.type(self) to get a string representation of their name, which facilitates inheritance

### DIFF
--- a/Concat.lua
+++ b/Concat.lua
@@ -1,7 +1,7 @@
 local Concat, parent = torch.class('nn.Concat', 'nn.Container')
 
 function Concat:__init(dimension)
-   parent.__init(self, dimension)
+   parent.__init(self)
    self.size = torch.LongStorage()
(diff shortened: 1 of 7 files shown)


jonathantompson:
"fixes the Parallel:updateOutput where self.module[i]:updateOutput(input)
was being called twice"  It was?  Wow, that's bad.

nicholas-leonard:
@jonathantompson Yeah!

soumith:
thats a great catch! i guess no one noticed it because it produces the same results, but hell, who would want the extra fprop!

PR #136, opened 2015-01-06
new Copy constructor arguments

nicholas-leonard:
Hi,

This PR adds two optional arguments to nn.Copy (see doc for explanation). Also adds a nn.Copy unit test.


soumith:
nice, needed features!

PR #139, opened 2015-01-08
Added more informative pretty-printing.

jjh42:


Linear.lua
```diff
@@ -99,3 +99,9 @@ end
 
 -- we do not need to accumulate parameters when sharing
 Linear.sharedAccUpdateGradParameters = Linear.accUpdateGradParameters
+
+
+function Linear:__tostring__()
+  return 'nn.Linear' ..
```
soumith: it would be better to do:
return torch.type(self) ..

so that if you ever subclass nn.Linear, the string returned will have the subclass name

Reshape.lua
```diff
@@ -61,3 +61,8 @@ function Reshape:updateGradInput(input, gradOutput)
    self.gradInput:viewAs(gradOutput, input)
    return self.gradInput
 end
+
+
+function Reshape:__tostring__()
+  return 'nn.Reshape(' .. table.concat(self.size:totable(), 'x') .. ')'
```
soumith: same comment as above

jjh42:
@soumith good point. have fixed and added a test

nicholas-leonard:
Yay. Nice stuff.

soumith:
thanks @jjh42 ! looks good.

PR #140, opened 2015-01-10
nn returns package table

akfidjeland:


andresy:
thanks! [should have been done since ages, and should do the same for other modules]
btw, why travis fails??

soumith:
Travis is sucking to run unit tests, because of openmp or openblas bug on VM. Has been happening for a while, I dont know how to fix it

Issue #141, opened 2015-01-15
Spatial Adaptive Max Pooling


fmassa:
@soumith We have implemented both c and cuda versions of Spatial Adaptive Max Pooling, which is used in SPP paper. 
Basicaly for a given output size and for any input it always returns a tensor of this output size.
We can put it to inn or to nn and cunn, what do you think ?

soumith:
that sounds like a cool module. traditionally, to make the output length-invariant, we used to completely collapse the input in one dimension (like nn.Max() or nn.Sum()). This one will have much better output statistics and retain more information.

We're happy to have it in nn/cunn with the standard requirements: documentation + unit test

jonathantompson:
@fmassa Do you mean this paper: http://www.eecs.berkeley.edu/~jiayq/assets/pdf/cvpr12_pooling.pdf ?

fmassa:
@jonathantompson I was thinking about this paper actually http://arxiv.org/pdf/1406.4729v3.pdf.
The specific details are not present in the paper, but in the authors code.
This SpatialAdaptiveMaxPooling is similar to the standard SpatialMaxPooling, but the kW, kH, dW and dH varies with respect to the position of the pooling window in the image.

fmassa:
The cuda version is almost ready, there are just some corner cases to check.
BTW, I'm using the old cutorch API, without states. Should I wait for it to be merged before PR ?

soumith:
dont wait for it to merge, i'll repatch your modules. i wont be able to merge the cutorch/cunn STATE patches until next week.

alexbw:
Getting an import error on `require 'nn'`, because SpatialAdaptiveMaxPooling.lua does not exist in the main repo right now.

soumith:
reverted the PR, usually I test it out end to end, but today in haste i just did a code review and merged. apologies.

PR #142, opened 2015-01-16
Added SpatialAdaptiveMaxPooling

fmassa:
Adds SpatialAdaptiveMaxPooling as discussed in https://github.com/torch/nn/issues/141


soumith:
looks good to me! thanks a lot.

PR #143, opened 2015-01-16
Revert "Added SpatialAdaptiveMaxPooling"

soumith:
Reverts torch/nn#142 build breaks! please test before sending the PR


fmassa:
I'm really sorry, I forgot the SpatialAdaptiveMaxPooling.lua file. First PR...
Should be fixed with https://github.com/torch/nn/pull/144. @szagoruyko also tested it and it works.

PR #145, opened 2015-01-17
SpatialConvolutionMM supports dw/dh != 1

Aysegul:


test.lua
```diff
@@ -1095,15 +1095,18 @@ function nntest.SpatialConvolution()
...
+   local kj = math.random(1,5)
+   local di =  math.random(1,4)
+   local dj =  math.random(1,4)
+   local padding = math.random(0,2)
    local outi = math.random(5,9)
    local outj = math.random(5,9)
-   local ini = outi-1+ki
-   local inj = outj-1+kj
-   local module = nn.SpatialConvolutionMM(from, to, ki, kj)
+   local ini = (outi-1)*si+ki-padding*2
```
soumith: si and sj are not defined, you must mean di and dj.
On that note, did you unit test this? as your PR as such wont run the unit test because si/sj are not defined.

soumith:
there's a bug in the unit test, made an inline comment.
great stuff from you as usual ayse, let me know when that is fixed.

Aysegul:
fixed it now, sorry had copied a wrong test before. 
Thanks a lot soumith.

soumith:
Thanks ayse, had to squash your commits, and manual merge, merged in trunk now:
https://github.com/torch/nn/commits/master

PR #149, opened 2015-01-21
adding getParametersByDevice and a scatter-gather pattern for parameters...

soumith:
.... Some other minor changes.

This allows us to start thinking about moving the nn.DataParallel and nn.ModelParallel containers into torch/nn rather than in fbnn/fbcunn which have a huge set of dependencies.

### DIFF
--- a/Module.lua
+++ b/Module.lua
@@ -3,6 +3,7 @@ local Module = torch.class('nn.Module')
 function Module:__init()
    self.gradInput = torch.Tensor()
    self.output = torch.Tensor()
+   self._type = self.output:type()
 end
 
 function Module:parameters()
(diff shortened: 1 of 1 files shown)


koraykv:
Should we put these in cunn instead of nn? It depends on cutorch and many of the code logic is specific to GPUs.

soumith:
cunn at this point does not have a lua side of things, but we can introduce it there.

I think, forgetting cunn for once, this wants to introduce a notion of parameters on a device (any device, even say a remote machine, or tomorrow we get a FPGA API). 
Right now, the logic has gpu_assignments and stuff, which I agree we should be cleaned up, separated and moved to cunn.

Another small function it introduces, if you notice, is the :for_each, i.e. an apply operation over each module in the network, which is missing now, and is very useful.

andresy:
note: given the standard way of naming functions is forEach or foreach in torch, for_each seems bizarre in nn.

koraykv:
I see, but it also uses depends on cutorch at this time maybe those bits can be moved too.

About forEach or _for_each or for_each :), in nngraph there is an apply function call and since this is the same, maybe we should name this apply. And maybe this should be added to Container.lua too?

andresy:
agreed

jonathantompson:
"cunn at this point does not have a lua side of things"

I wonder if we can rethink splitting cunn and nn.  It seems that many
things get a little easier if cunn is merged with nn and the cmake file
takes care of whether or not to compile the cuda framework.  Here's a few
reasons why I think it makes sense:

1. The code in cunn is logically equivalent to the CPU counterpart so they
should exist in the same package for better readability.
2. We wouldn't have to do 2 pull requests every time we want to add
functionality to nn and cunn (ie Soumith's rules state now that all new
Modules must have a cpu and gpu implementation = 2 PRs).
3. Testing becomes easier because you no longer have upgrade dependencies
between nn and cunn.

Anyway, I feel like I've mentioned this before so I'll stop now :-)

soumith:
agreed, for_each has to become apply

clementfarabet:
Hey! what's the status of this PR? Besides the for_each that should renamed forEach ( :-) ), is the rest of the patch backward-compatible? Also, @soumith , would this make your [image-net examples](https://github.com/soumith/imagenet-multiGPU.torch) work without having to pull the separate branch? It looks like if this gets merged, then these [4 modules](https://github.com/soumith/imagenet-multiGPU.torch/tree/master/fbcunn_files) could be merged into cunn? I'm just catching up with this.

soumith:
it's on my TODO list to clean this up, remove most of it, make the :for_each to be :apply, and the GPU specific stuff be minimalized and put in cunn.

soumith:
Will happen in the "near future".

clementfarabet:
Ok, great.

soumith:
not merging this in. Moving to DataParallelTable makes things cleaner.

PR #150, opened 2015-01-22
Simpler LPPooling with AveragePooling instead of SubSampling

szagoruyko:
Has to be a bit faster and cleaner. I also had a case when parameters of SubSampling changed during training with CUDA.


soumith:
this will not be compatible with models loaded from disk. 

To keep it os, I suggest that you keep the empty gradParameters stub functions

szagoruyko:
agree, updated

soumith:
looks good to me!

PR #151, opened 2015-01-23
Matrix matrix multiplication layer.

dominikgrewe:
Multiplies two matrices (or two batches of matrices).
This is a pure Lua implementation taking advantage of the new torch.bmm
function for minibatch inputs.

### DIFF
--- a/MM.lua
+++ b/MM.lua
@@ -1,0 +1,88 @@
+--[[ Module to perform matrix multiplication on two minibatch inputs,
+     producing a minibatch.
+]]
+
+local MM, parent = torch.class('nn.MM', 'nn.Module')
+
+--[[ The constructor takes two optional arguments, specifying whether or not transpose
+     any of the input matrices before perfoming the multiplication.
+]]
+function MM:__init(transA, transB)
+  parent.__init(self)
+
+  self.transA = transA or false
+  self.transB = transB or false
+
+  self.gradInput = {torch.Tensor(), torch.Tensor()}
+end
+
+function MM:updateOutput(input)
+  assert(#input == 2, 'input must be a pair of minibatch matrices')
+  local a, b = unpack(input)
+  assert(a:nDimension() == 2 or a:nDimension() == 3, 'input tensors must be 2D or 3D')
+
+  if a:nDimension() == 2 then
+    assert(b:nDimension() == 2, 'second input tensor must be 2D')
+
+    if self.transA then a = a:t() end
+    if self.transB then b = b:t() end
+    assert(a:size(2) == b:size(1), 'matrix sizes do not match')
+
+    self.output:resize(a:size(1), b:size(2))
+    self.output:mm(a, b)
+  else
+    assert(b:nDimension() == 3, 'second input tensor must be 3D')
+    assert(a:size(1) == b:size(1), 'inputs must contain the same number of minibatches')
+
+    if self.transA then a = a:transpose(2, 3) end
+    if self.transB then b = b:transpose(2, 3) end
+    assert(a:size(3) == b:size(2), 'matrix sizes do not match')
+
+    self.output:bmm(a, b)
+  end
+
+  return self.output
+end
+
+function MM:updateGradInput(input, gradOutput)
+  assert(#input == 2, 'input must be a pair of tensors')
+  local a, b = unpack(input)
+  self.gradInput[1]:resizeAs(a)
+  self.gradInput[2]:resizeAs(b)
+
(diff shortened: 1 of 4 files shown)


soumith:
this seems slightly short-sighted. 
Would it not be much better if there's one module called MatrixMatrixMultiply, or nn.mm that either took in a single matrix-pair os a batch of matrix-pairs (just like many other nn modules which take a single sample or a batch of samples....)

will let others comment as well.

dominikgrewe:
Fair point. If the policy is that every module should support single inputs, then we should do that here too.
I've updated the PR.

MatrixMatrixMultiply.lua
```diff
@@ -0,0 +1,88 @@
...
+  self.gradInput = {torch.Tensor(), torch.Tensor()}
+end
+
+function MatrixMatrixMultiply:updateOutput(input)
+  assert(#input == 2, 'input must be a pair of minibatch matrices')
+  local a, b = unpack(input)
+  assert(a:nDimension() == 2 or a:nDimension() == 3, 'arguments must be a 2D or 3D Tensor')
+
+  if a:nDimension() == 2 then
+    assert(b:nDimension() == 2, 'arg 2 must be a 3D Tensor')
```
nicholas-leonard: 3D Tensor?

soumith: Yes 3D for batch mm

soumith: Ah I see for non-batch it is printed wrong

dominikgrewe:
Good catch, thanks! I've updated that one and also below.

soumith:
The last comment from me is about the naming.

Traditionally, the nn package has nn.dot, or nn.CAdd or nn.Add, nn.Select etc, which imply that it is doing what it's torch equivalent would do.
If that spirit is maintained, this should be named nn.mm , what do you guys think...

soumith:
or nn.MM

nicholas-leonard:
nn.MM

dominikgrewe:
Okay, I've renamed the module "nn.MM"

MM.lua
```diff
@@ -0,0 +1,88 @@
...
+  assert(#input == 2, 'input must be a pair of minibatch matrices')
+  local a, b = unpack(input)
+  self.gradInput[1]:resizeAs(a)
+  self.gradInput[2]:resizeAs(b)
+
+  assert(gradOutput:nDimension() == 2 or gradOutput:nDimension() == 3, 'arguments must be a 2D or 3D Tensor')
+
+  local h_dim, w_dim, f
+  if gradOutput:nDimension() == 2 then
+    assert(a:nDimension() == 2, 'arg 2 must be a 2D Tensor')
```
nicholas-leonard: Could you make the messages here more specific? Like arg 2a vs arg 2b or something.

MM.lua
```diff
@@ -0,0 +1,88 @@
...
+
+    if self.transA then a = a:t() end
+    if self.transB then b = b:t() end
+    assert(a:size(2) == b:size(1), 'matrix sizes do not match')
+
+    self.output:resize(a:size(1), b:size(2))
+    self.output:mm(a, b)
+  else
+    assert(b:nDimension() == 3, 'arg 2 must be a 3D Tensor')
+    assert(a:size(1), b:size(2), 'inputs must contain the same number of minibatches')
```
nicholas-leonard: `a:size(1) == b:size(1)`?

doc/simple.md
```diff
@@ -831,3 +832,14 @@ gnuplot.grid(true)
...
+## MM ##
+
+`module` = `nn.MM(transA, transB)`
+
+Performs multiplications on one or more pairs of matrices.
+If `transA` is set, the first matrix is transposed before multiplication.
+If `transB` is set, the second matrix is transposed before multiplication.
+By default, the matrices do not get transposed.
+
+The module also accepts 3D inputs which are interpreted as batches of matrices.
```
nicholas-leonard: Can you mention what axis order is expected :  A : b x m x n, B: b x n x m. It makes it easier to understand. Maybe add a small example.

nicholas-leonard:
Very cool module.

MM.lua
dominikgrewe: I've change the message to hopefully make it clearer.

MM.lua
dominikgrewe: Good catch, thanks!

doc/simple.md
dominikgrewe: I've extended the documentation a bit and also provided an example.

soumith:
@dominikgrewe looks good to me. squash and will merge.

dominikgrewe:
Thanks! Squashed.

soumith:
thanks, great PR!

PR #154, opened 2015-01-31
L1Penalty bug fixed

jakezhaojb:
`self.provideOutput` bug fixed.

### DIFF
--- a/L1Penalty.lua
+++ b/L1Penalty.lua
@@ -4,10 +4,11 @@ local L1Penalty, parent = torch.class('nn.L1Penalty','nn.Module')
 --[gradOutput] to the gradient of the L1 loss. The [input] is copied to 
 --the [output]. 
 
-function L1Penalty:__init(l1weight, sizeAverage)
+function L1Penalty:__init(l1weight, sizeAverage, provideOutput)
     parent.__init(self)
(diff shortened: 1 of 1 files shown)


soumith:
Thanks jake

L1Penalty.lua
```diff
@@ -4,10 +4,11 @@ local L1Penalty, parent = torch.class('nn.L1Penalty','nn.Module')
 --[gradOutput] to the gradient of the L1 loss. The [input] is copied to 
 --the [output]. 
 
-function L1Penalty:__init(l1weight, sizeAverage)
+function L1Penalty:__init(l1weight, sizeAverage, provideOutput)
     parent.__init(self)
     self.l1weight = l1weight 
     self.sizeAverage = sizeAverage or false  
+    self.provideOutput = provideOutput or true
```
fmassa: Line 11 always evaluate to true, even if provideOutput is false

soumith: you are right. i should not merge PRs from mobile, even if they are super-small!

soumith: fixed in trunk via https://github.com/torch/nn/commit/b74da0080e3f6af53836e36478d6df58c572d351

PR #156, opened 2015-02-04
Resized nn.MM output to support CudaTensors.

fidlej:
The CudaTensor.bmm expects the output to be already a 3D tensor.

### DIFF
--- a/MM.lua
+++ b/MM.lua
@@ -38,6 +38,7 @@ function MM:updateOutput(input)
     if self.transB then b = b:transpose(2, 3) end
     assert(a:size(3) == b:size(2), 'matrix sizes do not match')
 
+    self.output:resize(a:size(1), a:size(2), b:size(3))
     self.output:bmm(a, b)
   end
 


soumith:
hmmm, bmm seems to try to resize it: https://github.com/torch/cutorch/blob/master/TensorMath.lua#L346 wonder why that does not work. cc: @dominikgrewe

dominikgrewe:
The torch version of bmm does, but I just noticed that the cutorch version doesn't. The cutorch version creates a new tensor of the right size when calling `torch.bmm(a, b)` but it doesn't resize `c` in `c:bmm(a, b)`

I'll fix it in cutorch to be consistent with torch. Then we don't need this fix.

dominikgrewe:
Actually, torch.mm has the same behaviour as cutorch.bmm (i.e. not resizing `c` in `c:mm(a, b)`). So maybe it's torch.bmm that's behaving the "wrong" way?

Considering that we resize in the non-batch case, I think we should also resize in the batch case. @soumith What do you think?

soumith:
yea if we resize the non-batch case, then the batch  case has to be consistent. i agree.
can you provide pointers to the code, i can fix it. or a PR would be nice too.

dominikgrewe:
Just to clarify: when I talked about resizing the batch case because we resize the non-batch case, I was talking about `nn.MM`. In other words, I think we should merge this PR.

To make torch and cutorch consistent, I'd suggest we make sure that the torch version of `bmm` behaves the same as `mm` (and as the cutorch version of `bmm`). If we change `bmm` to use the new `baddbmm` (as we do in cutorch), this will follow automatically. I'm happy to prepare a PR for that.

dominikgrewe:
PR for torch here: torch/torch7#144

soumith:
fair point, i see in line 31, there's a similar resize for non-batch case.

PR #157, opened 2015-02-04
L2 MultiMarginCriterion

szagoruyko:
This PR adds p=2 in:
```lua
loss(x,y) = forward(x,y) = sum_i(max(0, 1 - (x[y] - x[i]))^p) / x:size(1)
```
Also added test, cuda version is coming.

### DIFF
--- a/generic/MultiMarginCriterion.c
+++ b/generic/MultiMarginCriterion.c
@@ -6,6 +6,7 @@ static int nn_(MultiMarginCriterion_updateOutput)(lua_State *L)
 {
   THTensor *input = luaT_checkudata(L, 2, torch_Tensor);  
   int sizeAverage = luaT_getfieldcheckboolean(L, 1, "sizeAverage");
+  int p = luaT_getfieldchecknumber(L, 1, "p");
   real *input_data, *target_data;
   long nframe, dim;
   long t, d;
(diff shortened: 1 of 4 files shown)


generic/MultiMarginCriterion.c
```diff
@@ -54,7 +55,7 @@ static int nn_(MultiMarginCriterion_updateOutput)(lua_State *L)
         continue;
     
       if(z > 0)
-        sum += z;
+        sum += (p==1) ? z : z*z;
```
soumith: this only works for p=1 and p=2.

generic/MultiMarginCriterion.c
```diff
@@ -127,8 +129,9 @@ static int nn_(MultiMarginCriterion_updateGradInput)(lua_State *L)
     
       if(z > 0)
       {
-        gradInput_target -= g;
-        gradInput_data[d] = g;
+        real h = (p == 1) ? g : 2*g*z;
```
soumith: same, only works for p=1 or p=2. if this is by design, assert in the constructor. if not, fix it appropriately.

szagoruyko: added an assert in constructor for p=1 and p=2

soumith:
thanks!

PR #158, opened 2015-02-06
Module:listModules()

nicholas-leonard:
This PR includes doc and unit tests. The method has already been used for nn.Recurrent in nnx.

### DIFF
--- a/Module.lua
+++ b/Module.lua
@@ -287,3 +287,27 @@ function Module:findModules(typename, container)
   return nodes, containers
 end
 
+-- returns a list of modules
+function Module:listModules()
+   local function tinsert(to, from)
+      if torch.type(from) == 'table' then
+         for i=1,#from do
+            tinsert(to,from[i])
+         end
+      else
+         table.insert(to,from)
+      end
+   end
+   -- include self first
+   local modules = {self}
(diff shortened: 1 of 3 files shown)


soumith:
there is a "for_each" submitted in this PR https://github.com/torch/nn/pull/149
That is probably going to be merged, making this redundant.

That will mostly be renamed to :apply, just like in nngraph, and torch tensors.

andresy:
we have here and there things in torch which implement :totable() ... should it be that here?

nicholas-leonard:
It isn't the same as for_each nor apply which takes a callback. This method is already used in nn.Recurrent which I intend to transfer from nnx to nn. I don't see why we could have both for_each and listModules? They have different uses in mind. listModules is more like parameters() where a flattened list of modules (instead of parameters) is returned.

soumith:
okay great. how do you feel about :listModules -> :totable()

nicholas-leonard:
To me totable() is more ambiguous than listModules(). Like, what exactly is totable? modules, parameters, outputs? But if you insist, I can take the time to change the nn.Recurrent code to use totable() instead.

soumith:
thanks!

PR #160, opened 2015-02-06
nn.Recurrent

nicholas-leonard:
This PR provides the Recurrent module that can be used to build RNNs. Depends on #158 (for listModules). Includes docs and unit tests.

I also removed redundant lines from `ParallelTable:__tostring__` where `(i == self.modules)` is never true.

### DIFF
--- a/Recurrent.lua
+++ b/Recurrent.lua
@@ -1,0 +1,529 @@
+------------------------------------------------------------------------
+--[[ Recurrent ]]--
+-- Ref. A.: http://goo.gl/vtVGkO (Mikolov et al.)
+-- B. http://goo.gl/hu1Lqm
+-- Processes the sequence one timestep (forward/backward) at a time. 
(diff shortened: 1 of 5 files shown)


Recurrent.lua
```diff
@@ -0,0 +1,529 @@
...
+   gradOutput_:resizeAs(gradOutput):copy(gradOutput)
+end
+
+function Recurrent:accGradParameters(input, gradOutput, scale)
+   -- Back-Propagate Through Time (BPTT) happens in updateParameters()
+   -- for now we just keep a list of the scales
+   self.scales[self.step-1] = scale
+end
+
+-- not to be confused with the hit movie Back to the Future
```
soumith: lol!

nicholas-leonard: forgot I put that there lol.

nicholas-leonard:
I will be keeping this in nnx until we get a working LSTM implementation.

PR #161, opened 2015-02-09
making SpatialConvolution use SpatialConvolutionMM ops at the backend

### LINKED ISSUE #159: make SpatialConvolution an alias for SpatialConvolutionMM
No one practically uses SpatialConvolution, it does not work on CUDA, it confuses new users.

So, just make SpatialConvolution as an alias for SpatialConvolutionMM.

Approved by the core devs.

soumith:
fixes #159


soumith:
pretty straight-forward PR, made sure that the loading is backward-compatible for older serialized models. merging, but feel free to comment.

PR #165, opened 2015-02-14
Parametric ReLU

szagoruyko:
Parametric ReLU from this paper http://arxiv.org/abs/1502.01852
Has also test and doc with image.


generic/PReLU.c
```diff
@@ -0,0 +1,240 @@
...
+
+      if(input->size[(input_ndim + 1) % 2] != nOutputPlane)
+	THError("wrong number of input planes");
+    }
+
+    real* output_data = THTensor_(data)(output);
+    real* input_data = THTensor_(data)(input);
+    real* weight_data = THTensor_(data)(weight);
+    long i,j,k;
+#pragma omp parallel for private(i)
```
nagadomi:
```C
#pragma omp parallel for private(j, k)
```

nicholas-leonard:
Looks nice.

generic/PReLU.c
szagoruyko: fixed thanks

soumith:
looks good! squash your commits and I'll merge.
Mention in the doc that you should not apply weight decay to prelu units.

soumith:
Ah i see you already have the "weight decay should not used on it" line.

nicholas-leonard:
Also, could you add a link to the article to the doc?

szagoruyko:
added the link to the doc and squashed

soumith:
awesome! thanks sergey

PR #169, opened 2015-02-23
CosineEmbeddingCriterion: zero grads below margin

deltheil:
For negative pairs (`y == - 1`) the margin was NOT taken into account when computing the gradients.

In other words `updateGradInput` did return non zero gradients when the cosine similarity was below the margin (e.g this is not the case with [L1HingeEmbeddingCriterion](https://github.com/torch/nn/blob/822a742/L1HingeEmbeddingCriterion.lua#L33-L35) which behaves properly).

I have included a tiny non regression test.


soumith:
Thanks cedric, looks good.

PR #172, opened 2015-02-27
Prevent sharing parameters between unrelated modules

mys007:
I'm proposing a safety check to prevent accidental sharing of parameters between e.g. convolutions of different sizes.

### DIFF
--- a/Module.lua
+++ b/Module.lua
@@ -93,6 +93,7 @@ function Module:share(mlp, ...)
    local arg = {...}
    for i,v in ipairs(arg) do
       if self[v] ~= nil then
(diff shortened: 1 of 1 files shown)


soumith:
a correct check would be :nElement() == :nElement(). isSameSizeAs is too restrictive for people wanting to do funky things...

nicholas-leonard:
I am not a big fan of this, even with nElement. Sometimes you can create a smaller module (instead of cloning the big one), and just share that to save on memory. With this kind of assert, that wouldn't be possible.

soumith:
i see nick's point. you often share temporary buffers (like if you use the SpatialConvolution, sharing the  unfolding buffer .finput across all modules is perfectly valid and reduces consumption).

PR #175, opened 2015-02-27
Update criterion.md

Atcold:
Missing code type layout


soumith:
that is not lua, so no need of lua highlighting imo

Atcold:
Everything is lua... If you check all the docs, everything is highlighted nicely. Functions are one colour, methods are another one, number etc...
It would look nicer, but I forgot to add to the previous commit.

PR #176, opened 2015-03-03
Add batch mode for VolumetricMaxPooling

### LINKED ISSUE #171: VolumetricMaxPooling does not support batch mode
Only works for 4D Tensor inputs.

mlosch:
Resolves #171.
Note: Torch Tensor does not offer a resize6d function, which is necessary for batch mode. I replaced that with the generic tensor resize function for now. This requires an additional free() call.


soumith:
Awesome! thanks Max

PR #178, opened 2015-03-04
Fixed SpatialAveragePooling - added division by filter size

adampolyak:
SpatialAveragePooling layer only summed the filters without dividing by filter size, the following code failed:
```lua
require 'nn'
require 'cudnn'
a = torch.rand(1,1,2,2):fill(1)
l = nn.SpatialAveragePooling(2,2)
b = l:forward(a)

a2 = torch.rand(1,1,2,2):fill(1):cuda()
l2 = cudnn.SpatialAveragePooling(2,2):cuda()
b2 = l:forward(a)

assert(b==b2)
```
Fixed both forward and backward.

### DIFF
--- a/generic/SpatialAveragePooling.c
+++ b/generic/SpatialAveragePooling.c
@@ -83,7 +83,7 @@ static int nn_(SpatialAveragePooling_updateOutput)(lua_State *L)
             ptr_input += inputWidth; /* next input line */
           }
           /* Update output */
-          *ptr_output++ += sum;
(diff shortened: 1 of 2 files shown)


soumith:
thanks!

soumith:
also probably need to change cunn

szagoruyko:
but it breaks saved models

soumith:
ok reverted it, can you send a patch that does not break saved models? Also maybe a cunn change as well.

PR #180, opened 2015-03-04
SpatialAveragePooling divides by kW*kH

szagoruyko:
Fixes https://github.com/torch/nn/pull/178  with compatibility with saved models
Note: cudnn, ccn2, imagine-nn are dividing by kW*kH
Also had to fix SpatialLPPooling


soumith:
so quick! thanks

adampolyak:
Fast indeed! Thanks

PR #182, opened 2015-03-05
fixed Power.lua

rotmanmi:
The previous version won't support inputs which contain 0s due to the fact it was dividing by the input. This version solves this problem.

### DIFF
--- a/Power.lua
+++ b/Power.lua
@@ -15,7 +15,10 @@ function Power:updateOutput(input)
 end
 
 function Power:updateGradInput(input, gradOutput)
+   self.buffer = self.buffer or input.new()
+   self.buffer:resizeAs(input):copy(input)
+   self.buffer:pow(self.pow - 1)
    self.gradInput:resizeAs(input):copy(gradOutput)
-   self.gradInput:cmul(self.output):cdiv(input):mul(self.pow)
+   self.gradInput:cmul(self.buffer):mul(self.pow)
    return self.gradInput
 end


Power.lua
```diff
@@ -1,21 +1,21 @@
...
+
+function Power:updateOutput(input)
+   self.output:resizeAs(input):copy(input)
+   self.output:pow(self.pow)
+   return self.output
+end
+
+function Power:updateGradInput(input, gradOutput)
+   self.gradInput:resizeAs(input):copy(gradOutput)
+   self.gradInput:cmul(torch.pow(input,self.pow-1)):mul(self.pow)
```
soumith: torch.pow(input, self.pow-1) will create a newly allocated tensor every single time as a temporary buffer. This is bad practice and will also give  a bad slow-down on GPU.

So instead, have a temporary buffer:
```lua
self.buffer:resizeAs(input):copy(input)
self.buffer:pow(self.pow-1)
self.gradInput:cmul(self.buffer):mul(self.pow)
```

soumith:
thansk for the change. looks good. if you squash your commits, I will merge it in.

soumith:
thanks Michael, I squashed your commits in #185 and pushed it to master.

For future refernce, the key here is to do git push --force after rebasing. 
Your commit has been intact with proper credit https://github.com/torch/nn/commit/ffaf24a73562b18cf992406b5f2d8a8174e8a4dd

PR #188, opened 2015-03-13
added doc + test case for CrossEntropyCriterion

andresy:


soumith:
thanks ronan! you're the best

PR #192, opened 2015-03-16
Remove return statement

lukemarris:
I think this line is causing problems.


andresy:
what problem?

lukemarris:
Sorry I made a mistake this is not breaking things but perhaps it is still inconsistent with the other modules?

soumith:
@lukemarris This is correct behavior wrt lua 5.2 and forward. I can patch the other modules to follow this if you'd like.

lukemarris:
@soumith No need. Thank you for the explanation.

PR #196, opened 2015-03-18
Added 2 calls to collect garbage for getParameters

zhangxiangxiao:
These 2 calls of collectgarbage() can save a large chunk of temporary memory when calling getParameters(). It is particularly useful when working with GPUs on a model with large number of parameters, for which it would otherwise be impossible to even call getParameters() (out of memory cuda error).

### DIFF
--- a/Module.lua
+++ b/Module.lua
@@ -241,7 +241,9 @@ function Module:getParameters()
 
    -- flatten parameters and gradients
    local flatParameters = flatten(parameters)
+   collectgarbage()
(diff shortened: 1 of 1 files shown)


soumith:
Looks good to me. as it is not part of performant loop, it is fine. Anyone has any problems with this?

clementfarabet:
no this is great

On Wed, Mar 18, 2015 at 10:06 AM, Soumith Chintala <notifications@github.com

andresy:
ok for me

PR #197, opened 2015-03-19
Pass weights to cuda implementation of ClassNLLCriterion

mlosch:
Prepares criterion for new functionality in cunn package.


soumith:
Hey max, you have access to the self objext, so you don't actually have to change the function interface. (Look at SpatialConvolutionMM.c/cu and how it gets the finput buffer as an example )

It is a small change, and keeps things consistent, so I'm closing this PR.

Let me know if you want help with this, I'm happy to lend a hand.

PR #200, opened 2015-03-20
Module:sharedClone()

nicholas-leonard:
This PR adds Module:sharedClone(). 

With this PR, the common pattern of:
```lua
clone = mlp:clone() -- this copies all the parameters and gradParameters
clone:share(mlp,"weight","gradWeight","bias","gradBias") -- this deletes and replaces them with mlp's
```
will become the more memory efficient:
```lua
clone = mlp:sharedClone()
```

And because each sub-module can override static variables `__parameters__` and `__gradParameters__`. You can call huge structures of Modules with different parameters names and still expect them to correctly be shared.

Not sure this will make it into nn. So I will only work on the doc when this gets approval. If you would like this method to be usable for attributes other than parameters/gradParameters, I could make this method `Module:sharedParamClone(bool shareGradParams)` and make it call `Module:sharedClone(...)` where `...` is the attributes like : `weight, bias, etc.`.

### DIFF
--- a/Module.lua
+++ b/Module.lua
@@ -1,5 +1,8 @@
 local Module = torch.class('nn.Module')
 
+Module.__parameters__ = {'weight', 'bias'}
+Module.__gradParameters__ = {'gradWeight', 'gradBias'}
+
 function Module:__init()
    self.gradInput = torch.Tensor()
    self.output = torch.Tensor()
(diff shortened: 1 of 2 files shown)


nicholas-leonard:
Not enough interest.

PR #202, opened 2015-03-21
ZeroGrad

nicholas-leonard:
Adds a Module that zeros the backward propagated gradients. 

I will add doc and unit tests if you want this.


soumith:
I'm ok with this, but will check with others and get back.

PR #204, opened 2015-03-24
add example for reshape with mini-batch

wb14123:
Reshape with the batch parameter makes people a little confused. This patch adds an example for it.


soumith:
Thanks Bin!

PR #206, opened 2015-03-24
SpatialConvolutionMM: fixed corruption on small padded tensors

mys007:
There is a bug in SpatialConvolutionMM when using padding and convolving tensors with filters which are larger than the input. For example, the following snippet produces a segfault:
```
require 'nn'
M = nn.SpatialConvolutionMM(4,4,5,5,1,1,2)
T = torch.Tensor(4,1,1):zero()
print(M(T))
```
My assumption is that the code on lines 104-110 should produce the same result as lines 96-101 when dW=1. Therefore the fix.


soumith:
Thanks mate!

PR #207, opened 2015-03-24
Return self on Linear.reset so it can be used inline.

jjh42:
(this is especially helpful when using nngraph for doing reset inline).

Removed whitespace.


soumith:
Thanks Jonathan

PR #209, opened 2015-03-24
empty addBuffer fix

nicholas-leonard:
This PR fixes a bug where the Linear.addBuffer is empty. This happens when using some of the torch-utils tools where non parameter buffers are resized to zero when saved to disk, and then reloaded. 

Not sure this is a good reason to make the PR, but in any case, it fixes the bug.


soumith:
just ran into this today :)

PR #213, opened 2015-03-26
avoid for loop. seems to speed things up on GPU.

blundellc:


nicholas-leonard:
nice.

soumith:
nice trick. many thanks @blundellc

Issue #211, opened 2015-03-25
SpatialConvolutionMM cuda/cpu interoperability broken


mys007:
There are two caches in SpatialConvolutionMM: fgradInput and finput. The problem is that they are used in a different way in the 'nn' and 'cunn' C-implementations. Specifically, the cunn-implementation expects to find 1s in fgradInput while the nn-implementation stores arbitrary data in it. Thus, converting the module from CPU to CUDA makes it produce garbage in a very evil way because nothing crashes. For example

```
require 'nn'
require 'cunn'
require 'cutorch'

torch.setdefaulttensortype('torch.FloatTensor')

T = torch.Tensor(10,10,10):zero()
M = nn.SpatialConvolutionMM(10,1,5,5,1,1,0)

F = M:forward(T)
M:backward(T,F)

--M.fgradInput = torch.Tensor()   --UNCOMMENT ME TO FIX THIS

M=M:cuda()
C = M(T:cuda())

assert(math.abs( torch.norm(F:cuda()-C)) < 1e-5)
```

The proposed fix is to clear the caches at type conversion:
```
function SpatialConvolutionMM:type(type)
   self.finput = torch.Tensor()
   self.fgradInput = torch.Tensor()
   return parent.type(self,type)
end
```

soumith:
this is a pretty good catch. can you send a PR for this?

mys007:
Done...

PR #214, opened 2015-03-26
Fix for SpatialConvolutionMM cuda/cpu interoperability

mys007:
See https://github.com/torch/nn/issues/211


soumith:
thanks @mys007

PR #216, opened 2015-03-26
Added SpatialDropout + doc + test.

jonathantompson:
This is a super simple variation on Dropout that I have found empirically works better than dropout for fully-convolutional networks.  I explain about it in a little more detail here: http://arxiv.org/abs/1411.4280


soumith:
thanks Jono

PR #217, opened 2015-03-31
Making Squential() more talkative

Atcold:
Every time sequential breaks, it returns non module-specific error. Now it does.

### DIFF
--- a/Sequential.lua
+++ b/Sequential.lua
@@ -21,8 +21,10 @@ end
 
 function Sequential:updateOutput(input)
    local currentOutput = input
-   for i=1,#self.modules do 
-      currentOutput = self.modules[i]:updateOutput(currentOutput)
+   for i=1,#self.modules do
+      assert(pcall(function()
+         currentOutput = self.modules[i]:updateOutput(currentOutput)
+      end), 'Troubles with module ' .. i)
    end 
    self.output = currentOutput
    return currentOutput
@@ -33,7 +35,9 @@ function Sequential:updateGradInput(input, gradOutput)
    local currentModule = self.modules[#self.modules]
    for i=#self.modules-1,1,-1 do
       local previousModule = self.modules[i]
-      currentGradOutput = currentModule:updateGradInput(previousModule.output, currentGradOutput)
+      assert(pcall(function()
+         currentGradOutput = currentModule:updateGradInput(previousModule.output, currentGradOutput)
+      end), 'Troubles with module ' .. i+1)
       currentModule = previousModule
    end
    currentGradOutput = currentModule:updateGradInput(input, currentGradOutput)


soumith:
this is not a good idea. pcall is presumably slow, and also now you dont get the stack-trace of the error.

Atcold:
You get it
```
/usr/local/share/lua/5.1/nn/Sequential.lua:25: Troubles with module 21
stack traceback:
        [C]: in function 'assert'
        /usr/local/share/lua/5.1/nn/Sequential.lua:25: in function 'f'
        [string "local f = function() return netNew:forward(to..."]:1: in main chunk
        [C]: in function 'xpcall'
        /usr/local/share/lua/5.1/trepl/init.lua:620: in function 'repl'
        /usr/local/lib/luarocks/rocks/trepl/scm-1/bin/th:185: in main chunk
        [C]: at 0x010e6b57b0
```

Atcold:
Whereas, before, you were simply getting
```
/usr/local/share/lua/5.1/nn/Sequential.lua:25: attempt to index a nil value
stack traceback:
        /usr/local/share/lua/5.1/nn/Sequential.lua:25: in function 'f'
        [string "local f = function() return netNew:forward(to..."]:1: in main chunk
        [C]: in function 'xpcall'
        /usr/local/share/lua/5.1/trepl/init.lua:620: in function 'repl'
        /usr/local/lib/luarocks/rocks/trepl/scm-1/bin/th:185: in main chunk
        [C]: at 0x01091787b0
```

Atcold:
And if your model has 30+ modules, you just get crazy. I'm using `assert()` to show you the error stack.

soumith:
yes, but you lose the entire trace that the module generated.

soumith:
for example, what your actual error was.

Atcold:
Still, you don't know *which* module generated the error. Let's say that a `ReLU()` complains... *Which* one is the complaining one?

soumith:
@Atcold you can clearly see that in the stack trace.
```
require 'nn'
m=nn.Sequential()
m:add(nn.ReLU())

m:forward(torch.randn(10):float()) -- will fail because passing float to double network
```

```
th test.lua
/Users/xxx/torch/install/bin/luajit: /Users/xxx/torch/install/share/lua/5.1/nn/Threshold.lua:13: bad argument #1 (field output is not a torch.FloatTensor)
stack traceback:
	[C]: in function 'Threshold_updateOutput'
	/Users/chinso/torch/install/share/lua/5.1/nn/Threshold.lua:13: in function 'updateOutput'
	/Users/chinso/torch/install/share/lua/5.1/nn/Sequential.lua:25: in function 'forward'
	test.lua:5: in main chunk
	[C]: in function 'dofile'
	...inso/torch/install/lib/luarocks/rocks/trepl/scm-1/bin/th:131: in main chunk
	[C]: at 0x01094a3890
```

Atcold:
Not if you have 20 `ReLU()`s. We just got crazy the other day about this.

koraykv:
nn.Sequential is not the only thing you would want this right? At some
point, all container modules will have the same problem. And I guess one
can write a plug-in function to implement this behaviour on any container
they want. Having this inside a core module does not seem to be the best
option I think.

Atcold:
@koraykv, thank you for commenting.
Although I am a Torch user and documenter, I am not a programmer myself. Hence, I have troubles picking up your suggestion (I'm not sure what a *pick-in* function is), but yes, it would be very helpful having some sort of indication about what broke the code.
Now I just edit myself stuff in `/usr/local/share`, but not everyone has writing permissions there and it's still quite geeky for other users / students.

PR #220, opened 2015-04-02
Replicate now works across arbitrary dimensions.

egrefen:
@soumith Here is a clean version of the previous pull request.

nn.Replicate now works across arbitrary dimensions.


soumith:
Thanks Edward!

PR #231, opened 2015-04-20
Exp made lua only

szagoruyko:
C and CUDA code is redundant.


dominikgrewe:
Thanks!
Would you mind preparing a corresponding patch for cunn as well?

szagoruyko:
sure, will do

PR #233, opened 2015-04-21
ClassNLLCriterion now can take target as a number or Tensor of dim1 with...

soumith:
... 1 element in non-minibatch mode


soumith:
@Atcold you were looking for this fix.

Atcold:
@soumith, tnx.

PR #234, opened 2015-04-21
adding direct :backward to Concat, DepthConcat, Sequential

soumith:
This adds :backward to the three containers Concat, DepthConcat and Sequential.

If a module has a more efficient one-step :backward rather than updateGradInput + accGradParameters separately, then this will help the module.

The PR is open for comment, @andresy approved it already.

### DIFF
--- a/Concat.lua
+++ b/Concat.lua
@@ -61,6 +61,25 @@ function Concat:accGradParameters(input, gradOutput, scale)
    end
 end
 
+function Concat:backward(input, gradOutput, scale)
+   self.gradInput:resizeAs(input)
+   scale = scale or 1
+   local offset = 1
+   for i,module in ipairs(self.modules) do
+      local currentOutput = module.output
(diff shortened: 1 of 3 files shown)


nicholas-leonard:
This is a good idea. We already do this for nn.Recurrent and nn.LSTM (which are also containers).

szagoruyko:
will a module having only backward be accepted in nn?

soumith:
@szagoruyko hmmm, i dont think we are ready for that yet. We have to change all the containers this way first, not just these three containers in this PR

koraykv:
Just to give some context, Torch used to have only a backward method and the reason for separating these two is to be able use torch models for algorithms other than supervised learning where you do a forward/backward. For many problems that one needs to run an inference process, calculating the gradients wrt parameters actually is unnecessary and takes a lot of time.

andresy:
I concur there.

PR #235, opened 2015-04-21
adding in-place exp, sigmoid, softplus, sqrt, tanh

soumith:
cunn equivalents are ready as well.

### DIFF
--- a/Exp.lua
+++ b/Exp.lua
@@ -1,9 +1,30 @@
-local Exp = torch.class('nn.Exp', 'nn.Module')
+local Exp, parent = torch.class('nn.Exp', 'nn.Module')
+
+function Exp:__init(ip)
+   parent.__init(self)
+   -- default for inplace is false
+   self.inplace = ip or false
+   if (ip and type(ip) ~= 'boolean') then
+      error('in-place flag must be boolean')
+   end
+end
 
 function Exp:updateOutput(input)
-  return self.output:exp(input)
+  if self.inplace then
+     input:exp()
+     self.output = input
+  else
+     self.output:exp(input)
+  end
+  return self.output
 end
 
 function Exp:updateGradInput(input, gradOutput)
-  return self.gradInput:cmul(self.output, gradOutput)
+  if self.inplace then
+     gradOutput:cmul(self.output)
+     self.gradInput = gradOutput
+  else
+     self.gradInput:cmul(self.output, gradOutput)
+  end
+  return self.gradInput
 end
--- a/Sigmoid.lua
+++ b/Sigmoid.lua
@@ -1,6 +1,16 @@
-local Sigmoid = torch.class('nn.Sigmoid', 'nn.Module')
+local Sigmoid, parent = torch.class('nn.Sigmoid', 'nn.Module')
+
+function Sigmoid:__init(ip)
+   parent.__init(self)
+   -- default for inplace is false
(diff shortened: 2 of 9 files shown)


soumith:
need to update docs.

fmassa:
Those are nice additions to nn!
I just have a small remark regarding some specific combinations of inplace modules in the backward pass. If a module uses its output to compute the gradient, then inplace operation in the next model shouldn't be performed. Also, two inplace modules can't be used in sequence.
For example, the following network
```lua
m = nn.Sequential():add(nn.Exp()):add(nn.Sqrt(true))
```
won't compute correct gradients in the `Exp` module, because the backward pass in `Exp` uses `self.output`, which was changed in the inplace operation of `Sqrt`. It doesn't happen offen, but it might be difficult to catch it.

As most of `nn` modules, including convolution, pooling and linear, doesn't use `self.output` to compute their gradients, most networks will work just fine with those inplace modules though.

nicholas-leonard:
All those exceptions should be included in the doc.

Nicholas Léonard
450-210-1214

AjayTalati:
The caveats @fmassa has detailed should as be added to the `nngraph` docs. It's common to place nodes within nodes, especially exponentials and square roots.

fmassa:
For bijective functions, in-place operations can be concatenated together, but for that the backward pass should restore the input's module to its initial value. This introduces a few more computations in the backward pass, but maybe it's worth it ?

Exp.lua
```diff
@@ -1,9 +1,30 @@
-local Exp = torch.class('nn.Exp', 'nn.Module')
+local Exp, parent = torch.class('nn.Exp', 'nn.Module')
+
+function Exp:__init(ip)
+   parent.__init(self)
+   -- default for inplace is false
+   self.inplace = ip or false
+   if (ip and type(ip) ~= 'boolean') then
```
dominikgrewe: Nit: No need for parentheses around conditions in Lua.
Same elsewhere.

dominikgrewe:
The code looks good to me, but I agree that the possible pitfalls should be clearly documented, otherwise it's easy to shoot yourself in the foot.

> For bijective functions, in-place operations can be concatenated together, but for that the backward pass should restore the input's module to its initial value. This introduces a few more computations in the backward pass, but maybe it's worth it ?

That would require a module to know that the next module is in-place, doesn't it? Which isn't how nn works. Or do you want to re-compute regardless? Some modules seem to do that, e.g. `nn.AddConstant`.

fmassa:
@dominikgrewe given that `nn` modules are blind wrt their neighbor modules, the solution I mentionned is to always restore the input in the backward pass. As the majority of in place modules are pointwise operations, that wouldn't bring much overhead in the end, I think. That's how I did in `AddConstant` and `MulConstant`, as a proof of concept. But if one have several in place modules like that in a network, that could start to become a bit costly for no gains in most of the time.
But maybe is simply better to mention in the docs that one can't use two in place modules one after another, or that some combinations of modules with in place are not supported.

andreaskoepf:
I like the idea of @fmassa to use the inverse to restore inputs during the backward pass. Just as an speed-over-memory alternative: Inplace mode could be enabled in 'evaluation mode' only based on the module's 'train' member (currently used for dropouts and batch-normalization). Of course this would not help if the main motivation to introduce inplace-operations was to reduce the memory-usage during training...

PR #236, opened 2015-04-21
adding Volumetric average pooling, and some fixes for VolumetricMaxPooling

soumith:
Unit-tested. 
Need to add docs.


soumith:
cunn equivalents are ready and will be PR as well.

PR #240, opened 2015-04-22
fix View (hopefuly)

andresy:
now handles better -1 cases

i added yet another test case for this module.
better to try on your side, who knows?


soumith:
Seems to work, checked other cases like: nn.View(2,-1,5):setNumInputDims(3)

andresy:
hoorayh!

PR #241, opened 2015-04-22
Adding in-place AddConstant and MulConstant

fmassa:
Additionally, it also restores the initial value of the input in its backward pass, allowing it to be used after other in-place modules.
For example,
```lua
m = nn.Sequential()
m:add(nn.AddConstant(5,true)):add(nn.MulConstant(2,true)):add(nn.AddConstant(1,true))
```
works as its out-place equivalent, for both forward and backward functions.

### DIFF
--- a/AddConstant.lua
+++ b/AddConstant.lua
@@ -1,20 +1,37 @@
 local AddConstant, parent = torch.class('nn.AddConstant', 'nn.Module')
 
-function AddConstant:__init(constant_scalar)
+function AddConstant:__init(constant_scalar,ip)
   parent.__init(self)
   assert(type(constant_scalar) == 'number', 'input is not scalar!')
   self.constant_scalar = constant_scalar
+  
+  -- default for inplace is false
+   self.inplace = ip or false
+   if (ip and type(ip) ~= 'boolean') then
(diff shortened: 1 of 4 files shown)


soumith:
nice! I have still not gotten around to adding docs for my in-place :)

mys007:
I want to question restoring the input in the backward pass here. This IMHO happens only in these two modules, other modules with in-place support (Threshold, Dropout, AddTable) don't seem to do this, which is inconsistent. But I'm actually not sure what is the good thing to do. With restoring I the results of FW pass can change in BW pass and without restoring no module can't rely on its `module.output` in its `updateGradInput()`...

fmassa:
There was some discussion about that in https://github.com/torch/nn/pull/235
We can only restore the output if the function is bijective, so it probably can't be done for `Threshold` and `Dropout`.
And well, wrt the results of the FW pass changing in BW pass, if you have in-place modules, you end up losing the intermediate outputs of some modules anyway, which can induce bugs so you should be careful with that, [see this issue for an example](https://github.com/szagoruyko/loadcaffe/issues/19).

mys007:
Ah, thanks for the pointer to the discussion. I agree this should be better documented:).

PR #242, opened 2015-04-23
Sequential:__len() operator added

diz-vara:
now nn.Sequential will support # operator - and who() will not crash


fmassa:
Just for information, this crash problem with `who()` has been corrected some time ago in https://github.com/torch/trepl/commit/06c2a74fef1695843f0343c251a8fd701139bf6e

soumith:
Thanks anton!

PR #245, opened 2015-04-24
Nicer error message when flattening parameters with inconsistent types.

timharley:
Also:
* Tidied up some whitespace.
* Replaced a legacy check with the newer torch.isTensor()


soumith:
Thanks tim

PR #250, opened 2015-04-28
Make type() truly recursive.

dominikgrewe:
Recursively iterate over the whole table, converting each tensor to the
given type. Removes need for many specialized type() functions.

### DIFF
--- a/ConcatTable.lua
+++ b/ConcatTable.lua
@@ -86,15 +86,6 @@ function ConcatTable:zeroGradParameters()
    end
 end
 
-function ConcatTable:type(type)
-   parent.type(self, type)
-   if torch.type(self.gradInput) == 'table' then
-      for i, gradInput in ipairs(self.gradInput) do
-         self.gradInput[i] = gradInput:type(type)
-      end
-   end
-end
-
 function ConcatTable:__tostring__()
    local tab = '  '
    local line = '\n'
--- a/Criterion.lua
+++ b/Criterion.lua
@@ -31,9 +31,7 @@ end
 function Criterion:type(type)
    -- find all tensors and convert them
    for key,param in pairs(self) do
-      if torch.typename(param) and torch.typename(param):find('torch%..+Tensor') then
-         self[key] = param:type(type)
-      end
+      self[key] = nn._utils.recursiveType(param, type)
    end
    return self
 end
--- a/CrossEntropyCriterion.lua
+++ b/CrossEntropyCriterion.lua
@@ -25,11 +25,4 @@ function CrossEntropyCriterion:updateGradInput(input, target)
    return self.gradInput
 end
 
-function CrossEntropyCriterion:type(name)
-   Criterion.type(self, name)
-   self.lsm:type(name)
-   self.nll:type(name)
-   return self
-end
-
 return nn.CrossEntropyCriterion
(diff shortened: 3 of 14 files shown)


jonathantompson:
How do we test this to make sure nothing else has broken?  It also seems like it's just a re-factoring right?  The recursiveType function is moved into utils, but that's the only real change to module.  Why move it into a separate lua source file?

dominikgrewe:
The reason I've moved it to a separate file is so that `nn.Criterion:type()` can re-use it.

jonathantompson:
Fair enough.

nicholas-leonard:
Can we make it nn.utils instead of nn._utils. I feel like we could add more functions to it. Might as well make it a public util.

koraykv:
@soumith  can you take a look at this. Because this actually modifies the recursive behavior.

soumith:
sure, will do.

soumith:
Hmmm, looking into it, this breaks multi-GPU setups. 
The big question is: Should you recursively go all the way to the leaves yourself as the root node and re-type things?

If a DataParallel module (for example nn.DataParallelTable in cunn) overrides the type() function, so that it can distribute model clones to different GPUs, calling :float() and :cuda() on it will basically make it 1-GPU.

dominikgrewe:
If a module overwrites the type() function then that should always be called. If you call it on the module directly, that's what's going to happen, isn't it? If the module is nested within another module, as long as it's in its parent's `modules` table, its type() function should get called correctly.

soumith:
model = nn.Sequential()
model:add(nn.DataParallel(...))
model:add(nn.Linear(...))
model:cuda()

nn.DataParallel's :type() has no say in this, nor is it ever called. utils.lua line 3 is called on model (which is nn.Sequential), i.e. function nn._utils.recursiveType(param, type_str).

Now that recursively converts all tensors it found in the table and it's children. It comes to model.modules[1] which is nn.DataParallel, calls another recursiveType on it, which looks at all it's tables etc.

Am I missing something? nn.DataParallel's type() is never called in the use case I've given.

dominikgrewe:
nn.Sequential (and all other container modules I believe) store their sub-modules in `self.modules`. The type() function explicitly calls type() on each element of `self.modules` (see https://github.com/torch/nn/blob/master/Module.lua#L139) and I'm not changing that.

soumith:
oh ok, you're right (sorry my bad). let me take a pass through nn to make sure no members are hard-coded to be a particular type.

dominikgrewe:
We could potentially be more general by checking for modules in recursiveType. If something is an nn.Module or a tensor, then call type on it, otherwise iterate over the table and call recursiveType. Then we don't need to rely on sub-modules being in `self.modules`.

soumith:
took a pass over nn, looks good. The only thing that came close was self.batchSize in nn.LookupTable which was a LongStorage (not a tensor ha!)

soumith:
and @dominikgrewe yes, that would be much more general, to see if it's a child of nn.Module

nicholas-leonard:
I still think that nn._utils should be nn.utils. Why make it private (i.e. `_`) ?

soumith:
agreed. nn.utils can be public.

dominikgrewe:
That's fine with me. I'll prepare a patch for that which will also use checks for `nn.Module` instead of using `self.modules`.

PR #251, opened 2015-04-28
ParallelCriterion

nicholas-leonard:
This PR adds the ParallelCriterion (formerly [SuperCriterion](https://github.com/clementfarabet/lua---nnx/blob/master/SuperCriterion.lua)). It includes unit tests and doc. It also includes unit tests for the related MultiCriterion and some doc to differentiate the two. I changed the name to make it more consistent with the naming convention of Modules.

### DIFF
--- a/MultiCriterion.lua
+++ b/MultiCriterion.lua
@@ -30,3 +30,10 @@ function MultiCriterion:updateGradInput(input, target)
    end
    return self.gradInput
 end
+
+function MultiCriterion:type(type)
+   for i,criterion in ipairs(self.criterions) do
+      criterion:type(type)
+   end
+   return parent.type(self, type)
+end
--- a/ParallelCriterion.lua
+++ b/ParallelCriterion.lua
@@ -1,0 +1,52 @@
+local ParallelCriterion, parent = torch.class('nn.ParallelCriterion', 'nn.Criterion')
+
+function ParallelCriterion:__init(repeatTarget)
+   parent.__init(self)
+   self.criterions = {}
+   self.weights = {}
+   self.gradInput = {}
+   self.repeatTarget = repeatTarget
+end
+
+function ParallelCriterion:add(criterion, weight)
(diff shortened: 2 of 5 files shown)


AjayTalati:
Cool thing to have :+1: and nice example!

Thank you :)

AjayTalati:
It there a default weighting for each criteria, of one? What I mean is in the example,

`pc = nn.ParallelCriterion():add(nll, 0.5):add(mse)`

Does this mean the `mse` criteria has a weighting of `1` or `0.5` ? Or is there a mistake in the example and it should be,

`pc = nn.ParallelCriterion():add(nll, 0.5):add(mse, 0.5)`

I'm not sure I understand?

nicholas-leonard:
Default is 1

andresy:
Hi Nicholas,

Funny, I wrote this criterion several times, and thought recently I should open source it ;)
Not sure it is a good idea to make this "if type() == 'table'". One could imagine we one to pass the same table to every criterion. In the end, it is more a flag (or another class?) which should decide if one chooses target[i] or target.

nicholas-leonard:
Hi @andresy If we want to pass same table (or tensor) to every criterion, there is MultiCriterion for that. Is this what you meant?

andresy:
I forgot about this one, but yes, indeed

nicholas-leonard:
Adding typ() unit test. Wait out.

nicholas-leonard:
Ok added type() tests.

soumith:
ready to merge. @nicholas-leonard are you keeping the non-table target logic in ParallelCriterion as well? That is not documented in the docs.

soumith:
i.e. the case where target is the same for all criteria (the case that is handled by MultiCriterion)

nicholas-leonard:
Yeah just added a `repeatTarget` to explicitly trigger this mode (target is same for all criteria). Updated doc and unit tests.

PR #253, opened 2015-04-30
small improvements in ConcatTable and ClassNLLCriterion

diz-vara:
Corrects fprop in ConcatTable when model was saved with nilling (replacing all outputs with empyTensor)
corrects the behavior of who() in presence of ClassNLLCriterion
(I do not have trepl in my Windows installation)

### DIFF
--- a/ClassNLLCriterion.lua
+++ b/ClassNLLCriterion.lua
@@ -10,6 +10,16 @@ function ClassNLLCriterion:__init(weights)
    end
 end
 
+
+function ClassNLLCriterion:__len()
+   if (self.weights) then
+      return #self.weights
+   else
+      return 0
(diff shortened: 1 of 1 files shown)


soumith:
@diz-vara what nilling modules are you using? The nilling module logic is clearly wrong. We should correct that rather than this.

soumith:
I'll merge your ClassNLLCriterion length commit, if you isolate it.

diz-vara:
I've removed ConcatTable nilling correction
That wrong nilling was based upon 'person detector' demo:
https://github.com/torch/demos/blob/master/person-detector/train.lua

soumith:
thanks anton.

PR #254, opened 2015-04-30
Add :cudaOn(device) to cuda nn modules

adamlerer:
Extend cutorch :cudaOn(device) convenience method to cunn modules.
Tests are in cunn.

### DIFF
--- a/Module.lua
+++ b/Module.lua
@@ -113,18 +113,23 @@ function Module:clone(...)
    return clone
 end
 
-local function recursiveType(param, type_str)
+local function recursiveType(param, type_str, device)
    if torch.type(param) == 'table' then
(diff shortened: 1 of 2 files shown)


soumith:
hmmm, this does not go in "nn". There was some debate on a yet unmerged PR (I haven't isolated and refactored the parts yet) that all the CUDA specific logic really belongs in cunn. And it's a fair point because we are soon getting Xeon Phi, OpenCL etc.

soumith:
The PR I referenced was: https://github.com/torch/nn/pull/149

adamlerer:
Yeah, it would be great to put this in cunn. When will that PR be merged?

PR #259, opened 2015-05-05
Check for `nn.Module` and `nn.Criterion` in recursiveType.

dominikgrewe:


utils.lua
```diff
@@ -1,10 +1,13 @@
...
+nn.utils = {}
 
-function nn._utils.recursiveType(param, type_str)
+function nn.utils.recursiveType(param, type_str)
    if torch.type(param) == 'table' then
       for k, v in pairs(param) do
-         param[k] = nn._utils.recursiveType(v, type_str)
+         param[k] = nn.utils.recursiveType(v, type_str)
       end
+   elseif torch.isTypeOf(param, 'nn.Module') or
```
nicholas-leonard: Awesome. This will also work with Criterions which encapsulate Modules.

dominikgrewe:
@andresy Note that I removed `CrossEntropyCriterion:type` again, because I don't think it's necessary any more. Hope this was okay. Let me know if not.

PR #260, opened 2015-05-05
Adding Batch L2 Normalization Layer that L2 normalizes rows of input Tensor

karpathy:
This layer L2 normalizes an `n x d` Tensor. I tested the implementation on both CPU/GPU and gradient checked it with `jac` and also with my (other) custom code. I also compared to a slower version that uses a loop, as can be seen in this [gist](https://gist.github.com/karpathy/f3ee599538ff78e1bbe9). (This batched version is quite a lot faster)

Also, I have only worked with Torch for ~2 weeks so exercise caution :)

### DIFF
--- a/L2Normalize.lua
+++ b/L2Normalize.lua
@@ -1,0 +1,40 @@
+
+--[[ 
+   This layer expects an [n x d] Tensor and normalizes each
+   row to have unit L2 norm.
+]]--
+local L2Normalize, parent = torch.class('nn.L2Normalize', 'nn.Module')
+function L2Normalize:__init()
+   parent.__init(self)
+end
+function L2Normalize:updateOutput(input)
+   assert(input:dim() == 2, 'only mini-batch supported (2D tensor), got '
+             .. input:dim() .. 'D tensor instead')
+   self.output:resizeAs(input)
+   local norms = torch.cmul(input,input):sum(2):sqrt()
+   self.output:copy(input):cdiv(norms:expandAs(input))
+   return self.output
+end
+
+function L2Normalize:updateGradInput(input, gradOutput)
+   assert(input:dim() == 2, 'only mini-batch supported')
+   assert(gradOutput:dim() == 2, 'only mini-batch supported')
+   local n = input:size(1) -- batch size
+   local d = input:size(2) -- dimensionality of vectors
+   
+   local sums = torch.sum(torch.cmul(input,input), 2):view(-1)
+   local divterms = torch.pow(sums,3/2)
+   -- compute diagonal term
+   local diag = torch.eye(d):typeAs(input):repeatTensor(n,1):view(n,d,d)
+                   :cmul(sums:view(n,1,1):expand(n,d,d))
+   -- compute cross term
+   local b1 = input:reshape(n,d,1)
+   local b2 = input:reshape(n,1,d)
+   local cross = - torch.bmm(b1,b2)
+   -- compute the local gradient of the L2 transformation
+   local dsum = torch.cdiv(diag + cross, divterms:view(n,1,1):expand(n,d,d))
+   -- chain the gradient
+   self.gradInput = torch.bmm(dsum, gradOutput:view(n,d,1))
+   return self.gradInput
+end
+
--- a/L2Normalize.lua
+++ b/L2Normalize.lua
@@ -33,8 +33,7 @@ function L2Normalize:updateGradInput(input, gradOutput)
    local b2 = input:view(n,1,d)
    self.diag:add(-torch.bmm(b1,b2))
    -- compute the local gradient of the L2 transformation
-   self.buffer:pow(3)
-   self.diag:cdiv(self.buffer:view(n,1,1):expand(n,d,d))
+   self.diag:cdiv(torch.pow(self.buffer,3):view(n,1,1):expand(n,d,d))
    -- chain the gradient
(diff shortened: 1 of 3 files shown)


soumith:
will give in-line feedback. this has some common patterns that we dont want in nn.

L2Normalize.lua
```diff
@@ -0,0 +1,40 @@
...
+]]--
+local L2Normalize, parent = torch.class('nn.L2Normalize', 'nn.Module')
+function L2Normalize:__init()
+   parent.__init(self)
+end
+function L2Normalize:updateOutput(input)
+   assert(input:dim() == 2, 'only mini-batch supported (2D tensor), got '
+             .. input:dim() .. 'D tensor instead')
+   self.output:resizeAs(input)
+   local norms = torch.cmul(input,input):sum(2):sqrt()
```
soumith: torch.xxx(...) usually does a malloc for the result tensor.
mallocs during the training loop of the nn are usually quite bad for performance.
Another reason mallocs are bad is that they introduce a synchronize point in CUDA, so you essentially kill any multi-GPU code.

instead, consider having a buffer.

self.norms  = self.norms or input.new()
self.norms:sum(self.buffer:cmul(input, input), 2):sqrt()

L2Normalize.lua
```diff
@@ -0,0 +1,40 @@
...
+   return self.output
+end
+
+function L2Normalize:updateGradInput(input, gradOutput)
+   assert(input:dim() == 2, 'only mini-batch supported')
+   assert(gradOutput:dim() == 2, 'only mini-batch supported')
+   local n = input:size(1) -- batch size
+   local d = input:size(2) -- dimensionality of vectors
+
+   local sums = torch.sum(torch.cmul(input,input), 2):view(-1)
```
soumith: you can reuse self.buffer and self.norms here for sums and divterms.

L2Normalize.lua
```diff
@@ -0,0 +1,40 @@
...
+   local divterms = torch.pow(sums,3/2)
+   -- compute diagonal term
+   local diag = torch.eye(d):typeAs(input):repeatTensor(n,1):view(n,d,d)
```
soumith: you are creating a torch.eye on every single call to backward, it can be reused

L2Normalize.lua
```diff
@@ -0,0 +1,40 @@
...
+                   :cmul(sums:view(n,1,1):expand(n,d,d))
+   -- compute cross term
+   local b1 = input:reshape(n,d,1)
```
soumith: reshape is actually an expensive operation in the sense that, reshape = malloc + copy.
What you are probably looking for is:
input:view(n,d,1). view does not create a new storage, but re-views the current storage as a different shape.

L2Normalize.lua
```diff
@@ -0,0 +1,40 @@
...
+   local b2 = input:reshape(n,1,d)
+   local cross = - torch.bmm(b1,b2)
+   -- compute the local gradient of the L2 transformation
+   local dsum = torch.cdiv(diag + cross, divterms:view(n,1,1):expand(n,d,d))
```
soumith: diag + cross is actually: result = diag + cross, meaning that result has another malloc.
since diag is not used later, you can probably do: diag:add(cross)

soumith:
overall the barrier of contributing is lower to nnx (nn-experimental) which is here: https://github.com/clementfarabet/lua---nnx

I made a bunch of comments, but hope they dont stop you from contributing in the future :)

L2Normalize.lua
karpathy: Thanks,

some of this was intentional on my part because I'm applying this in space-limited setting (I'm constantly running out of memory on GPU), so saving all of these cached intermediates (+ at every step of my LSTM) was costly. But if this is considered preferable I'll rewrite.

karpathy:
Thanks for comments! Some of these issues I was aware of and some of them were just silly mistakes (e.g. reshape) - sorry about that. I did a rewrite that should now be much more time-efficient (at cost of space and readability), but it works.

Atcold:
I think it's missing a check on the `input:dim()`. If a user send a non-batched input, the module should output the normalised single vector (see [this](https://github.com/rolfe/recpool/blob/master/NormalizeTensorL1.lua) example).

soumith:
squashed the commits and pushed to master via https://github.com/torch/nn/commit/905ea8c1a4033af2af0f90e92e16597f442f3512

karpathy:
yay!! By the way, suggested improvements going forward:

- Incorporate a smoothing term to prevent division by zero
- Allow 1-dimensional inputs as suggested by Atcold.

Atcold:
@karpathy, I would suggest 
 - a *smoothing* by default of `1e-5` like shown [here](https://github.com/torch/nn/blob/master/doc/simple.md#nn.BatchNormalization). If the user wants to set something else, then let her do it by an optional parameter `[eps]` in the module initialisation;
 - if you don't want to write new code, 1-dimensional inputs could be `view`ed as `1 x length`. So `if dim(input) == 1 then newInput = input:view(1,-1)`.

nicholas-leonard:
Some documentation would be nice.

soumith:
I totally missed that this PR doesn't have documentation. Indeed some documentation would be nice...

karpathy:
I'll add it. I have a patch for `eps` and I tried to patch the support for 1D case but that ended up taking an hour yesterday, it's not as trivial as changing the view with 1 line of code. I'll work on it.

PR #267, opened 2015-05-18
L2norm

karpathy:
Adding an optional `eps` smoothing parameter and documentation for the L2 normalization layer.

I chose the default as 1e-10; Notice that this is added to the sum of squares, so if a row is all zero then the square root will cause the effective smoothing to be 1e-5.

### DIFF
--- a/doc/simple.md
+++ b/doc/simple.md
@@ -511,10 +511,14 @@ Narrow is application of [narrow](https://github.com/torch/torch7/blob/master/do
 ## Replicate ##
 
 ```lua
-module = nn.Replicate(nFeature, dim)
+module = nn.Replicate(nFeature [, dim, ndim])
 ```
 
-This class creates an output where the input is replicated `nFeature` times along dimension `dim` (default 1). There is no memory allocation or memory copy in this module. It sets the [stride](https://github.com/torch/torch7/blob/master/doc/tensor.md#torch.Tensor.stride) along the `dim`th dimension to zero.
+This class creates an output where the input is replicated `nFeature` times along dimension `dim` (default 1). 
+There is no memory allocation or memory copy in this module. 
+It sets the [stride](https://github.com/torch/torch7/blob/master/doc/tensor.md#torch.Tensor.stride) along the `dim`th dimension to zero.
+When provided, `ndim` should specify the number of non-batch dimensions.
+This allows the module to replicate the same non-batch dimension `dim` for both batch and non-batch `inputs`.
 
 ```lua
 > x = torch.linspace(1, 5, 5)
--- a/doc/simple.md
+++ b/doc/simple.md
@@ -27,6 +27,7 @@ Simple Modules are used for various tasks like adapting Tensor methods and provi
    * [Power](#nn.Power) : an element-wise [pow](https://github.com/torch/torch7/blob/master/doc/maths.md#res-torchpowres-x) operation ;
    * [Square](#nn.Square) : an element-wise square operation ;
    * [Sqrt](#nn.Sqrt) : an element-wise [sqrt](https://github.com/torch/torch7/blob/master/doc/maths.md#res-torchsqrtres-x) operation ;
+   * [L2Normalize](#nn.l2norm) : normalizes each row of input to have unit L2 norm ;
    * [MM](#nn.MM) : matrix-matrix multiplication (also supports batches of matrices) ;
  * Miscellaneous Modules :
    * [BatchNormalization](#nn.BatchNormalization) - mean/std normalization over the mini-batch inputs (with an optional affine transform) ;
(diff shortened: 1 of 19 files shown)


doc/simple.md
```diff
@@ -863,6 +864,22 @@ gnuplot.grid(true)
...
+```
+
+Normalizes each row of an input 2D Tensor to have unit L2 norm. The smoothing parameter `eps` prevents division by zero when a row contains all zero elements (default = `1e-10`).
+
+```lua
+A = torch.randn(3, 5)
+m = nn.L2Normalize()
+B = m:forward(A) -- B is also 3 x 5
+-- take the L2 norm over the second axis:
+norms = torch.norm(B, 2, 2) -- norms is [1, 1, 1]
```
nicholas-leonard:
```
>print(torch.norm(B,2,2))
```

nicholas-leonard:
Not sure if its too late to change the name, but `nn.Norm(dim, val, eps)` would be a more general name and module fitting nicely with `torch.norm`.

karpathy:
I think the point about `nn.Norm` is a good one, since now that I look at it big chunks of `nn` are already nicely consistent and parallel to torch commands.

One thing to consider is that for special cases such as `val = 2,1` that are most commonly used there is a  very efficient implementation. I also have an `nn.L1normalize` written up that is probably the most efficient way, and it looks difference from `nn.L2normalize`. However, these special cases could be dispatched separately and hidden neatly inside the `nn.Norm`.

The problem is that I personally cannot commit time to implement the full general `nn.Norm` function. I could maybe implement the `nn.Norm` and only make it work for the special cases `val = 1,2` or assert fail otherwise.

Based on this perhaps it's best to remove the original `L2Normalize` commit and work towards the more general `nn.Norm` layer somehow.

soumith:
@karpathy If you implement nn.Norm for norm=1 and 2, and assert otherwise, I can find someone to implement the other. I can revert the L2Normalize layer immediately in trunk, it just got in last week, so no one knows about it.

soumith:
Reverted in https://github.com/torch/nn/pull/268

bamos:
Hi, I'm interested in using this normalization patch and having it merged into the master branch.

@karpathy - I can merge your l1 and l2 norm implementations into a new `nn.Norm` and assert false for other norms. I can write additional tests if needed. Have you posted your l1 normalization code somewhere?

Also, I'm interested in using the l2 norm with CUDA and am happy to help with the implementation. Should I file an issue in `torch/cunn` to track this?

-Brandon.

fmassa:
@bamos I was also thinking about finishing this module to accept other norms, as well as non-batch mode. Current implementation is entirely on lua, so it'll work for cuda as well.

karpathy:
My code for L1 normalization layer:

```lua
--[[
L1 normalize rows of input Tensor
--]]

local L1Normalize, parent = torch.class('nn.L1Normalize', 'nn.Module')

function L1Normalize:__init()
  parent.__init(self)
end

function L1Normalize:updateOutput(input)
  self.output:resizeAs(input)
  self.buffer:sum(input, 2)
  self.output:copy(input):cdiv(self.buffer:expandAs(input))
  return self.output
end

function L1Normalize:updateGradInput(input, gradOutput)
  local size = input:size(1)
  local n = input:size(1) -- batch size
  local d = input:size(2) -- dimensionality of vectors
  -- compute diagonal term
  self.eye = self.eye or torch.eye(d):typeAs(input):repeatTensor(n,1):view(n,d,d)
  self.diag = self.diag or self.eye.new()
  self.diag:cmul(self.eye, self.buffer:view(n,1,1):expand(n,d,d))
  -- compute the cross term
  self.diag:add(-input:view(n,1,d):expand(n,d,d))
  -- compute the local gradient of the L2 transformation
  self.diag:cdiv(torch.pow(self.buffer,2):view(n,1,1):expand(n,d,d))
  -- chain the gradient
  self.gradInput:resize(n,d,1):bmm(self.diag, gradOutput:view(n,d,1)):resize(n,d)
  return self.gradInput
end
```

make sure to double check this one with edge cases - I remember it gave the right answers and gradient checked for NxD matrix, but I didn't put it through a comprehensive barrage of tests, like maybe Nx1 inputs, etc.

fmassa:
@karpathy Thanks for the L1 code! I think there should be an `abs` in the forward pass though (it might have passed the tests if you used it after a ReLU layer, or using positive random numbers).
@bamos do you want to collaborate on that ? I'll probably have some time to look into it think weekend.

bamos:
@fmassa - sure! Since I'm new to Torch, can you take a first pass and leave stubs/TODOs for me to fill in? Even things you might consider tedious. I can pick up on Monday or whenever you're done.

karpathy:
@fmassa sorry about the `abs` you're right. I was planning to use the code with positive inputs to (as a most common use case I assume) normalize probabilities to sum to 1, which is probably why this slipped through.

bamos:
Hi @fmassa and @karpathy 

Thanks for the quick responses here back in July and
for quickly merging it into the master branch.
I've been pulling the latest versions from @fmassa and
just released a project with it as a dependency:
https://github.com/cmusatyalab/openface

-Brandon.

soumith:
closing because redundant after @fmassa 's nn.Normalize

PR #269, opened 2015-05-20
Remove unused and expensive init logic from nn.SpatialConvolutionMap.

georgevdd:


koraykv:
This function was necessary because cunn version was using it, but cunn/SpatialConvolutionMap was removed since it was not widely used and was an inefficient implementation

PR #270, opened 2015-05-20
Speed up LookupTable

colesbury:
Adds an efficient CPU implementation of LookupTable. The forward and
backward passes are implemented in C, which makes it easier to add
an efficient CUDA implementation in cunn.

I've added a benchmark here:
https://gist.github.com/colesbury/442c3b6f1ac73cf52036

Thanks to David Grangier for the FasterLookupTable on which it's based.

### DIFF
--- a/LookupTable.lua
+++ b/LookupTable.lua
@@ -2,156 +2,75 @@ local LookupTable, parent = torch.class('nn.LookupTable', 'nn.Module')
 
 LookupTable.__version = 3
 
-function LookupTable:__init(nIndex, ...)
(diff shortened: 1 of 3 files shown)


soumith:
nice, it looks to be 4x - 40x faster!

Merging this would break CUDA, so will wait for a CUDA-side PR to merge both at once.

colesbury:
See https://github.com/torch/cunn/pull/90 for CUDA version
See https://github.com/torch/torch7/pull/235 for change to tensor:index to speed up forward pass

PR #271, opened 2015-05-21
MixtureTable lazy initialized buffers

nicholas-leonard:
Also added `nn.utils.recursiveResizeAs`.


soumith:
thanks @nicholas-leonard

PR #273, opened 2015-05-23
moved zeroing to blas side in batch nn.Linear:forward

szagoruyko:
removes "fill" calls because they can be done in blas. gives nice speedup for small matrices on CUDA, doesn't hurt cpu.


soumith:
nice.

PR #276, opened 2015-05-26
Add more useful string description of layers.

orian:
Adds:
 - pbb in Dropout and SpatialDropout
 - non default params in SpatialConvolution, SpatialConvonlutionMM, SpatialMaxPooling

### DIFF
--- a/Dropout.lua
+++ b/Dropout.lua
@@ -40,3 +40,7 @@ end
 function Dropout:setp(p)
    self.p = p
 end
+
+function Dropout:__tostring__()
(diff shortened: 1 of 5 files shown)


soumith:
thank you! that looks really helpful.

szagoruyko:
just my opinion on this, it is very cluttering. Seems to me that it would be right to have two separate prints, the one we had before "minimalistic" and the new one which prints all the information about each layer.
or at least ```nn.SpatialConvolution(in: 2, out: 96, kW: 5, kH: 5)``` should become ```nn.SpatialConvolution(2, 96, 5, 5)``` etc.

PR #277, opened 2015-05-27
Removed special case for nunit = 1 in Linear

szagoruyko:
As addmm was fixed the special case in Linear is not needed (was added here https://github.com/torch/nn/pull/25). There is a negligible numerical difference with CUBLAS between the general and special case.


soumith:
if you say so :)

PR #278, opened 2015-05-28
Volumetric Average Pooling + doc + unit test, better performance for Volumetric Max Pooling

soumith:


soumith:
unit tested. docs. standard stuff. prepping for cunn PR.

PR #284, opened 2015-06-03
Fix View to accept non-contiguous input and gradOutput

fmassa:
As `torch.view` requires contiguous data, `nn.View` didn't work for non-contigous input or gradOutput.

BTW, with this commit, `nn.Reshape` and `nn.View` behave in a similar manner (except for the interface of `nn.Reshape`, which is less generic than `nn.View`).
In fact, since commit https://github.com/torch/nn/commit/75568888fb18158925c7953729f29f90778059ab, `nn.Reshape` uses `torch.view` internally, thus avoiding memory copy when the input is contiguous.

### DIFF
--- a/View.lua
+++ b/View.lua
@@ -24,6 +24,10 @@ function View:__init(...)
    self.output = nil
    self.gradInput = nil
    self.numInputDims = nil
+   
+   -- only used for non-contiguous input or gradOutput
+   self._input = torch.Tensor()
+   self._gradOutput = torch.Tensor()
 end
 
 function View:setNumInputDims(numInputDims)
@@ -72,6 +76,12 @@ local function batchsize(input, size, numInputDims, numElements)
 end
 
 function View:updateOutput(input)
+   if not input:isContiguous() then
+      self._input = self._input or input.new()
+      self._input:resizeAs(input)
+      self._input:copy(input)
+      input = self._input
+   end
    local bsz = batchsize(input, self.size, self.numInputDims, self.numElements)
    if bsz then
       self.output = input:view(bsz, table.unpack(self.size:totable()))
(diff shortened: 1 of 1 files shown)


szagoruyko:
also I would expect nn.View behave as :view() and nn.Reshape as :reshape(), otherwise what it is the point of having nn.Reshape?

soumith:
What @szagoruyko said is correct. nn.Reshape's behavior today (with zero-copy when possible) was if I remember, introduced in the small gap between torch.view being merged and nn.View came into existence. We should keep the operators like these as consistent as possible with their torch equivalents.

nicholas-leonard:
Yeah but it is still more efficient to have nn.Reshape torch.view when
possible as the torch.reshape automatically copies the tensor.

Nicholas Léonard
917-741-7570

fmassa:
@szagoruyko I've updated the commit (even though it didn't break old models because there couldn't be any old models supporting non-contiguous data).
Anyway, I think that one of the great things of `nn.View`  is that you don't need to specify the output size, giving `-1` and setting `numInputDims` is enough, which is not possible in `nn.Reshape`. Thus I think allowing `nn.View` to accept non-contiguous data is still valid.

soumith:
@fmassa I actually disagree with you. nn.View should be the equivalent of torch.view.
If anything, we sould revert nn.Reshape to the same as torch.reshape, but errrr nn.Reshape was already changed....and most people are largely oblivious to nn.Reshape's and use it as a default of sorts.

I am closing this PR to clean up the queue of pending PRs and if any of the maintainers thinks I am wrong, feel free to reopen/merge.

fmassa:
@soumith I think the goal of this PR was misinterpreted (probably due to my last paragraph in the first message). I just thought that it was worthy to mention that nn.Reshape was practically equivalent to nn.View, just less generic. If I understand correctly, even torch.reshape requires contigous input for it to work properly, but nn.Reshapes works just fine with non-contiguous data. So why can't nn.View also accept non-contiguous data ?

PR #285, opened 2015-06-03
Add support for negative indices in nn.SplitTable

sergomezcol:
This module can now be used when the total number of dimensions is unknown.


nicholas-leonard:
Nice. It would be nice to have the main torch support negative indexing as well.

GeorgOstrovski:
This one would be quite nice. ping @soumith

soumith:
sorry for the delay, it took me quite some time to wrap my head around this. I thought it was more complicated than it is.

PR #286, opened 2015-06-06
SparseLinear.c: fix missing OpenMP include

deltheil:
This fixes the `omp_get_thread_num` implicit declaration warning.


soumith:
thanks!

PR #290, opened 2015-06-17
set self.train=[true,false] for containers

nicholas-leonard:
The train attribute of containers should also be set to true or false (not just non-container modules).


soumith:
:+1:

PR #292, opened 2015-06-17
Add docs for the SplitTable and JoinTable modules

rfaulkner:
I had to reverse engineer these modules to understand exactly what they were doing in context so I figured I'd add a simple description and example for clarity.


fmassa:
https://github.com/torch/nn/blob/master/doc/table.md#nn.JoinTable ?

soumith:
As francisco pointed out, we have pretty good docs separately in the doc folder.

rfaulkner:
oh!  sorry i missed that those are brilliant, thanks for pointing me to them.

PR #293, opened 2015-06-17
SpatialConvolutionMM supports padW/padH != 1

fmassa:
This commit allows different paddings for each spatial dimension.

A PR with the Cuda version will come tomorrow.

I changed the print function a bit, as it was calling a `self.pad_h` and `self.pad_w` that, to the best of my knowledge, didn't exist anywhere (`cudnn` uses `padH` and `padW`). I think though that the current print version is ambiguous when `pad !=0` and `dW=dH=1`, or `pad=0` and `dW!=1 or dH!=1`, but I decided not to change it in this commit.

### DIFF
--- a/SpatialConvolution.lua
+++ b/SpatialConvolution.lua
@@ -1,6 +1,6 @@
 local SpatialConvolution, parent = torch.class('nn.SpatialConvolution', 'nn.Module')
 
-function SpatialConvolution:__init(nInputPlane, nOutputPlane, kW, kH, dW, dH, padding)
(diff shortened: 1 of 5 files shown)


nagadomi:
This PR breaks backward compatible of model file that trained with older version.

soumith:
Will look into it asap, what's the error you get? The PR tries to keep back comparability, but maybe i missed something

nagadomi:
Sorry I forget to update cunn.
It was my misunderstanding.

PR #294, opened 2015-06-18
Added a Batch version HingeEMbeddingCriterion

jakezhaojb:
Previous HingeEmbeddingCriterion can't handle Tensors as input. This file does that job


soumith:
@jakezhaojb what you want to do is make HingeEmbeddingCriterion (the existing one) support either batches or non-batches. Having a separate class for just the batch case is unnecessary and is not the approach we take for all our other nn modules.

Can you integrate this into nn.HingeEmbeddingCriterion

Also, unit tests + documentation is a must for the nn repository because it is widely used.

PR #295, opened 2015-06-18
LookupTable makeInputContiguous only once

nicholas-leonard:
Incremented the version attribute for the major changes introduced in the last PR related to this module. And makes the input contiguous once in updateOutput, instead of both updateOutput and accGradParameters. Unit tests pass.


soumith:
thanks nick!

PR #296, opened 2015-06-19
Fix print and correct non-contiguous test in SpatialConvolutionMM

fmassa:
@szagoruyko pointed that sometimes the tests for `SpatialConvolutionMM` were failing, with error
`SpatialConvolutionMM$ Error: matrices expected, got 2D, 1D tensors at /opt/rocks/torch7/lib/TH/generic/THTensorMath.c:594`.
There was an inversion of the dimensions while defining the non-contiguous tensor, thus some tests fed invalid input sizes.
Do you think we should we assert that proper-sized input tensors are given to the forward pass ?
Plus, small fix in the `__tostring__` methods.

### DIFF
--- a/SpatialConvolution.lua
+++ b/SpatialConvolution.lua
@@ -137,7 +137,7 @@ function SpatialConvolution:__tostring__()
    if self.padding and self.padding ~= 0 then
      s = s .. ', ' .. self.padding .. ',' .. self.padding
    elseif (self.padW or self.padH) and (self.padW ~= 0 or self.padH ~= 0) then
-     s = s .. ', ' .. self.padW .. ',' .. self.padW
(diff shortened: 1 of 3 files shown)


soumith:
we already assert them at the C level I thought...

fmassa:
I just checked, both `nn` and `cunn` do not check if the input dimension is valid before doing the matrix multiplication. The error is throw in the blas level.

fmassa:
By dimension I mean the size of the input, not the number of dimensions.

soumith:
ah ok, yes we do not check it seems...yes it would be good to check at C level and spit out a more clear error.

szagoruyko:
have done several thousand runs of the test, the issue is gone.

PR #297, opened 2015-06-19
Assert output size in SpatialConvolutionMM + doc fix

fmassa:


generic/SpatialConvolutionMM.c
```diff
@@ -194,6 +194,9 @@ static int nn_(SpatialConvolutionMM_updateOutput)(lua_State *L)
   outputWidth  = (inputWidth + 2*padW - kW) / dW + 1;
   outputHeight = (inputHeight + 2*padH - kH) / dH + 1;
 
+  if (outputWidth < 1 || outputHeight < 1)
+    THError("Output size is too small");
```
soumith: Can you print out a more informative error, like: give input size: 128x32x16x16. Calculated output size: 128x64x0x0. Output size is too small.

fmassa: Done! I'm usually too verbose, but maybe this time I was not specific enough :)

soumith:
this is great! thanks.

PR #300, opened 2015-06-21
CosineDistance supports batch mode

fmassa:
I tried to keep backward compatibility with saved models.
Also, added a test, which was missing.

### DIFF
--- a/CosineDistance.lua
+++ b/CosineDistance.lua
@@ -12,28 +12,33 @@ function CosineDistance:updateOutput(input)
       input1 = input1:view(1,-1)
       input2 = input2:view(1,-1)
    end
-   
+
    if not self.buffer then
       self.buffer = input1.new()
       self.w1  = input1.new()
       self.w22 = input1.new()
-      self.w2  = input1.new()
+      self.w  = input1.new()
       self.w32 = input1.new()
-      self.w3  = input1.new()
+      self.ones = input1.new()
    end
- 
+
    self.buffer:cmul(input1,input2)
    self.w1:sum(self.buffer,2)
 
+   local epsilon = 1e-12
    self.buffer:cmul(input1,input1)
-   self.w22:sum(self.buffer,2)
-   self.w2:sqrt(self.w22)
+   self.w22:sum(self.buffer,2):add(epsilon)
+   self.ones:resizeAs(self.w22):fill(1)
+   self.w22:cdiv(self.ones, self.w22)
+   self.w:resizeAs(self.w22):copy(self.w22)
 
    self.buffer:cmul(input2,input2)
-   self.w32:sum(self.buffer,2)
-   self.w3:sqrt(self.w32)
+   self.w32:sum(self.buffer,2):add(epsilon)
+   self.w32:cdiv(self.ones, self.w32)
+   self.w:cmul(self.w32)
+   self.w:sqrt()
 
-   self.output:cdiv(self.w1,self.w2):cdiv(self.w3)
+   self.output:cmul(self.w1,self.w)
    self.output = self.output:select(2,1)
 
    return self.output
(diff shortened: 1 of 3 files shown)


soumith:
hi, did you test this on anything real-world? Does it work without side-effects? I reviewed it and I think it looks good, but I also dont fully trust my code reviews. thanks for the Unit test.

fmassa:
I'm testing it now.
The behaviour is equivalent to the previous version of `CosineDistance`, but it's sensitive to zero inputs, generating nans (as before BTW).
I'll fix that and let you know.

nkoumchatzky:
This PR is definitely a good idea. However, we could improve the performance by:
- replacing a bunch of costly divisions by pre-computing then and using multiplications afterwards
- Factorize the two sqrt computations into one

Proposed code for updateOutput and updateGradInput (tested locally with your test.lua addition):
```
function CosineDistance:updateOutput(input)
   local input1, input2 = input[1], input[2]

   if input1:dim() == 1 then
      input1 = input1:view(1,-1)
      input2 = input2:view(1,-1)
   end

   if not self.buffer then
      self.buffer = input1.new()
      self.w1  = input1.new()
      self.w22 = input1.new()
      self.w  = input1.new()
      self.w32 = input1.new()
      self.ones = input1.new()
   end

   self.buffer:cmul(input1,input2)
   self.w1:sum(self.buffer,2)

   self.buffer:cmul(input1,input1)
   self.w22:sum(self.buffer,2)
   self.ones:resizeAs(self.w22):fill(1)
   self.w22:cdiv(self.ones, self.w22)
   self.w:resizeAs(self.w22):copy(self.w22)

   self.buffer:cmul(input2,input2)
   self.w32:sum(self.buffer,2)
   self.w32:cdiv(self.ones, self.w32)
   self.w:cmul(self.w32)
   self.w:sqrt()

   self.output:cmul(self.w1,self.w)
   self.output = self.output:select(2,1)

   return self.output
end

function CosineDistance:updateGradInput(input, gradOutput)
   local v1  = input[1]
   local v2  = input[2]
   local not_batch = false

   if v1:dim() == 1 then
      v1 = v1:view(1,-1)
      v2 = v2:view(1,-1)
      not_batch = true
   end

   local gw1 = self.gradInput[1]
   local gw2 = self.gradInput[2]
   gw1:resizeAs(v1):copy(v2)
   gw2:resizeAs(v1):copy(v1)

   self.w = self.w:expandAs(v1)
   self.buffer:cmul(self.w1,self.w22)
   self.buffer = self.buffer:expandAs(v1)
   gw1:addcmul(-1,self.buffer,v1)
   gw1:cmul(self.w)

   self.buffer:cmul(self.w1,self.w32)
   self.buffer = self.buffer:expandAs(v1)
   gw2:addcmul(-1,self.buffer,v2)
   gw2:cmul(self.w)

   local go = gradOutput:view(-1,1):expandAs(v1)
   gw1:cmul(go)
   gw2:cmul(go)

   if not_batch then
      self.gradInput[1] = gw1:select(1,1)
      self.gradInput[2] = gw2:select(1,1)
   end

   return self.gradInput
end
```

fmassa:
Thanks for the optimizations @nkoumchatzky ! I've added them to the PR.
There are two more changes in the new commit compared to the previous version:
- add a small constant to avoid nans if the input tensor is zero
- a fix for torch bug https://github.com/torch/torch7/issues/289 . After the first backward pass, `self.buffer` had zero stride and the same size as the inputs, so the following forward (and backward) passes were not being computed properly.

I've used a slightly different version of this fix in a project and it works. I'll double check this new version tomorrow and squash the commits.

fmassa:
@soutmith I've tested the PR in a project and it works without problem. There is a small hack in the end of `updateGradInput` to bypass https://github.com/torch/torch7/issues/289 . But if PR https://github.com/torch/torch7/pull/295 gets merged, I can remove it.

soumith:
thanks!

Issue #187, opened 2015-03-11
Preserve sharing semantics when typecasting a network [for example to :cuda()]


soumith:
Currently, if one creates a network (say a siamese network) where some modules in the network share weights with one another, then once the network is typed to another type like :cuda or :float, then the sharing is untied.

Not only is this confusing, but also undocumented.

It is also fixable pretty simply.

I will prepare a patch for this.

jonathantompson:
Nice! I've always had to rebuild them from scratch.

nicholas-leonard:
Plus one for this. I currently need it.

soumith:
added documentation: https://github.com/torch/nn/commit/de9418dada8a837c231c267a3ce9273f36072170

nicholas-leonard:
I am working on an actual implementation that fixes this.

nicholas-leonard:
https://github.com/nicholas-leonard/rnn/blob/master/Module.lua#L76

soumith:
that looks pretty good to me, can you do a PR for nn, to have it be the default behavior of :type()

nicholas-leonard:
Yeah I can PR. Only issue with making default is all the modules that overload type. They won't pass around the castmap. For example, see Euclidean.

AjayTalati:
nicholas-leonard , soumith this sounds cool :+1:  - will it be easy to understand, for fools like me?

I'm still trying to understand this, [ model_utils.clone_many_times(net, T) ](https://github.com/AjayTalati/practical6/blob/master/model_utils.lua) ? 

Are there are easy papers or reviews of this? I should understand this code, but I don't?

Atcold:
Cool, just tumbled into this same problem. In my case the untieing is caused by issuing `:getParameters()` in order to feed `optim`.

dominikgrewe:
Could we have a multi-pass solution that works with custom implementations of `type`?

1) Traverse modules to look for shared storages.
2) Do the things `type` currently does.
3) If storage sharing was detected in step 1, restore it.

Steps 1 and 3 would only be run on the top-level module. `type` could do this automatically by setting a flag that says whether the current module is the top-level module or not.
The only complication would be that container modules with a custom type method would have to re-implement this logic. Wdyt?

nicholas-leonard:
@dominikgrewe This is how I do it at the moment : https://github.com/nicholas-leonard/dpnn/blob/master/Module.lua#L100

Atcold:
[...]

PR #303, opened 2015-06-24
nn.Module preserve type sharing semantics (#187); add nn.Module.apply

adamlerer:
This PR contains two features:

1) Preserve sharing semantics when typecasting a network (https://github.com/torch/nn/issues/187). Shared references to tensors, as well as tensors sharing the same underlying storage, are preserved. 
The new :type() function typecasts the entire storage rather than just the tensor view, so it's possible to waste memory if you typecast a tensor that's a view on a small part of a much larger storage. Suggestions welcome.

2) Add `nn.Module:apply()` function similar to nngraph. (nngraph's `gModule:apply()` will be subsumed by `nn.Module:apply()`).

### DIFF
--- a/Module.lua
+++ b/Module.lua
@@ -116,10 +116,11 @@ end
 function Module:type(type, tensorCache)
    assert(type, 'Module: must provide a type to convert to')
 
+   tensorCache = tensorCache or {}
+
    -- find all tensors and convert them
    for key,param in pairs(self) do
-      self[key] = nn.utils.recursiveType(param, type)
-
+      self[key] = nn.utils.recursiveType(param, type, tensorCache)
    end
 
    return self
@@ -279,6 +280,19 @@ function Module:__call__(input, gradOutput)
    else
       return self.output
    end
+end
+
+-- Run a callback (called with the module as an argument) in preorder over this
+-- module and its children.
+--
+function Module:apply(callback)
+    callback(self)
+
+    if self.modules then
+        for _, module in ipairs(self.modules) do
+            module:apply(callback)
+        end
+    end
 end
 
 function Module:findModules(typename, container)
--- a/CMul.lua
+++ b/CMul.lua
@@ -114,7 +114,7 @@ function CMul:accGradParameters(input, gradOutput, scale)
    end
 end
 
-function CMul:type(type)
+function CMul:type(type, tensorCache)
    if type then
       self._input = nil
       self._output = nil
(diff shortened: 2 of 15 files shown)


soumith:
@clementfarabet @koraykv have a look. I think it's a really nice PR that finally gets missing functionality when weights etc. inside a network are shared, and you typecast to :cuda() for example.

This also provides a mechanism for typecasting two separate networks (by passing in a common table to the :type function), preserving their sharing after the typecast.

@nicholas-leonard @dominikgrewe @szagoruyko

adamlerer:
@soumith the recommended way to share two separate networks (in the README) is:

```lua
nn.utils.recursiveType({net1, net2}, 'torch.FloatTensor')
```
although you can also do

```lua
tensorCache = {}
net1:type('torch.FloatTensor', tensorCache)
net2:type('torch.FloatTensor', tensorCache)
```

andresy:
thanks guys! i would not rely on torch.cdata() though (trying to keep away from ffi for now). instead of tostring(torch.cdata()) you could use torch.pointer().

soumith:
@andresy good catch. torch.pointer is the better way to go, to keep pure-lua.

andresy:
i am pretty sure you put that in there to be sure i read the pr.

adamlerer:
Thanks, I switched to `torch.pointer()`

soumith:
I will let this sit for another 2 weeks. With ICML, everyone disappeared, and I want us to test the stability of this stuff a little more before merging to main line.

Module.lua
```diff
@@ -113,12 +113,16 @@ function Module:clone(...)
    return clone
 end
 
-function Module:type(type)
-   assert(type, 'Module: must provide a type to convert to')
```
fidlej: The assert would be useful to detect mistakes early.

fidlej:
I like the proposed solution. It allows to preserve the sharing and avoids breaking existing code.

Module.lua
adamlerer: will fix

adamlerer:
@soumith time to merge this?

soumith:
yea sure. rebase and address Ivo's comment.

soumith:
well...., after I merged your #363 , this is conflicting again. sorry!

adamlerer:
@soumith rebased again. Thanks!

soumith:
thanks adam. i'm glad this is finally in!

PR #306, opened 2015-07-01
Add unit tests for hessian.lua, fix bugs detected by the tests

erosennin:
This PR adds unit tests for the Hessian code and fixes several bugs found during testing:

1. Fix initialization of `diagHessianBias` for `nn.SpatialConvolution`.

2. Fix computing `diagHessianBias` for `nn.SpatialFullConvolution`.

3. Call `module:forward()` with the proper input before calling `accGradParameters()`.
Without that, `accDiagHessianParameters()` produces incorrect results for several convolution classes, which leads to weird bugs like this: https://github.com/koraykv/unsup/issues/30#issuecomment-112900751

4. Factor out duplicate code from `Module.getParameters()`.
The replacement `Module.getParameters()` function in `hessian.lua` containes a chunk of outdated code copied from `Module.lua` three years ago. This code has not been kept in sync with original code in `Module.lua` and fails unit tests. This patch moves the duplicate code to `Module.flatten()`, which is now used by both the original `Module.getParameters()` in `Module.lua` and the replacement `Module.getParameters()` in `hessian.lua`.

### DIFF
--- a/Jacobian.lua
+++ b/Jacobian.lua
@@ -92,6 +92,110 @@ function nn.Jacobian.forward(module, input, param, perturbation)
    return jacobian
 end
 
+function nn.Jacobian.backwardDiagHessian(module, input, diagHessianParamName)
+   -- Compute the second derivatives (diagonal Hessian elements)
+   -- by backpropagation (using the code from hessian.lua).
+   --
+   -- This function computes the diagonal Hessian elements of the following function:
+   --
+   -- F(x_1, x_2, ..., x_n) = y_1^2/2 + y_2^2/2 + ... + y_m^2/2,
+   --
+   -- where
+   -- x_1, ..., x_n are the input values and parameters of the given module,
+   -- y_1, ..., y_m are the output values of the given module.
+   --
+   -- All x_i and y_i values are scalars here. In other words,
+   -- x_1, ..., x_n denote the scalar elements of the module input tensor,
+   --             the scalar elements of module.weight,
+   --             and the scalar elements of module.bias;
+   -- y_1, ..., y_m are the scalar elements of the module output tensor.
+   --
+   -- The diagonal Hessian elements of F are computed with respect to
+   -- the module input values and parameters (x_1, .., x_n).
+   --
+   -- The function F is chosen for its convenient properties:
+   --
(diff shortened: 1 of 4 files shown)


erosennin:
Any feedback? :)

If the pull request is too large, I can split in into smaller ones.

soumith:
Hi, the PR is not large, but it does not seem to be against trunk (if you see the changes to Module.lua, for example getParameters etc.). Can you rebase your commits to be on top of trunk?

erosennin:
Seems already on top on trunk to me. The last commit in trunk is b7aa53d96fbb6c0f2eaa1976b28c5cf12edf1ced from 19 days ago, right?

soumith:
oh nevermind, duh! i see what you did there. thanks. taking a look at the PR now.

erosennin:
Oh, I get it now, you must be referring to the getParamsByDevice branch. Let's wait until it's merged then.

Jacobian.lua
```diff
@@ -92,6 +92,110 @@ function nn.Jacobian.forward(module, input, param, perturbation)
...
+   -- In other words, the diagonal Hessian elements of F with respect
+   -- to the module OUTPUT values (y_1, ... y_m) are equal to 1.
+   --
+   -- Because of that, computing the diagonal Hessian elements of F
+   -- with respect to the module INPUT values and PARAMETERS (x_1, ..., x_n)
+   -- can be done by calling updateDiagHessianInput() and accDiagHessianParameters()
+   -- using a tensor of ones as diagHessianOutput.
+
+   module:forward(input)
+   local diagHessianOutput = module.output.new():resizeAs(module.output):fill(1)
```
soumith: why not just directly fill the output with 1:
local diagHessianOutput = module.output:fill(1)

not sure why you are creating a new output tensor etc...

soumith: uh, nevermind. the diaghessianoutput is equal to one, not the output itself.

soumith:
this PR looks good to go. If you squash your commits I can merge. cc: @adamlerer who has a PR in flight around this code.

erosennin:
OK, done.

soumith:
thanks @erosennin

PR #309, opened 2015-07-07
SpatialMaxPooling supports padding and ceil mode

fmassa:
- Adds padding
- adds ceil mode similar to `cudnn`, allowing `loadcaffe` to run on CPU
- changes the way the max indexes are stored, saving memory and computation time. For huge tensors we could have problems in the backward pass though, as we are storing the indices in a `Tensor` and not `LongTensor`.

@szagoruyko is preparing a PR for `cunn` with the same changes.


soumith:
thanks francisco!

PR #313, opened 2015-07-12
HingeEmbeddingCriterion: batch mode

mys007:
Rewrote HingeEmbeddingCriterion to support batch mode. Is compatible with the original version (accepting non-tensor y as well).

### DIFF
--- a/HingeEmbeddingCriterion.lua
+++ b/HingeEmbeddingCriterion.lua
@@ -2,24 +2,42 @@ local HingeEmbeddingCriterion, parent = torch.class('nn.HingeEmbeddingCriterion'
 
 function HingeEmbeddingCriterion:__init(margin)
    parent.__init(self)
-   margin = margin or 1 
-   self.margin = margin 
-   self.gradInput = torch.Tensor(1)
+   self.margin = margin or 1
+   self.sizeAverage = true
 end 
  
 function HingeEmbeddingCriterion:updateOutput(input,y)
-   self.output = input[1]
-   if y == -1 then
-	   self.output = math.max(0,self.margin - self.output);
+   self.buffer = self.buffer or input.new()
+   if not torch.isTensor(y) then 
+      self.ty = self.ty or input.new():resize(1)
+      self.ty[1]=y
+      y=self.ty
    end
+
+   self.buffer:resizeAs(input):copy(input)
+   self.buffer[torch.eq(y, -1)] = 0
+   self.output = self.buffer:sum()
+   
+   self.buffer:fill(self.margin):add(-1, input)
+   self.buffer:cmax(0)
+   self.buffer[torch.eq(y, 1)] = 0
+   self.output = self.output + self.buffer:sum()
+   
+   if (self.sizeAverage == nil or self.sizeAverage == true) then 
+      self.output = self.output / input:nElement()
+   end
+
    return self.output
 end
 
 function HingeEmbeddingCriterion:updateGradInput(input, y)
-  self.gradInput[1] = y
-  local dist = input[1]
-  if y == -1 and  dist > self.margin then
-     self.gradInput[1] = 0;
-  end
-  return self.gradInput 
+   if not torch.isTensor(y) then self.ty[1]=y; y=self.ty end
+   self.gradInput:resizeAs(input):copy(y)
+   self.gradInput[torch.cmul(torch.eq(y, -1), torch.gt(input, self.margin))] = 0
+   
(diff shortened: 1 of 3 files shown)


HingeEmbeddingCriterion.lua
```diff
@@ -2,24 +2,38 @@ local HingeEmbeddingCriterion, parent = torch.class('nn.HingeEmbeddingCriterion'
...
  
 function HingeEmbeddingCriterion:updateOutput(input,y)
-   self.output = input[1]
-   if y == -1 then
-	   self.output = math.max(0,self.margin - self.output);
+   self.buffer = self.buffer or input.new()
+   if not torch.isTensor(y) then self.ty = self.ty or input.new():resize(1); self.ty[1]=y; y=self.ty end
+
+   self.buffer:resizeAs(input):copy(input)
+   self.buffer[torch.eq(y, -1)] = 0
```
fmassa: Martin, you are allocating new memory each time you call `torch.eq`. Also, there is a PR from @dominikgrewe on torch which adds element-wise max, which could be usefull in here.

mys007: Thanks for the comment, your right. This assumes that the memory manager does a bad job though, is it really like that? Nevertheless, do you know how to change that without using 2 more buffers?

HingeEmbeddingCriterion.lua
```diff
@@ -2,24 +2,38 @@ local HingeEmbeddingCriterion, parent = torch.class('nn.HingeEmbeddingCriterion'
...
+   self.margin = margin or 1
+   self.sizeAverage = true
 end 
  
 function HingeEmbeddingCriterion:updateOutput(input,y)
-   self.output = input[1]
-   if y == -1 then
-	   self.output = math.max(0,self.margin - self.output);
+   self.buffer = self.buffer or input.new()
+   if not torch.isTensor(y) then self.ty = self.ty or input.new():resize(1); self.ty[1]=y; y=self.ty end
```
soumith: can you please expand this, compacting 3 lines together is somewhat unreadable. we have one-line compaction all over the place, and that is reasonable.

if not torch.isTensor(y) then 
      self.ty = self.ty or input.new():resize(1); 
      self.ty[1]=y; 
      y=self.ty 
   end

HingeEmbeddingCriterion.lua
soumith: element-wise min/max have landed.

HingeEmbeddingCriterion.lua
```diff
@@ -2,24 +2,38 @@ local HingeEmbeddingCriterion, parent = torch.class('nn.HingeEmbeddingCriterion'
...
+
+   self.buffer:resizeAs(input):copy(input)
+   self.buffer[torch.eq(y, -1)] = 0
+   self.output = self.buffer:sum()
+   
+   self.buffer = -input + self.margin
```
soumith: self.buffer:add(-1, input):add(self.margin)

HingeEmbeddingCriterion.lua
```diff
@@ -2,24 +2,38 @@ local HingeEmbeddingCriterion, parent = torch.class('nn.HingeEmbeddingCriterion'
...
-  if y == -1 and  dist > self.margin then
-     self.gradInput[1] = 0;
-  end
-  return self.gradInput 
+   if not torch.isTensor(y) then self.ty[1]=y; y=self.ty end
+   self.gradInput:resizeAs(input):copy(y)
+   self.gradInput[torch.cmul(torch.eq(y, -1), torch.gt(input, self.margin))] = 0
+   
+   if (self.sizeAverage == nil or self.sizeAverage == true) then 
+      self.gradInput = self.gradInput / input:nElement()
```
soumith: self.gradInput:mul(1 / input:nElement())

soumith:
made in-line comments.

HingeEmbeddingCriterion.lua
hughperkins: @mys007: your code is lua code in nn.  It can be used not only with cpu-based Tensors, but also with CUDA and OpenCL tensors.

mys007:
Thank for your feedback everybody, I updated the code accordingly. Note that I still use indexing by temporary tensors but there's IMHO no way around it except for introducing many buffers.
@hughperkins Yes, that's the motivation. Or was it a warning:)?

soumith:
could you squash your commits? ready for merge.

mys007:
done

PR #317, opened 2015-07-15
nn.NarrowTable

nicholas-leonard:
includes unit tests, doc.

### DIFF
--- a/NarrowTable.lua
+++ b/NarrowTable.lua
@@ -1,0 +1,41 @@
+local NarrowTable, parent = torch.class('nn.NarrowTable', 'nn.Module')
+
+function NarrowTable:__init(offset, length)
+   parent.__init(self)
+   self.offset = offset
+   self.length = length or 1
+   if not offset then
+      error('nn.NarrowTable(offset, length)')
+   end
+   
+   self.output = {}
+   self.gradInput = {}
+end
+
+function NarrowTable:updateOutput(input)
+   for k,v in ipairs(self.output) do self.output[k] = nil end
+   for i=1,self.length do
+      self.output[i] = input[self.offset+i-1]
+   end
+   return self.output
+end
+
+function NarrowTable:updateGradInput(input, gradOutput)
+   for i=1,#gradOutput do
+      self.gradInput[self.offset+i-1] = gradOutput[i]
+   end
+   for i=1,#input do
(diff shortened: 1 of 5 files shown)


NarrowTable.lua
```diff
@@ -0,0 +1,40 @@
...
+   if not offset then
+      error('nn.NarrowTable(offset, length)')
+   end
+   
+   self.output = {}
+   self.gradInput = {}
+end
+
+function NarrowTable:updateOutput(input)
+   for k,v in ipairs(self.output) do self.output[k] = v end
```
soumith: wait, what is going on here? is this a redundant line?

NarrowTable.lua
```diff
@@ -0,0 +1,40 @@
...
+   for i=1,#input do
+      if (i < self.offset) or (i >= self.offset + self.length) then
+         self.gradInput[i] = nn.utils.recursiveResizeAs(self.gradInput[i], input[i])
+         nn.utils.recursiveFill(self.gradInput[i], 0)
+      end
+   end
+   return self.gradInput
+end 
+
+function NarrowTable:type(type)
```
soumith: There's a PR we are thinking of merging soon, it requires a small change here: https://github.com/torch/nn/pull/303/files

Can you do the change here as well.

function NarrowTable:type(type, tensorCache)

NarrowTable.lua
```diff
@@ -0,0 +1,40 @@
...
+   self.output = {}
+   self.gradInput = {}
+   return parent.type(self, type)
```
soumith: return parent.type(self, type, tensorCache)

NarrowTable.lua
nicholas-leonard: No it just empties the table. That ways self.output always points to the same table.

soumith: hmm, I am very confused. You are just reassigning all the integer keys of self.output to their values. I dont see what has changed after this line, self.output is exactly the same.

nicholas-leonard: Hmm yes this is indeed an error. should be = nil.

Nicholas Léonard
917-741-7570

nicholas-leonard:
Ok. changes made. Added unit test to test variable length input.

nicholas-leonard:
This going in?

soumith:
Yes, for sure

soumith:
thanks!

PR #324, opened 2015-07-21
Remove unused code in nn.Module

albanD:
I think this code should have been removed in 50de1a9.
This is causing access to undeclared global variable (when running with `-g` flag) and could have some side effect if these globals are already used.


soumith:
thanks @albanD

PR #326, opened 2015-07-23
Implementation of nn.Min and nn.Max in Lua.

dominikgrewe:


dominikgrewe:
I've renamed `self.indices` to `self._indices`, because it's just an internal variable. If people actually access it, I'll change that back.

Max.lua
```diff
@@ -4,13 +4,38 @@ function Max:__init(dimension)
    parent.__init(self)
    dimension = dimension or 1
    self.dimension = dimension
-   self.indices = torch.Tensor()
+   self._indices = torch.LongTensor()
+   self._output = torch.Tensor()
```
soumith: Move these to be lazy initialized for backward compatibility

dominikgrewe:
> so gradOutput cannot be non-contiguous?

Good catch. All I want to do is add an empty dimension to `gradOutput`. Is there a way to do that without using `view`?

hughperkins:
What do you mean by 'add an empty dimension'?  Can you use something like unfold(1,1,1)?

dominikgrewe:
I probably should have said "add a singleton dimension". Unfortunately nn.Min and nn.Max are not consistent with torch.min/max, so for the backwards pass I have to expand the gradOutput, e.g. from MxN to Mx1xN if self.dimension is 2.

unfold(1, 1, 1) adds a singleton dimension as the last dimension, so combined with a permute it would have the desired effect. But if I do that I wonder if I might as well just create my own view, i.e. use `set` with my own size and stride.

dominikgrewe:
I've added the lazy initialization and also a `nn.utils.addSingletonDimension`, to replace the `view`. It's a bit hacky, but it seems clearer to me than using unfold + permute.

soumith:
thank you!

PR #330, opened 2015-07-25
[Parallel,Multi]Criterion table inputs

### LINKED ISSUE #272: Bug in ParallelCriterion
[ParallelCriterion](https://github.com/torch/nn/blob/master/ParallelCriterion.lua) seems hard-coded to work with criteria that accept tensors as input. So using it with [MarginRankingCriterion](https://github.com/torch/nn/blob/master/doc/criterion.md#nn.MarginRankingCriterion), which takes a table as input, gives the error 
> .../torch/install/share/lua/5.1/nn/ParallelCriterion.lua:41: attempt to call field 'new' (a nil value)

Here's a minimal test case: 
```lua
require 'nn'
input = {{torch.rand(1), torch.rand(1)}}
margin = nn.MarginRankingCriterion(1)
pc = nn.ParallelCriterion(true):add(margin)
pc:forward(input, 1)
pc:backward(input, 1)
```

Looking at the [code](https://github.com/torch/nn/blob/master/ParallelCriterion.lua#L41), I'm not sure what the correct behaviour is in this case; maybe a deep copy of the criterion's `gradInput` with all Tensors multiplied by the criterion's weight? 

At the very least we should indicate in the documentation that ParallelCriterion (and [MultiCriterion](https://github.com/torch/nn/blob/master/MultiCriterion.lua) as well) only work for Tensor-based criteria.

nicholas-leonard:
fixes #272

### DIFF
--- a/MultiCriterion.lua
+++ b/MultiCriterion.lua
@@ -23,10 +23,10 @@ function MultiCriterion:updateOutput(input, target)
 end
 
 function MultiCriterion:updateGradInput(input, target)
-   self.gradInput:resizeAs(input)
-   self.gradInput:zero()
+   self.gradInput = nn.utils.recursiveResizeAs(self.gradInput, input)
+   nn.utils.recursiveFill(self.gradInput, 0)
    for i=1,#self.criterions do
-      self.gradInput:add(self.weights[i], self.criterions[i]:updateGradInput(input, target))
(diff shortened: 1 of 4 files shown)


soumith:
Thanks, looks good

nicholas-leonard:
Funny how I need more table support from nn the more table support nn provides...

Issue #332, opened 2015-07-28
CAddTable has the same bug that JoinTable did


shawnjhenry:
Hi,

A while ago I noticed a bug in JoinTable here: https://github.com/torch/nn/issues/321.  It appears that CAddTable has exactly the same issue.  It might be a good idea to check that all of the table modules that should accept tables of varying size actually can.

Thanks,
Shawn

soumith:
hi! sorry about this. fixed 4 more modules in #333, looking into the rest now.

shawnjhenry:
No problem - thanks for fixing them so quickly!

shawnjhenry:
Hey, sorry to bother you again, but there are still more table modules with the bug.  ConcatTable is one of them.

Thanks,
Shawn

soumith:
thanks a lot for the report, fixing it.

soumith:
hey @shawnjhenry I'm not able to reproduce it for ConcatTable. Can you give me a short example script to reproduce it?

shawnjhenry:
Sure.

```
test = nn.ConcatTable()
test:add(nn.Identity())
test:add(nn.Identity())

x = {torch.randn(5), torch.randn(5)}
y = {torch.randn(5)}

print("output1")
print(test:forward(x))
print("gradInput1")
print(test:backward(x, {x, x}))
print("output2")
print(test:forward(y))
print("gradInput2")
print(test:backward(y, {y, y}))
```

As you can see, gradInput2 is a table of size two, when in fact it should be a table of size one.

Thanks,
Shawn

soumith:
@shawnjhenry concattable bug fixed via https://github.com/torch/nn/pull/378

PR #334, opened 2015-07-28
Fix SelectTable variable input size bug

soumith:
If you give a larger input, and then give a smaller input next, the gradInput will have artifacts from the previous backward.


soumith:
Similar to the ones reported and fixed in #332

PR #339, opened 2015-07-31
fixing corner-case of last dimension in addSingletonDimension

soumith:
Tensor of 3 dims can have a 4th singleton dimension added. The assertion was wrong but the rest of the code is correct. cc: @dominikgrewe


dominikgrewe:
Thanks! We just noticed the same :) I'll update the test so we also hit that case.

PR #341, opened 2015-08-01
Adds nn.Normalize

fmassa:
Generalizes https://github.com/torch/nn/pull/260 from @karpathy to accept arbitrary `L_p` norms.
A few remarks:
* Maybe `LpNormalize` or `Normalize` is a better name ?
* Only accepts 1D/2D inputs. @nicholas-leonard proposed to add a `dim` parameter to make it equivalent to `torch.norm`. Is it worth it given that we have `SpatialBatchNormalization` ?
* `updateGradInput` is quite memory consuming for large `d`.

cc: @bamos

### DIFF
--- a/Normalize.lua
+++ b/Normalize.lua
@@ -1,0 +1,87 @@
+local Normalize, parent = torch.class('nn.Normalize', 'nn.Module')
+
+function Normalize:__init(p,eps)
+  parent.__init(self)
+  assert(p,'p-norm not provided')
+  assert(p > 0, p..'-norm not supported')
+  self.p = p
+  self.eps = eps or 1e-10
+end
+
+function Normalize:updateOutput(input)
+  assert(input:dim() <= 2, 'only 1d layer supported')
+  local is_batch = true
+  if input:dim() == 1 then
+    input = input:view(1,-1)
+    is_batch = false
+  end
+
+  self.output:resizeAs(input)
+
+  self.norm = self.norm or input.new()
+  self.normp = self.normp or input.new()
+  self.buffer = self.buffer or input.new()
+
+  if self.p % 2 ~= 0 then
+    self.buffer:abs(input):pow(self.p)
+  else
+    self.buffer:pow(input,self.p)
+  end
+  self.normp:sum(self.buffer,2):add(self.eps)
+  self.norm:pow(self.normp,1/self.p)
+  self.output:cdiv(input,self.norm:view(-1,1):expandAs(self.output))
+
+  if not is_batch then
+    self.output = self.output[1]
+  end
+  return self.output
+end
+
+function Normalize:updateGradInput(input, gradOutput)
+  assert(input:dim() <= 2, 'only 1d layer supported')
+  assert(gradOutput:dim() <= 2, 'only 1d layer supported')
+
+  local is_batch = true
+  if input:dim() == 1 then
+    input = input:view(1,-1)
+    is_batch = false
+  end
+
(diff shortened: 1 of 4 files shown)


szagoruyko:
definitely the name has to be Normalize, not Norm
is it hard to extend to other dimensions? SpatialBatchNormalization is different

fmassa:
@szagoruyko Maybe it's not too difficult to extend it to work for other dimensions, by viewing the input to be 2D with the last dimension the desired normalized dimension, but one has to take care about batched/non-batched inputs as well. Maybe we could add a `setNumInputDims` function, as in `nn.View` ?
I'll spend some more time trying to make it more generic.

Everybody ok to renaming it to `Normalize` ?

karpathy:
Why is it obvious that the name should be Normalize and not Norm? As far as I can tell almost all operations from torch are ported without name changes to nn layers. E.g. analogous to `max` operation there is a `nn.Max`, so why is it obvious that `norm` operation should be `nn.Normalize`? My first instinct would be to stick with the current naming for naming consistency.

soumith:
torch.norm returns the p-norm. isn't this module also normalizing the input according to the returned norm?

karpathy:
Good point. Almost suggest that there should be both nn.Norm that does exactly what norm does, and then nn.Normalize that also does a div right afterwards. But perhaps that gets a bit too hairy then :)

soumith:
That's actually a good idea :)

soumith:
@fmassa looks like it's good to go is it?

fmassa:
It's good. I just need to squash the commits.
I didn't had the time to add a dimension parameter though, it would complicate a bit the logic because of the `setNumInputDims` function. But it could be added later if needed.

soumith:
squash, good to go.

fmassa:
@soumith just squashed the commits.

soumith:
thanks a lot Francisco, for the excellent PR.

ffmpbgrnn:
Lots of memory is needed in backprop of this module. One reason might be creating the `eyeExpand` matrix and later doing multiplication. In my case, when the batch size is 64, input dimension is 4800, two `Normalize` layer would be out of memory for one 4G memory GPU. Any idea to implement a more space efficient Normalize layer?

fmassa:
@ffmpbgrnn here is a version of `Normalize` which uses much less memory (it doesn't depend on the batch size anymore). It should be slower on GPU. The tests passes so it should be fine. Use it with `fastMode(false)`.
https://github.com/fmassa/nn/commit/015ba9c9252bea61c3f3e6cccc6f0d81cede8bdc

ffmpbgrnn:
@fmassa thanks for your work. I will give a try!

fmassa:
@ffmpbgrnn so, does this patch works fine for you ? Is it much slower than the previous version ?
Maybe we could push this simplified version to master (taking of the `faster` mode to make things simple) ?
cc @soumith

ffmpbgrnn:
@fmassa , sorry for my late response..Yes, it passes the test. But when I test `fastMode(false)` mode on GPU, it's even faster than the `fastMode(true)`.
```lua
require 'nn'
require 'cutorch'
require 'cunn'

local module = nn.Normalize(2):cuda()
module:fastMode(false)
local input = torch.rand(64, 2400):cuda()
local t = torch.Timer()
for i = 1, 100 do
	module:forward(input)
	module:backward(input, input)
	print(i)
end
print(t:time().real/100)
```

> fastMode(false), output: 0.09
> fastMode(true), output: 0.12

PR #344, opened 2015-08-04
fixed nn.Padding

mys007:
nn.Padding was broken: input was returned in backprop instead of gradInput. Pitty that the test didn't catch it.


soumith:
ugh, sorry about this.

nicholas-leonard:
My bad. Thanks.

PR #345, opened 2015-08-04
Fix contiguous gradoutput bug in nn.Mean

kosklain:
Currently, `nn.Mean` is assuming contiguous gradOutput in its backward pass. Therefore, if, for example, two `nn.Mean` modules are chained, the backward pass will not produce correct gradients.

This minimal example can be used to reproduce the problem. This pr fixes it.

```lua
-- This script demonstrates that the non-contiguous output of nn.Mean():backward() leads
-- to backpropagation errors when it is fed into another nn.Mean():backward().
require 'nn'

-- Initialise data and backpropped gradient.
local data = torch.randn(3, 2, 7, 7)
local outGrad = torch.randn(3, 2)

-- Create a stack of two nn.Mean() layers.
local mod1a = nn.Mean(4)
local mod1b = nn.Mean(3)

-- Run fprop & bprop through the stack.
local resFwd1a = mod1a:forward(data)
local resFwd1b = mod1b:forward(resFwd1a)
local resBwd1b = mod1b:backward(resFwd1a, outGrad)
local resBwd1a = mod1a:backward(data, resBwd1b)

-- Returns false.
print(resBwd1b:isContiguous())

-- Should return 0, but it does not.
print((resBwd1a - torch.repeatTensor(outGrad:view(3, 2, 1, 1), 1, 1, 7, 7) / 49):norm())

-- Now, we make the output contiguous and it works.
resBwd1a = mod1a:backward(data, resBwd1b:contiguous())
print((resBwd1a - torch.repeatTensor(outGrad:view(3, 2, 1, 1), 1, 1, 7, 7) / 49):norm())

### DIFF
--- a/Mean.lua
+++ b/Mean.lua
@@ -4,6 +4,7 @@ function Mean:__init(dimension)
    parent.__init(self)
    dimension = dimension or 1
    self.dimension = dimension
+   self._gradInput = torch.Tensor()
 end
 
 function Mean:updateOutput(input)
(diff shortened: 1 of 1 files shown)


dominikgrewe:
This looks good to me. @soumith What do you think?

soumith:
Thanks kosklain

PR #347, opened 2015-08-11
fix indentation

hughperkins:


soumith:
it's a bit out of date, and i'd rather keep the git history that's convenient for git blame, rather than mass-changing the indentations. It's a bit of clowntown I know, but.......

PR #348, opened 2015-08-11
nn.readthedocs.org

nicholas-leonard:
These changes will give you this : http://nn.readthedocs.org/en/readthedocs/index.html . Users will need to update dok as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 [![Build Status](https://travis-ci.org/torch/nn.svg?branch=master)](https://travis-ci.org/torch/nn)
-<a name="nn.dok"/>
+<a name="nn.dok"></a>
 # Neural Network Package #
 
 This package provides an easy and modular way to build and train simple or complex neural networks using [Torch](https://github.com/torch/torch7/blob/master/README.md):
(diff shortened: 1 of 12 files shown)


nicholas-leonard:
@clementfarabet @andresy @soumith @koraykv If you guys like the idea of setting up a readthedocs documentation site, then I will proceed to do the same with torch, image, and so on. It shouldn't change any of the look and feel of the github docs. However, if we do decide to move to readthedocs, all the inter-repo hyperlinks will need to point to readthedocs instead of github. So basically, waiting for your approval to go forward with this migration of docs to readthedocs.org.

andresy:
Just curious: there is no way do to this with relative paths, somehow? It would be nice if one could also read the doc offline. In any case, I think it is nicer like this, thanks!

nicholas-leonard:
@andresy You can use relative paths within a repository. But I know of no way to keep the links between different repositories relative (so that the github ones point to github, and the readthedocs ones point to readthedocs).

hughperkins:
That looks quite nice :-)

soumith:
this looks quite cool nick!

PR #351, opened 2015-08-15
Forcing zeroing when self.output is resized (nn.Linear)

clementfarabet:
This fixes https://github.com/torch/nn/commit/8abe926769258b63ec0938f35d85a58960f01425


soumith:
Thanks clem!

PR #353, opened 2015-08-21
Improve ClassNLLCriterion

adamlerer:
Moved ClassNLLCriterion to C. Normalization with weights is now done by the total weight of the minibatch, rather than its size

### DIFF
--- a/ClassNLLCriterion.lua
+++ b/ClassNLLCriterion.lua
@@ -1,14 +1,26 @@
-local ClassNLLCriterion, parent = torch.class('nn.ClassNLLCriterion', 'nn.Criterion')
+local ClassNLLCriterion, parent = torch.class(
+    'nn.ClassNLLCriterion',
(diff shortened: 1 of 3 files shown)


soumith:
cc: @jonathantompson note that the normalization is changed to what was a more sensible way when using weights.

jonathantompson:
Ahh. OK.  Would it be a better idea to have a flag for this new normalization scheme (rather than changing default behavior transparently)?

soumith:
idk, it's up to you. the option was only recently added anyways right.

soumith:
@jono do you really want me to introduce a flag for this normalization scheme? afaik only you are using this right, because it was recently introduced?

jonathantompson:
Nah, it's OK. I'm sure no one else will care.

soumith:
@jonathantompson thanks!

PR #355, opened 2015-08-25
Fix SpatialBatchNormalization documentation example.

sbodenstein:
The SpatialBatchNormalizationExample does not evaluate, as the colon operator is not used (model.forward is rather than model:forward).


soumith:
thank you!

PR #359, opened 2015-08-30
CosineEmbeddingCriterion supports batch mode

fmassa:
Extends `CosineEmbedidingCriterion` to support batch mode. A few remarks
* I couldn't avoid memory allocation in https://github.com/torch/nn/compare/master...fmassa:cosine_embed_batch?expand=1#diff-d0f98c4047e236a26fe4f983397e0e15R63 and https://github.com/torch/nn/compare/master...fmassa:cosine_embed_batch?expand=1#diff-d0f98c4047e236a26fe4f983397e0e15R65
* added a slightly modified version of `criterionJacobianTest1DTable` for `table` inputs


soumith:
thanks a lot! This looks quite good.

PR #360, opened 2015-09-01
Optimized Spatial Batch Normalization

egonina:
This patch adds calls to the optimized GPU routines for spatial batch normalization. The optimized routines give up to 2x performance improvement over using the original batch normalization GPU implementation for training models. This change is in conjunction with PR: https://github.com/torch/cunn/pull/133. Defaults to the original implementation for problem dimensions that it does not support.

cc @clementfarabet


soumith:
fixed via #550

PR #362, opened 2015-09-02
Using the speed-up version of updateParameters for SparseLinear

zhangxiangxiao:
Is there any reason not to do this in the current code? The special SparseLinear_updateParameters function is implemented in C but SparseLinear.updateParameters still points to Module.updateParameters. Let me know if there is some purpose I did not see.


soumith:
dont see a reason why not. let's see what is uncovered.

PR #363, opened 2015-09-03
getParameters: improve memory use, fix bug with non-compact tensors

adamlerer:
This was meant to reduce temp buffer use when calling getParameters on a cunn model.

The old code uses 2x temporary buffer space. This code reduces it to 1x, and adds a hook so you can construct that buffer on the CPU rather than the GPU.

While rewriting this function, I also found a couple pretty severe bugs. Basically, if you have a non-compact parameter tensor (i.e. non-contiguous even after transpose), then the tensor will now point into unallocated memory. You can't completely fix that bug, because in some cases there's no way to "compact" the parameters while still having constant strides. So I have to just assert against that case.

### DIFF
--- a/Module.lua
+++ b/Module.lua
@@ -113,12 +113,15 @@ function Module:clone(...)
    return clone
 end
 
-function Module:type(type)
+function Module:type(type, tensorCache)
    assert(type, 'Module: must provide a type to convert to')
+
    -- find all tensors and convert them
    for key,param in pairs(self) do
       self[key] = nn.utils.recursiveType(param, type)
+
    end
+
    return self
 end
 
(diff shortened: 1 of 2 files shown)


soumith:
thanks! after month-long testing internally, this is stable and fixes critical bugs, so I am merging sooner than later.

noa:
I'm seeing the following error:

```
luajit: /usr/local/share/lua/5.1/nn/Module.lua:258: expected arguments: [*CudaTensor*] CudaTensor CudaTensor
stack traceback:
        [C]: in function 'maskedSelect'
        /usr/local/share/lua/5.1/nn/Module.lua:258: in function 'flatten'
        /usr/local/share/lua/5.1/nn/hessian.lua:332: in function 'getParameters'
[...]
```

In `hessian.lua`, `flatten` works for `parameters` and `gradParameters` (`nparameters` == `nUsedParameters`), and the error above is for `hessianParameters` (where `nparameters` != `nUsedParameters`).

soumith:
cannot reproduce. this is the test script that I am using:

```lua
require 'nn'
require 'cunn'
MyModule = nn.Sequential():add(nn.Linear(300,300))
MyModule:getParameters():uniform()
MyModule:cuda()
paramx, paramdx = MyModule:getParameters()


MyModule = nn.Sequential():add(nn.Linear(300,300))
MyModule:cuda()
paramx, paramdx = MyModule:getParameters()
paramx:uniform()
```

PR #368, opened 2015-09-07
Fix non-contiguous test in SpatialConvolutionMM

fmassa:
Both `output` and `outputc` pointed to the same tensor (`module.output`), so the tests would never fail. Same for `gradInput`.


dominikgrewe:
Good catch, thanks!

PR #370, opened 2015-09-09
Update SoftMax to work in spatial mode

jhjin:
The module applies softmax in space and works both on a single
vector/matrix (1D/2D) or spatial vector/matrix (3D/4D).

### DIFF
--- a/SpatialSoftMax.lua
+++ b/SpatialSoftMax.lua
@@ -1,0 +1,9 @@
+local SpatialSoftMax, _ = torch.class('nn.SpatialSoftMax', 'nn.Module')
+
+function SpatialSoftMax:updateOutput(input)
+   return input.nn.SoftMax_updateOutput(self, input)
+end
+
+function SpatialSoftMax:updateGradInput(input, gradOutput)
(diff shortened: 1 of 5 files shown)


jhjin:
One concern is that the backend is similar to `nn.SoftMax` and mostly indexing is different. It might be redundant to keep this code along with `nn.SoftMax`.

soumith:
@jhjin we could have a C implementation that is shared between SoftMax and SpatialSoftMax, and for the future, VolumetricSoftMax. If you send a PR at a later date to refactor that, I am happy to accept.

jhjin:
@soumith thanks for the comment. will do that.

soumith:
@jhjin found a bug. if the gradOutput is non-contiguous, you will get wrong results. Similar to how you added newContiguous for input, I suppose you could do the same with gradOutput

jhjin:
@soumith I've updated the branch to reflect the non-contiguous gradOutput/output issue. Similar changes were made in `cunn` as well https://github.com/torch/cunn/pull/135

soumith:
thanks a lot Jong!

culurciello:
well done

PR #372, opened 2015-09-11
Minor change for memory saving

korkinof:
Super-minor change. Not initialising those variables saved 30% of GPU memory when not training.


soumith:
thanks

PR #373, opened 2015-09-11
BCECriterion rewrite

soumith:
Author: Armand Joulin

Rewriting the BCECriterion. The goal is to make it clearer and to have the possibility to weight the different terms.

New features:
- can use weights
- use less memory during forward (3 times less) and backward (2 times less).
- slightly faster forward and backward on both gpu and cpu


soumith:
thanks Armand.

PR #374, opened 2015-09-12
Add SmoothL1Criterion

andreaskoepf:
Working on some Faster-RCNN code I wrote a simple (currently LUA-only) version of a SmoothL1Criterion. It handles outliers better than MSE but offers the same convergence behavior of a squared error term when the values become smaller (effectively it is AbsCriterion - 0.5 for large and MSE * 0.5 for small errors).

Since there are a couple of interesting loss-functions out there I was not sure whether you guys are considering the criterion-collection sealed or if additions are welcome. I am just going forward with this PR and try my luck.

Would it be worth to create .c/.cu impls? If yes, I could provide them as well.


szagoruyko:
oh cool I have c and cuda versions but was lazy to write the doc , we can merge
will do a PR

andreaskoepf:
@szagoruyko Fantastic, let's do this.

andreaskoepf:
#376 is the follow-up PR by @szagoruyko

PR #375, opened 2015-09-13
Set default value of total weight in ClassNLL to 1

jhjin:
`total_weight_tensor` becomes to have a meaningful number after loss:forward(). Calling loss:backward() without loss:forward() will end up with zero output https://github.com/torch/nn/commit/ef2e3d290418c411add8eee8dd5ca7d486362ad8#diff-e3ab2ae1e4ca35f9a37e0702421d52ffR97. This commit considers the case.


soumith:
thanks Jong!

PR #376, opened 2015-09-13
Add C SmoothL1 criterion

szagoruyko:
with a doc by @andreaskoepf from https://github.com/torch/nn/pull/374
CUDA PR is coming

### DIFF
--- a/generic/SmoothL1Criterion.c
+++ b/generic/SmoothL1Criterion.c
@@ -1,0 +1,60 @@
+#ifndef TH_GENERIC_FILE
+#define TH_GENERIC_FILE "generic/SmoothL1Criterion.c"
+#else
+
+static int nn_(SmoothL1Criterion_updateOutput)(lua_State *L)
+{
+  THTensor *input = luaT_checkudata(L, 2, torch_Tensor);
+  THTensor *target = luaT_checkudata(L, 3, torch_Tensor);
+  int sizeAverage = luaT_getfieldcheckboolean(L, 1, "sizeAverage");
+  real sum;
+
+  sum = 0;
+  TH_TENSOR_APPLY2(real, input, real, target,
+                   real z = fabs(*input_data - *target_data);
+                   sum += z < 1 ? 0.5*z*z : z - 0.5;)
+
+  if(sizeAverage)
+    sum /= THTensor_(nElement)(input);
+
+  lua_pushnumber(L, sum);
+  lua_setfield(L, 1, "output");
+
+  lua_pushnumber(L, sum);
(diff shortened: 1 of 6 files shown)


generic/SmoothL1Criterion.c
```diff
@@ -0,0 +1,60 @@
...
+static int nn_(SmoothL1Criterion_updateOutput)(lua_State *L)
+{
+  THTensor *input = luaT_checkudata(L, 2, torch_Tensor);
+  THTensor *target = luaT_checkudata(L, 3, torch_Tensor);
+  int sizeAverage = luaT_getfieldcheckboolean(L, 1, "sizeAverage");
+  real sum;
+
+  sum = 0;
+  TH_TENSOR_APPLY2(real, input, real, target,
+                   real z = fabs(*input_data - *target_data);
```
soumith: fabs will work correctly for double tensors. For float, you have to use fabsf.
These are the C99 prototypes for the function:
``C
     double fabs  (double x);
      float fabsf (float x);
``

szagoruyko: what would be the difference with https://github.com/torch/nn/blob/master/generic/AbsCriterion.c#L14 ?

soumith: ugh, i suppose that is equally incorrect.

andreaskoepf:
Isn't torch itself using fabs for FloatTensors as well? see https://github.com/torch/torch7/blob/086ec2a760a15a5165b488e25899f3ee70e2ec6e/TensorMath.lua#L946

generic/SmoothL1Criterion.c
andreaskoepf: Isn't torch itself using fabs for FloatTensors as well? see https://github.com/torch/torch7/blob/086ec2a760a15a5165b488e25899f3ee70e2ec6e/TensorMath.lua#L946

fmassa: yes, @andreaskoepf is right, `torch` has different `abs` for `int` and `long`, but for `double` and `float` it uses `fabs`
For a reference, look at [fabs](https://github.com/torch/torch7/blob/master/lib/TH/generic/THTensorMath.c#L1844), [labs](https://github.com/torch/torch7/blob/master/lib/TH/generic/THTensorMath.c#L1801) and [abs](https://github.com/torch/torch7/blob/master/lib/TH/generic/THTensorMath.c#L1805)

soumith: sergey said he'll correct this disrepancy in a separate PR overall for torch.

PR #377, opened 2015-09-13
in-place dropout

szagoruyko:
reviewed internally at facebook

### DIFF
--- a/Dropout.lua
+++ b/Dropout.lua
@@ -1,9 +1,10 @@
 local Dropout, Parent = torch.class('nn.Dropout', 'nn.Module')
 
-function Dropout:__init(p,v1)
+function Dropout:__init(p,v1,inplace)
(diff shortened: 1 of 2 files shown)


hughperkins:
you have inconcistences with the spacing after your commas `,`.  In the original code, it is aligned with Python pep8 guidelines of a space after each comma, eg:

```
function Dropout:updateGradInput(input, gradOutput)
```
... but in your code, there is no space:
```
function Dropout:__init(p,v1,inplace)
```

soumith:
fwiw, the original code is equally inconsistent. I think during a transition period of 6 months to 1 year, we should keep pointing out spacing and other nit-picked inconsistencies, but should merge them, so that the friction of sending in a PR is not perceived to have increased (which would turn away or delay some developers)

PR #378, opened 2015-09-14
fixing bug in ConcatTable variable length

soumith:
fixes test case in https://github.com/torch/nn/issues/332#issuecomment-126774117

added test case to unit tests


soumith:
cc: @nicholas-leonard

PR #381, opened 2015-09-15
minor math typo

eulerreich:


soumith:
there's an incoming PR on BCECriterion, merging that first.

PR #387, opened 2015-09-16
Fast weight decay for SparseLinear

zhangxiangxiao:
As title suggests, this pull request offers a fast weight decay for SparseLinear.

One thing to note is that the new weight decay mechanism has different meaning than before. The new mechanism applies decay only if the input vector contains a value of the corresponding index, whereas previously weight decay is applied no matter what the input is. This may or may not be what you want, although I do think that it is not useful to apply weight decay for weights that are not indexed.

Note also that by default the .weightDecay member is 0, and the constructor does not set it. I read this as weightDecay is actually a hack in case one really needs it.


soumith:
looks a bit hacky. But again, sparse stuff is always hard. Do you reaaaaly want this in?

zhangxiangxiao:
I agree that this is super hacky and non-canonical. But considering the slow version of weight decay is already there, this pull request is no more hacky that it already is.

So yes, I would love to see this in.

soumith:
done!

PR #390, opened 2015-09-17
Revert "fixing cuda getparams"

soumith:
Reverts torch/nn#389


soumith:
sorry, this breaks things for CPU. More proper fix coming in.

PR #392, opened 2015-09-20
Option to add Padding to Volumetric Convolution

kmul00:
Padding option wasn't previously applicable to the Volumetric Convolution, though the CUDA kernel has it. This is the CUDA version of it.

Padding can be provided in the same was as for `SpatialConvolution` :

    module = nn.SpatialConvolution(nInputPlane, nOutputPlane, kT, kW, kH, [dT], [dW], [dH], [padT], [padW], [padH])

Default is `0`

Note : I have sent a PR for the CUDA version of it, but I am unable to do it for the CPU version. It would be wonderful if someone can do the same for the CPU version `VolumetricConvolution.c` in the `nn` package.


soumith:
I'll wait on the CPU version to merge this in.

Issue #396, opened 2015-09-24
make nn.Linear handle 3D tensor


blackyang:
Hi there,

Could we extend nn.Linear so that it can handle 3D tensor?

When dealing with RNN, sometimes I have an output tensor with shape:
> (nb_batch * nb_seq_len * nb_dim1). 

I want to feed it into a fully connected layer in order to get a tensor of shape:
>(nb_batch * nb_seq_len * nb_dim2). 

However since both **nb_batch** and **nb_seq_len** may vary (namely both of them are unknown in general case. Although in practice I am able to know nb_batch myself), I couldn't reshape it into 2D tensor, use nn.Linear and then reshape back to 3D tensor. 

Thus I think it would be convenient if nn.Linear is able to handle 3D tensor. Such need is common when dealing with RNN. I guess it shouldn't be too hard to do this extension, I can give a try when I have time.

szagoruyko:
isn't https://github.com/torch/nn/blob/master/doc/simple.md#nn.MM what you want?

blackyang:
Hi @szagoruyko , sorry I am not sure whethe I get your point, but I want a **trainable** fully connected layer, not a matrix multiplication

szagoruyko:
oh yeah sure, I think I have seen it somewhere. You can take a look at nn.MM and nn.Linear implementations, has to be pretty easy to combine them.

blackyang:
Yes it should be easy to implement, I can give a try when I have time. Thanks for your pointer to nn.MM :)

hughperkins:
I'd like to see this too :-)

blackyang:
Hi all, I made a pull request  [here](https://github.com/torch/nn/pull/398)

ankitvgupta:
I know this thread been closed a while, but I had a similar need and solved it with nn.Bottle (https://github.com/torch/nn/blob/master/doc/containers.md#nn.Bottle) - just wanted to put that here since I suspect others will be searching for this too.

PR #398, opened 2015-09-24
make nn.Linear handle 3D tensor

blackyang:
Make nn.Linear handle 3D tensor, see #396 issue for details

Basically, when handling 3D tensor of shape (nb_batch, nb_length, nb_dim), I first resize it to (nb_batch *nb_length, nb_dim), do computation similar to that of handling 2D tensor, and then resize back to 3D tensor. I also add some test case in test.lua.

### DIFF
--- a/Linear.lua
+++ b/Linear.lua
@@ -62,10 +62,10 @@ function Linear:updateOutput(input)
       if not self.addBuffer or self.addBuffer:nElement() ~= (nframe * nLength) then
          self.addBuffer = input.new(nframe * nLength):fill(1)
       end
-      self.output:addmm(0, self.output, 1, input:view(nframe * nLength, nDimIn), self.weight:t())
+      self.output:addmm(0, self.output, 1, input:resize(nframe * nLength, nDimIn), self.weight:t())
       self.output:addr(1, self.addBuffer, self.bias)
-      self.output = self.output:view(nframe, nLength, nDimOut)
-      input = input:view(nframe, nLength, nDimIn)
+      self.output:resize(nframe, nLength, nDimOut)
+      input:resize(nframe, nLength, nDimIn)
    else
       error('input must be vector or matrix or 3D tensor')
    end
@@ -86,14 +86,14 @@ function Linear:updateGradInput(input, gradOutput)
       elseif input:dim() == 2 then
          self.gradInput:addmm(0, 1, gradOutput, self.weight)
       elseif input:dim() == 3 then
-         local nframe = input:size(1)
-         local nLength = input:size(2)
+         local nframe = gradOutput:size(1)
+         local nLength = gradOutput:size(2)
          local nDimOut = self.weight:size(1)
          local nDimIn = self.weight:size(2)
-         self.gradInput = self.gradInput:view(nframe * nLength, nDimIn)
-         self.gradInput:addmm(0, 1, gradOutput:view(nframe * nLength, nDimOut), self.weight)
-         self.gradInput = self.gradInput:view(nframe, nLength, nDimIn)
-         gradOutput = gradOutput:view(nframe, nLength, nDimOut)
+         self.gradInput:resize(nframe * nLength, nDimIn)
+         self.gradInput:addmm(0, 1, gradOutput:resize(nframe * nLength, nDimOut), self.weight)
+         self.gradInput:resize(nframe, nLength, nDimIn)
+         gradOutput:resize(nframe, nLength, nDimOut)
       end
 
       return self.gradInput
@@ -109,16 +109,16 @@ function Linear:accGradParameters(input, gradOutput, scale)
       self.gradWeight:addmm(scale, gradOutput:t(), input)
       self.gradBias:addmv(scale, gradOutput:t(), self.addBuffer)
    elseif input:dim() == 3 then
-      local nframe = input:size(1)
-      local nLength = input:size(2)
+      local nframe = gradOutput:size(1)
+      local nLength = gradOutput:size(2)
       local nDimOut = self.weight:size(1)
       local nDimIn = self.weight:size(2)
-      gradOutput = gradOutput:view(nframe * nLength, nDimOut)
-      input = input:view(nframe * nLength, nDimIn)
+      gradOutput:resize(nframe * nLength, nDimOut)
+      input:resize(nframe * nLength, nDimIn)
       self.gradWeight:addmm(scale, gradOutput:t(), input)
       self.gradBias:addmv(scale, gradOutput:t(), self.addBuffer)
-      gradOutput = gradOutput:view(nframe, nLength, nDimOut)
-      input = input:view(nframe, nLength, nDimIn)
+      gradOutput:resize(nframe, nLength, nDimOut)
+      input:resize(nframe, nLength, nDimIn)
    end
 end
 
--- a/Linear.lua
+++ b/Linear.lua
@@ -49,8 +49,25 @@ function Linear:updateOutput(input)
       end
       self.output:addmm(0, self.output, 1, input, self.weight:t())
       self.output:addr(1, self.addBuffer, self.bias)
+   elseif input:dim() == 3 then
+      local nframe = input:size(1)
+      local nLength = input:size(2)
+      local nDimOut = self.weight:size(1)
+      local nDimIn = self.weight:size(2)
+      local nElement = self.output:nElement()
+      self.output:resize(nframe * nLength, nDimOut)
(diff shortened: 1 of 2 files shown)


szagoruyko:
you don't actually need to modify nn.Linear, can just do this:

```lua
local Linear3D,parent = torch.class('nn.Linear3D, nn.Linear')

function Linear3D:__init(...)
  parent.__init(self,...)
end

function Linear3D:updateOutput(input)
  self.bs, self.nseq, self.dim1 = table.unpack(input:size():totable())

  self.output:resize(self.bs * self.nseq,self.dim2)
  parent.updateOutput(self,input:view(-1,dim1))
  self.output = self.output:view(self.bs, self.nseq, -1)
  return self.output
end

function Linear3D:updateGradInput(input, gradOutput)
  self.gradInput:resize(self.bs * self.nseq, self.dim1)
  parent.updateGradInput(self, input:view(-1,dim1), gradOutput:view(-1,dim2))
  self.gradInput = self.gradInput:view(self.bs, self.nseq, -1)
  return self.gradInput
end

-- accGradParameters
```

(quickly prototyped, don't expect to work immediately)

blackyang:
yes that is also one possible way :)

blackyang:
@szagoruyko  Do you know whether the difference between view() and resize() is large or not in term of running time? Thanks.

szagoruyko:
@blackyang view does not allocate memory, resize allocates in most cases. view is supposed to be used wherever possible

blackyang:
@szagoruyko  thanks for clarification, I would change resize() in my code into view().

hughperkins:
I think you're better off putting resize:
- if it needs to resize, putting view will fail
- if it doesnt need to resize, they're both effectively instantaneous

blackyang:
Hi @hughperkins , I was thinking view() as a strict mode of resize() in my last commit. Besides, I do leave some resize() in certain places such as [L58](https://github.com/torch/nn/pull/398/files#L58) and [L80](https://github.com/torch/nn/pull/398/files#L80)

But probably you're right, putting resize is somehow more general. I am not sure which way is better.

hughperkins:
As I remember, other modules all use resize, eg https://github.com/torch/nn/blob/master/CMul.lua#L11 https://github.com/torch/nn/blob/master/generic/SoftMax.c#L41

blackyang:
Oh I can change back to resize, thanks for your time. BTW which way is the better to roll back, git revert HEAD then commit?

blackyang:
seems resize() version is slower than view() version, since the latter one passed the build while the former failed because "no results received in last 10 minutes". Of course that could also due to randomization in test.lua

hughperkins:
re rollback: normally, you'll probably need to squash your commits before merge anyway, eg see `git rebase -i` (not very reliable, a bit conflict prone), or just in `gitk`, right click `master` branch commit, do 'reset such-and-such branch to here', 'mixed', then redo `git add --all .`, to create a new commit.

re: random: personally, I add `torch.manualSeed(123)` at the start of each test, so it's deterministic.  This is not standard practice though (though I'm kind of pushing this :-P)

re: test time: I think you need to give precise details on which test is failing, dimensions of input tensors etc.  If you use deterministic tests, ie `manualSeed`, then it should be easy to obtain these details, using `print` etc.

hughperkins:
Oh wait, resize on input .... sorry, I may have not read carefull the first time.  input sould not need resize.  view is correct for input.  Sorry :-P

blackyang:
Aha seems I need to revert a "revert" :) thanks for your clarification

hughperkins:
Yeah, sorry about that :-P

blackyang:
No worries :) I shouldn't resize input tensor

soumith:
hey @blackyang . nn.Linear is one of the most used modules in nn, and there are various places at which people only expect it to be accepting 1D or 2D inputs. Adding support for 3D inputs will likely insert silent bugs (when someone forgets an nn.View or nn.Reshape, as 3D also applies to output of non-minibatch SpatialConvolution).

Taking these points into consideration, I think you should separate out the 3D-supported version initially into a separate module, and put it under nnx: https://github.com/clementfarabet/lua---nnx

Over due time, if the 3D version has enough adoption, we will include it into this repo.

soumith:
The PR looks very clean and well-thought, I hope to see more contributions from you in the future :)

blackyang:
Hi @soumith no problem, I can make a pull request to nnx, thanks for your reply :)

BTW,
> separate out the 3D-supported version initially into a separate module

Do you mean something similar to szagoruyko's implementation above?

hughperkins:
I reckon it would be good to change the name, eg 'SpatialLinear', or 'FullyConnected'.  Changing the name to FullyConnected would have two advantages:
- no longer changes the way a commonly-used module works
- name matches what is written in papers :-)

blackyang:
Hi @soumith , there is a SpatialLinear module in nnx that roughly does the same thing. The SpatialLinear  module does the following conversion:
> (nb_dim1, width, height) --> (nb_dim2, width, height)

while mine does:
>(nb_batch, nb_length, nb_dim1) -->  (nb_batch, nb_length, nb_dim2)

Although the scenarios are different, the math beneath is the same. So I guess I shouldn't make PR to nnx. Besides, I think szagoruyko's implementation above might be a good idea, a new module inherits from nn.Linear

hughperkins:
@soumith I think we should add some more publicity to nnx perhaps?  I've been looking for something like SpatialLinear/FullyConnected since ~June, and didnt know about its existence until blackyang created this PR.  I even created my own fully connected once, https://github.com/hughperkins/clnn/blob/f9e69dcd5faa173460a35c6ac49efe8550be0521/FullyConnected.lua ,  not nearly as clean as blackyang's admittedly :-)  I ditched it, because i reckoned it would need a lot of socialization to get it accepted, and I have larger battles to fight :-)

blackyang:
nnx itself is less documented, perhaps the top TODO should be calling more documentation for nnx...

BTW, for those (e.g. me) who love szagoruyko's implementation above, take care of this line:
```lua
local Linear3D,parent = torch.class('nn.Linear3D, nn.Linear')
```

which should be:
```lua
local Linear3D,parent = torch.class('nn.Linear3D', 'nn.Linear')
```

I spent a lot of time wondering why there was no nn.Linear3D after complie.... hahaha :-P

szagoruyko:
lol this is a funny bug indeed, sorry

hughperkins:
Hmmm, that bug is hard to see :-)   Hmm, I hadnt noticed that Sergey had created a new name before.  Linear3D sounds good to me :-)

hughperkins:
Hmmm, nnx is part of distro, just noticed. Interesting.  So, should maybe be made more obvious in the core torch docs?

hughperkins:
Hmmm, nnx.SpatialLinear is not quite the same right?  Maybe I will just start using what Sergey wrote above, at least it's had a few pairs of eyes on it.

hughperkins:
Oooohhhhhh....  I remember why I abandoned this last time around.  https://groups.google.com/forum/#!searchin/torch7/hugh$20perkins$20fullyconnected/torch7/qhPw9m6X3pE/fIXtIhwxnSUJ  What I wrote is:

"Hmmm, you know what, I think I'm looking at this the wrong way.  Bascially everything from the first fc layer onwards has vectors for data samples, on up through the logsoftmax etc.

However, having said that, I do think there could be some simplification possibly, since stating the input to a layer explicitly is redundant I think?

eg, the following information is sufficient to specify the dimensions of a network: 32c5-mp3-32c5-mp2-150n-10n

The 150n takes whatever has arrived at that point, and connects it to 150 output neurons.  Likewise for 10n.

So, I'd like to be able to specify the network somethin glike:

```
model:add(nn.Conv(32,5,5))
model:add(nn.Relu())
model:add(nn.MaxPooling(3,3,3,3)
model:add(nn.Conv(32,5,5))
model:add(nn.Relu())
model:add(nn.MaxPooling(2,2,2,2)
model:add(nn.Linear(150))   <= the input dimensions are defined by the input, so dont need to calculate it ourselves?
model:add(nn.Tanh())
model:add(nn.Linear(10))  <= here too?
model:add(nn.LogSoftMax())
model:add(nn.NLLCriterion())
```
"

tastyminerals:
So, what's the status of `Linear3D`? It appears I need linear transformation for `4x5` --> `4x5x200`.

tastyminerals:
In case somebody wanders here and tries to ask a question where is `Linear3D`. Do use `nn.Bottle` for this.

Issue #397, opened 2015-09-24
about R-operator


cswhjiang:
R-operator [1] is useful in some gradient based methods, e.g. [2]. And it is implemented in Theano.  I think nn should provide similar function.
[1] Pearlmutter, Barak A. "Fast exact multiplication by the Hessian." Neural computation 6.1 (1994): 147-160.
[2] Dauphin, Yann N., et al. "RMSProp and equilibrated adaptive learning rates for non-convex optimization." arXiv preprint arXiv:1502.04390 (2015).

soumith:
@bartvm has some code.

cswhjiang:
TO @soumith : this (https://github.com/bartvm/nn) repo ?

bartvm:
@cswhjiang I made a PR with what I have, but you'll probably need to implement a few more modules before you can use it. If you end up doing that, feel free to make a PR to my fork!

cswhjiang:
Thanks. I will take a look first.

PR #399, opened 2015-09-25
WIP: R-op

bartvm:
This is some preliminary code for the R-op in Torch. Currently only the Linear, Tanh and Sigmoid modules are implemented, but it should be relatively straightforward to implement new modules. Making a PR to get some feedback, since there was some interest (#397), but it needs some work before it's actually in a useful state.

I compared the results to Theano for correctness.

### DIFF
--- a/doc/rop.md
+++ b/doc/rop.md
@@ -1,0 +1,54 @@
+# The R-op
+The R-op is an efficient method of calculating the matrix-vector product of the
+Hessian of a model and any vector. It is often used to quickly
+approximate the curvature in a given direction, to speed up
+optimization, or to calculate the full Hessian by repeatedly multiply
+with one-hot vectors.
+
+## Example
+Given the following function,
+```
+h = σ(Wx + b)
+y = σ(Wh + b)
+E = sum(y)
+```
+this code calculates the product of the Hessian with a vector of ones.
+```lua
+local model = nn.Sequential()
+   :add(nn.Linear(2, 2))
+   :add(nn.Sigmoid())
+   :add(nn.Linear(2, 2))
+   :add(nn.Sigmoid())
+
+-- We must collect the parameters, which also creates storage to store the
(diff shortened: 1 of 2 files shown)


cswhjiang:
@bartvm I wrote some testing code based on the rop.lua. I found an weird result. The code is [here](https://gist.github.com/cswhjiang/dc9be98c0963b3124d8c). If we use ````clone````, the norm of Hv2 is always 0. If we use ````copy````, the norm of Hv2 is correct. Is there a bug? Or I wrote the wrong code?

bartvm:
Without having looked too closely at the code, I think that `rParameters2:copy(rParameters_backup)` and  `rParameters2 = rParameters_backup:clone()` do very different things. With `copy` you copy the contents of `rParameters_backup` into `rParameters2`, modifying the latter (which is what you want). By calling `clone` on the other hand, you actually create an entirely new tensor and assign that to the Lua variable name `rParameters2` i.e. you create an entirely *new* `rParameters2` variable, leaving the old one empty (filled with zeros).

cswhjiang:
@bartvm Thanks for your response. I figured it out just now.

bartvm:
Just a note that the latest modules implemented by @cswhjiang should probably be tested against e.g. Theano to make sure that they are correct.

andresy:
Bart, note that you could create a package nnrop where you have a file per nn module (as in nn) and where you define these additional methods in there (the class system allows you to add methods). This way a "require 'nnrop'" would define those additional methods, and the definitions would be cleaner compared to have them in a unique big file.

bartvm:
@andresy Sure, I initially started with mirroring the approach from `hessian.lua`, but I agree that a separate package would be cleaner.

andresy:
yes, i somehow missed the hessian case (i actually did my own package for this at some point), but i think it should also be the same :)

Issue #395, opened 2015-09-22
updateParameters() behaves strangely when manually sharing via set() on parameter tensors


kbullaughey:
I ran into a problem that I'm not sure whether it is a bug or my misuse of `nn`. 

I find it conceptually simpler to manually set up parameter sharing by calling `set()` on parameter tensors I want tied:

    p2:set(p1)
    p3:set(p1)

Where I want p2 and p3 tied to p1

This works fine if I get linearized par and gradPar vectors using `getParameters()` and then manually update my parameter vector. I first get my vectorized parameters:

    par, gradPar = net:getParameters()

And during updates (after forward-backward):

    par:add(gradPar:mul(-learnRate))

However if I use the built-in function:

   net:updateParameters(learnRate)

Then the par vector after I call `updateParameters()` is different than one would expect and convergence is slow and fit is poor.

Here's a full illustration of the problem. First the code:

    require 'nn'
    
    torch.manualSeed(1)
    initialPar = torch.Tensor(12):uniform(-0.1, 0.1)
    x = torch.Tensor(3):uniform(-1, 1)
    y = torch.Tensor(3):uniform(-1, 1)
    outputLine = 1
    
    function makeNet(initialPar)
      local net = {}
      -- I create a simple network that consists of two 3x3 linear maps.
      local g = nn.Sequential()
      g:add(nn.Linear(3,3))
      g:add(nn.Linear(3,3))
      -- Tie parameters of two linear maps together
      local parTensors, gradParTensors = g:parameters()
      parTensors[3]:set(parTensors[1])
      parTensors[4]:set(parTensors[2])
      gradParTensors[3]:set(gradParTensors[1])
      gradParTensors[4]:set(gradParTensors[2])
    
      net.graph = g
      net.par, net.gradPar = g:getParameters()
      net.par:copy(initialPar)
      net.graph:zeroGradParameters()
      net.criterion = nn.MSECriterion()
      return net
    end
    
    function roundTrip(net, x,y)
      net.graph:forward(x)
      net.criterion:forward(net.graph.output, y)
      net.criterion:backward(net.graph.output, y)
      net.graph:backward(x, net.criterion.gradInput)
    end
    
    function log(s)
      print(outputLine .. ": " .. s)
      outputLine = outputLine + 1
    end
    
    function summarize(net, label)
      log("[" .. label .. "] par sum: " .. net.par:sum() .. "; gradPar sum: " .. net.gradPar:sum())
    end
    
    a = makeNet(initialPar)
[...]

PR #400, opened 2015-09-25
Also clone grad weights and biases

douwekiela:
As per https://github.com/torch/optim/issues/73, gradWeight and gradBias should also be cloned.


fmassa:
I think this should be reverted, because this example uses `updateParameters`, as discussed in https://github.com/torch/nn/issues/395 .

PR #401, opened 2015-09-26
Add eps correctly

jzbontar:
Found a bug in the new `BCECriterion`. The `updateOutput` method returns `NaN`
instead of zero when prediction is correct. The bug affects `FloatTensor` and
`CudaTensor`, but not `Tensor`.

    require 'torch'
    require 'nn'

    y = torch.FloatTensor{1, 0}
    t = torch.FloatTensor{1, 0}
    crit = nn.BCECriterion():float()

    -- returns nan, expected 0
    print(crit:forward(y, t))
    
The problem was that `eps` was handled incorrectly. Concretely,

    x:add(1 + eps)
    
is not the same as

    x:add(1):add(eps).
    
The first example just adds one when `x` is `FloatTensor` or `CudaTensor`.

### DIFF
--- a/BCECriterion.lua
+++ b/BCECriterion.lua
@@ -49,7 +49,7 @@ function BCECriterion:updateOutput(input, target)
     output = torch.dot(target, buffer)
 
     -- log(1 - input) * (1 - target)
-    buffer:mul(input, -1):add(1 + eps):log()
+    buffer:mul(input, -1):add(1):add(eps):log()
     if weights ~= nil then buffer:cmul(weights) end
 
     output = output + torch.sum(buffer)
@@ -86,7 +86,7 @@ function BCECriterion:updateGradInput(input, target)
 
     buffer:resizeAs(input)
     -- - x ( 1 + eps -x ) + eps
-    buffer:add(input, -1-eps):cmul(input):add(-eps)
+    buffer:add(input, -1):add(-eps):cmul(input):add(-eps)
 
     gradInput:resizeAs(input)
     -- y - x


soumith:
the reason seems to actually be that eps is defined as 1e-12. that is too small for float precision, and you might still generate nans.

jzbontar:
Hi Soumith,

I don't think 1e-12 is too small for floats. 1e-12 is a giant if you are at 0. 
Did you know that there are, according to the following program, 730,643,660 floats
between 0.0 and 1e-12.

    #include <stdio.h>
    #include <math.h>
    int main(void) {
        int cnt = 0;
        for (float f = 0.0; f < 1e-12; f = nextafterf(f, 1.0), cnt++)     
            ;
        printf("%d\n", cnt - 1);
        return 0;
    }


If anything, I vote to decrease epsilon. The smallest *leet* number that would
work is 1.337e-39, but alas it produces inf's in the gradient. The next smallest
is 1.337e-38 and it works, but is not as *leet* as 1.3e-37. I therefore propose
to change epsilon from 1e-12 to 1.3e-37. Who's with me?

PS Soumith, I was not able to come up with an example where the proposed commit 
produces nan's or inf's. Can you share your example?

hughperkins:
I havent read earlier than your most recent comment, so i may have misunderstood, but I reckon eps is related to the number of significant figures in the mantissa, rather than how many zillion decimal places you can put in the exponent.  The following for example returns inf:

```
require 'torch'

a = torch.FloatTensor({1})
a_plus_delta = torch.FloatTensor({1 + 1e-12})

local diff = torch.add(a, -1, a_plus_delta)
local one = torch.FloatTensor({1})
local one_over_diff = torch.cdiv(one, diff)
print('one_over_diff', one_over_diff)
```

Alternativelhy, the folowing gives nan:
```
require 'torch'

a = torch.FloatTensor({1})
a_plus_delta = torch.FloatTensor({1 + 1e-12})

local delta = torch.add(a, -1, a_plus_delta)
local zero = torch.FloatTensor({0})
local zero_over_delta = torch.cdiv(zero, delta)
print('zero_over_delta', zero_over_delta)
```

soumith:
@jzbontar thanks. It's a critical fix in an important criterion.

It's not that there aren't floats so small, but you'll have to take into account the float semantics (adding large number + small number, when will the large number actually have a guaranteed change).

PR #403, opened 2015-09-26
THTensor_(*) for abs, sqrt and tanh

szagoruyko:
removing legacy code

### DIFF
--- a/generic/Abs.c
+++ b/generic/Abs.c
@@ -8,9 +8,7 @@ static int nn_(Abs_updateOutput)(lua_State *L)
   THTensor *output = luaT_getfieldcheckudata(L, 1, "output", torch_Tensor);
 
   THTensor_(resizeAs)(output, input);
-
-  TH_TENSOR_APPLY2(real, output, real, input, \
-                   *output_data = fabs(*input_data);)
+  THTensor_(abs)(output, input);
   return 1;
 }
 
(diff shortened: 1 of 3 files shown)


soumith:
Thanks

fmassa:
There is actually a performance loss for `Sqrt` and `Tanh` with this PR.
Current pointwise `torch` functions are not paralellized, see https://github.com/torch/torch7/blob/master/lib/TH/generic/THTensorMath.c#L1854 and https://github.com/torch/torch7/blob/master/lib/TH/generic/THTensorMath.c#L1802
I agree though that `nn` functions should maximally rely on `torch` functions. In this case, shouldn't we add parallelization on pointwise functions in core `torch` ?
With current framework, I think it would require manually writing repeated code for each pointwise function, as it seems that one can not use preprocessor macros (`#pragma omp parfor`) inside a macro definition.

hughperkins:
```
With current framework, I think it would require manually writing repeated code for each pointwise function, as it seems that one can not use preprocessor macros (#pragma omp parfor) inside a macro definition. 
```

Hmmm, plus one point in favor of Jinja2/Lua templates :-P  (Mind you, this can also be argued to be true also for c++ templates too plausibly).

fmassa:
Well, there are workarounds to allow parallel functions inside a macro.
Here is first thing I tried that works (at least on gcc 4.6.3).  https://github.com/fmassa/torch7/commit/c0a665495c0c90d3a18763c12db778e89259d0dc
There are lots of compiler warnings (maybe because we are calling functions like `tanh` with float arguments, but it should be `tanhf` instead), but it seems to work.
Any better solutions that work on C99 ?

hughperkins:
seems like it is because you are sending a pointer to a function that takes a double, into a function that is expecting a pointer to a function that takes a float.  Sent a pull request that seems to remove build warnings (no idea if it runs correctly or not though :-P )

szagoruyko:
we have a fix for all math foo/foof cases, this PR was a preparation for it

PR #405, opened 2015-09-27
Adds SpatialDeconvolution

fmassa:
This is the C implementation for https://github.com/torch/cunn/pull/122.

It contains unit tests and a simple doc. Special thanks to @andreaskoepf for the fix in the `accGradParameters`, and @szagoruyko for the initial CUDA version upon which this implementation is based.

`im2col` and `col2im` were taken from [Caffe](https://github.com/BVLC/caffe/blob/master/src/caffe/util/im2col.cpp) implementation and are not parallelized. I could eventually reuse original [unfolded_copy](https://github.com/torch/nn/blob/master/generic/SpatialConvolutionMM.c#L67) from `SpatialConvolutionMM` if wanted.

Also, I made it consistent with other convolutional modules, as the current CUDA PR inverses `nInputPlane`and `nOutputPlane`.

I'll prepare a patch for the CUDA PR to compare both implementations.


soumith:
it is soooo wrong to call this stuff Deconvolution, but it seems to be a trend.

Deconvolution is something completely different: https://en.wikipedia.org/wiki/Deconvolution

soumith:
thanks a lot Francisco.

jonathantompson:
Actually, I really think we need to change the name here.  

Deconvolution has a pretty specific meaning in either image processing (as you pointed out @soumith) OR deep learning already (i.e.  Zeiler, M., Taylor, G., Fergus, R.: Adaptive deconvolutional networks for mid and high level feature learning. In: ICCV (2011)).

Calling this a SpatialDeconvolution is only going to lead to a LOT of confusion.

soumith:
Ok, I'll change the Spatial and Volumetric Deconvolution naming. Any suggested names @jonathantompson ?

jonathantompson:
Well, it's a upsampling layer with filtering...  SpatialUpsamplingFiltered ???  That sounds stupid.  Maybe @fmassa has a better alternative.

From the paper:

"In a sense, upsampling with factor f is convolution with a fractional input stride of 1/f. So long as f is integral, a natural way to upsample is therefore backwards convolution (sometimes called deconvolution)"

All I can tell you is that I don't know anyone who would call that "deconvolution", which would require some form of inference step to invert the linear convolution. I think some factual mistakes in this paper unfortunately made it through peer-review :-(

soumith:
One way is to factor it into SpatialConvolution, where this code-path is activated if the stride is 0 > stride < 1

soumith:
but we have to check that it's exactly 1 / f, where f is an integer. Which is a bit annoying.

koraykv:
Don't we already have a SpatialFullConvolution layer that does pretty much the same thing (I am sure this is much faster though)? The name Full comes from matlab convention categorizing convolutions into valid, same, full: valid for no padding, same padding such that output size is same as input size and full such that the output size is larger than input size.

soumith:
You're right. SpatialFullConvolution it is. We had this for ages, forgot of it's existence.

soumith:
we don't have a GPU version of SpatialFullConvolution. I'll work on the name-change + adapting the existing GPU implementation of SpatialDeconvolution

fmassa:
@koraykv Indeed, it's precisely the same thing as 'SpatialFullConvolution`, I hadn't noticed that :)
Thanks @soumith for refactoring `deconv -> fullconv`.

andreaskoepf:
Maybe it is a trivial question: But could somebody please explain the direct equivalence of 2d 'full' convolution as defined by [matlab](http://de.mathworks.com/help/matlab/ref/conv2.html) and the implementation that originally was called 'Deconvolution' (similar to the [Caffe DecovLayer](http://caffe.berkeleyvision.org/doxygen/classcaffe_1_1DeconvolutionLayer.html))?

The 'Deconvolution' operation was able to share (transposed) weights with a nn.SpatialConvolution module. The *UNUSUAL* part of a deconv operation is the [col2im](https://github.com/torch/nn/blob/637076937f2f9fa27e7463f3ac494d954f25f77b/generic/SpatialFullConvolution.c#L33) operation.

andreaskoepf:
For forward convolution operations (regardless if same, full or valid) one element of a feature plane is one generated by a filter evaluated over a one window at a position spanning all input planes. But 'deconvolution' does something different: it is a 1x1 convolution that produces kw*kh*output_planes feature planes.... this were the original kw*kh*input_planes of the forward SpatialConvolution -> in order to stuff them back may overlapping results are 'unpacked" and combined by col2im ... I do not see in which way this could be equivalent to a normal forward 'full' convolution.

andreaskoepf:
@jonathantompson you said 'some form of inference step to invert the linear convolution' would be missing but I do not think this is the case: If conv -> deconv layers are used in an auto-encoder setup the network training is effectively the 'inference step' ... and since the 'deconv' layer actually 'inverts' the convolution operation with the same number of weights (frequently even the same weights transposed) I do not see a reason  why calling it 'deconv' would be wrong. Also I think deconv is special and different form normal convolution since it has to combine several output together in order to restore the original input-dimensionality.

A couple of weeks ago I desrcibed the [decovolution operation](https://github.com/torch/cunn/pull/122#discussion_r40151044).

Also cross-check with [SpatialConvolutionMM.c](https://github.com/torch/nn/blob/master/generic/SpatialConvolutionMM.c) .. here I guess the functions unfolded_acc & unfolded_copy are related to im2col and col2im -> a main difference between SpatialConvolutionMM is that the accumulating operation is used during the backward pass but in the current impl of SpatialFullConvolution it is used in the [forward pass](https://github.com/torch/nn/blob/637076937f2f9fa27e7463f3ac494d954f25f77b/generic/SpatialFullConvolution.c#L143).

andreaskoepf:
Regarding naming: If you do not like 'deconvolution', what about SpatialBackwardConvolution - this would stress the fact that forward & backward passes are swapped...

fmassa:
@andreaskoepf I think the previous implementation of `SpatialFullConvolution` and `SpatialDeconvolution` are equivalent (up to the padding, which was not present in the original version of `SpatialFullConvolution`).

As you said, the forward in the previous `SpatialFullConvolution` calls [THTensor_(conv2Dmv)(output, 1.0, 1.0, input, tweight, dH, dW, "F", "C");](https://github.com/torch/nn/blob/b381c7c4a49ac4646029bf7f7e80756b5f20a3bc/generic/SpatialFullConvolution.c#L55).
Now look at the previous implementation of `SpatialConvolution.c`, the `updateGradInput` calls precisely the same function [THTensor_(conv2Dmv)(gradInput, 0.0, 1.0, gradOutput, tweight, dH, dW, "F","C");](https://github.com/torch/nn/blob/b381c7c4a49ac4646029bf7f7e80756b5f20a3bc/generic/SpatialConvolution.c#L113). Same thing applies for the backward.

Also, `SpatialConvolutionMM` is equivalent to `SpatialConvolution`, so this reasoning also applies to `SpatialDeconvolution`.

As both functions are equivalent, we should probably stick with the old naming.

jonathantompson:
https://en.wikipedia.org/wiki/Deconvolution

@andreaskoepf Deconvolution has nothing to do with trying to "undo" convolution + pooling.  It doesn't have anything to do with upsampling.  It is simply the name given to the operator that inverts a linear convolution.  This stage does not try to invert the convolution as that would require ACTUALLY solving the linear inverse. Also, unless you explicit constrain the convolution weights to be orthonormal (with the authors did not do), taking the transpose of the convolution is also not inverting.  Nothing about this layer is an inverse or a deconvolution.

Now don't get me wrong: people have been using this trick for years (which is perhaps my main gripe about this paper); performing in-network upsampling and letting the network choose how to properly interpolate spatially has been around about as long as convolutional networks (as has networks without fully-connected stages - but that's a whole other issue).  But to call it "deconvolution" is just plain wrong, and even somewhat misleading, and this should have been caught in peer review.

"SpatialBackwardConvolution"  This sounds great to me!

andreaskoepf:
Ok, the 'trick' of the SpatialFullConvolution module is that it uses transposed weights by default. As long as you do not view at the weight tensor it simply behaves like a full convolution, so the name seems to be ok.

lahwran:
Hey, I'd appreciate a link to this conversation in the docs that describe "full convolution"; I hadn't seen the term before and only knew of deconvolution from papers that used the term to mean "upsampling". I agree that the term is better, but not knowing what it meant, I was confused as to how Convolution and Full Convolution were different; was something less than full about the normal convolution implementation? A simple note explaining that "full convolution" refers to the operation sometimes known as upsampling, or occasionally incorrectly as "deconvolution", would also do the trick.

PR #407, opened 2015-09-27
LeakyReLU

rotmanmi:
Hi,

I've added a LeakyReLU module for nn.

Best,

Michael.

### DIFF
--- a/LeakyReLU.lua
+++ b/LeakyReLU.lua
@@ -1,0 +1,25 @@
+local LeakyReLU, parent = torch.class('nn.LeakyReLU','nn.Module')
+
+function LeakyReLU:__init(negval,ip)
+   parent.__init(self)
+   self.negval = negval or (1/100)
+   if  (negval and type(negval) ~= 'number') then
+      error('nn.LeakyReLU(value)')
+   end
+   -- default for inplace is false
+   self.inplace = ip or false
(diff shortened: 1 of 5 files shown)


szagoruyko:
Nice PR, thanks! here is a shorter one though:
```lua
local LeakyReLU, parent = torch.class('nn.LeakyReLU', 'nn.PReLU')

function LeakyReLU:__init(p)
  parent.__init(self)
  self.weight:fill(p)
  self.gradWeight:fill(0)
end

function LeakyReLU:__tostring__()
  return torch.type(self) .. string.format('(%g)', self.weight[1])
end

function LeakyReLU:accGradParameters(input, gradOutput, scale)
end

function LeakyReLU:zeroGradParameters()
end

function LeakyReLU:updateParameters(learningRate)
end
```

rotmanmi:
I think PReLU is not done inplace.

szagoruyko:
it is better to add inplace option to PReLU then

rotmanmi:
well, PReLU allows for negative derivative on the negative side and then
you cannot use it inplace. as this is dynamical, you cannot choose it
beforehand.

soumith:
@rotmanmi this looks fine. there are some merge conflicts, and the commits have to be squashed.

soumith:
I see you've opened https://github.com/torch/nn/pull/409/files , closing this then...

PR #408, opened 2015-09-28
Remove spurious mallocs from nn.Linear

adamlerer:
Remove a `new()` that introduces lots of spurious mallocs when the input tensor size is not constant.


soumith:
thanks Adam.

PR #411, opened 2015-10-01
Use custom range in HardTanh and mask it as Clamp

jhjin:
allows HardTanh to use a custom interval for linear region when min/max values
are specified in declaration. Thus its implementation can be shared with new
Clamp function.


szagoruyko:
I made a couple of comments, the rest looks good, seems quite useful

jhjin:
@szagoruyko  Thanks for your time. I updated the code to use the consistent ordering in arguments and considered backward compatibility. The same as for [cunn](https://github.com/torch/cunn/pull/149).

soumith:
Thanks a lot Jonghoon.

PR #413, opened 2015-10-03
Adds SpatialCrossLRN

fmassa:
Implements [Local Response Normalization between different channels](https://code.google.com/p/cuda-convnet2/wiki/LayerParams#Local_response_normalization_layer_%28across_maps%29).
Could eventually be adapted to use the CUDA implementation in [inn](https://github.com/szagoruyko/imagine-nn/blob/master/SpatialCrossResponseNormalization.cu).
I named it as in `cudnn`, because it's shorter than `inn`s name.


fmassa:
Hi Sergey,

Ok, I can rename `scale`, I just took the name from `inn`.

Regarding memory requirements, this PR only needs a buffer of the size of the `input` to store the `scale`, and the other buffers in `updateGradInput` doesn't require much memory, on the order of one element of `input`. I hesitated about adding a bigger buffer which would avoid having a for loop in the backward for the batch elements though...

I compared this implementation with `inn` and they seem to be equivalent.

The problem at the moment is that this version is incredibly slow, 700x slower than `inn` version on GPU forward, much slower on backward. I tried to optimize it a bit by removing the for loop on forward and replacing by a matrix multiplication, see [here](https://github.com/fmassa/nn/commit/0c2e9e36e9127e140b3edffe1ad0fd577f4131eb), but on CPU it's slower, but may be faster on GPU. (Actually, at least on my machine on CPU, most of the time is spent doing `pow(-beta)`. I found out that `pow` is super slow on CPU for `beta` different than 0,1,2, and replacing it by `log(x):mul(-beta):exp())` brings a speedup of 80% compared to simply doing `pow(x,-beta)`...)

szagoruyko:
@fmassa why care about GPU? we have kernels for that, can copy inn's version to cunn and put 'if' as here https://github.com/szagoruyko/nn/blob/batchnorm/SpatialBatchNormalization.lua#L78

fmassa:
Agreed.
I'll rename `scale` and adapt this version to accept the `inn` GPU implementation. Will prepare a patch later on today.

soumith:
pingping

fmassa:
Will address the points from Sergey in the beginning of January.

szagoruyko:
look like it can be merged

soumith:
Thanks both!

PR #416, opened 2015-10-05
Remove unnecessary mallocs from BCECriterion

fmassa:
When using `weights` with 2D `targets`, `:size()` is unnecessarily called, allocating a new tensor at every `:forward` and `:backward`.


dominikgrewe:
Good catch!

Issue #418, opened 2015-10-06
simple layers doc missing CDiv


eulerreich:
Just a heads up that the [simple layers doc](https://github.com/torch/nn/blob/master/doc/simple.md#nn.simplelayers.dok) doesn't have an entry for `CDiv` despite appearing to link to one.

fmassa:
@eulerreich thanks !
Fixed via https://github.com/torch/nn/pull/419

PR #419, opened 2015-10-06
Remove CDiv from doc

fmassa:
There is currently no module named `CDiv` in `nn`.
Maybe there were plans to add it but then it was forgotten ? Anyway, seems to be redundant given that we have `CMul`, and also more unstable than multiplications (divisions by zero and huge effects on output by varying a bit the weights)
Refers to https://github.com/torch/nn/issues/418


soumith:
oops. thanks Francisco.

fmassa:
@soumith you are welcome !

PR #422, opened 2015-10-07
Add note on training shared models

fmassa:
Usually when training models with shared parameters, by reflex we share both the `parameters` and the `gradParameters` (if not, `optim` will complain about size mismatches).
But sharing the `gradParameters` while using an optimization procedure that iterates over the modules (like `updateParameters`, or even Facebook's `nn.Optim' I think) is wrong. See for example https://github.com/torch/nn/issues/395

Adding it to the docs to make it clearer.

### DIFF
--- a/doc/overview.md
+++ b/doc/overview.md
@@ -21,7 +21,7 @@ high level way to train the neural network of choice, even though it is
 easy with a simple for loop to [train a neural network yourself](training.md#nn.DoItYourself).
 
 ## Detailed Overview ##
-This section provides a detailed overview of the neural network package. First the omnipresent [Module](#nn.overview.module) is examined, followed by some examples for [combining modules](#nn.overview.plugandplay) together. The last part explores facilities for [training a neural network](#nn.overview.training).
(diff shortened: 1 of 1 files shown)


doc/overview.md
```diff
@@ -141,3 +141,61 @@ end
...
+-- our optimization procedure will use all the parameters at once, because
+-- it requires the flattened parameters and gradParameters Tensors. Thus,
+-- we need to share both the parameters and the gradParameters
+mlp = nn.Sequential()
+linear = nn.Linear(2,2)
+-- need to share the parameters and the gradParameters as well
+linear_clone = linear:clone('weight','bias','gradWeight','gradBias')
+mlp:add(linear)
+mlp:add(linear_clone)
+parameters, gradParameters = mlp:getParameters()
```
albanD: Its really minor but wouldn't it be clearer to call these `params` and `gradParams` to be consistent with the names used in the function below?

fmassa: Yes, I'll change that.
I'll just wait for the others to comment before doing the changes.

soumith: can be merged after changes requested by Alban

fmassa:
@soumith @albanD just updated with the changes proposed by Alban and squashed the commits.

soumith:
thanks Francisco.

PR #424, opened 2015-10-10
Reduced peak memory requirements of Module.flatten()

mys007:
This patch decreases peak memory usage in Module.flatten(), partially reverting some of the changes made in 7105419c8382c3a9c5c257661c8ab49872fdcddc . 

- maskParameters: single tensor creation instead of temporary cast and clone.
- compactOffsets: lowered precision to int (originally it was just float). I even suggest decreasing it to short (are more than 2^15 parameters realistic?).
- adds back collectgarbage calls removed in the last refactoring to free all the temporary tensors created in flatten()

### DIFF
--- a/Module.lua
+++ b/Module.lua
@@ -222,8 +222,8 @@ function Module.flatten(parameters)
       tensorsCompact = tensorsCompact and isCompact(tmp)
    end
 
-   local maskParameters  = flatParameters:byte():clone()
-   local compactOffsets  = flatParameters:long():cumsum(1)
+   local maskParameters  = torch.ByteTensor(nParameters):copy(flatParameters)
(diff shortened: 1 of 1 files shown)


mys007:
Actually, the root of my memory issues was fixed by b3fc443f57e769daf84b39590dc901fe647fec4f , but I still think my patch is useful.

soumith:
Assumption (2) is wrong. It used to be a Float Tensor, and at some point we had a model that was large enough that it's offsets were going wrong. So such models do exist (easily if you have a local convolution layer).

soumith:
wontmerge because of the Int limits assumption

mys007:
Oh, sorry, I completely forgot about this PR. It was really a misconception with my int/short, I misunderstood the algorithm. And otherwise the patch is not that needed.

PR #426, opened 2015-10-11
Speedup and reduce memory usage in Normalize

fmassa:
It's almost 40x faster on CPU, and doesn't use extra memory. Should
be faster on GPU as well, but I haven't benchmarked.

* Instead of creating a temporary matrix `b1*b2` of size `(n,d,d)`, rearrange the computations to avoid creating this huge matrix which was eating all the memory.
* `forward` and `backward` supports varying dimensionality (which wasn't the case before because of the `eye` matrix)

Should be of interest to @ffmpbgrnn and @bamos.

### DIFF
--- a/Normalize.lua
+++ b/Normalize.lua
@@ -49,23 +49,29 @@ function Normalize:updateGradInput(input, gradOutput)
 
   local n = input:size(1) -- batch size
   local d = input:size(2) -- dimensionality of vectors
-  -- compute diagonal term
-  self.eye = self.eye or torch.eye(d):typeAs(input):view(1,d,d)
-  local eyeExpand = self.eye:expand(n,d,d)
-  self.diag = self.diag or self.eye.new()
-  self.diag:cmul(eyeExpand, self.normp:view(n,1,1):expand(n,d,d))
-  -- compute cross term
+
+  -- compute diagonal term with gradOutput
+  self.gradInput:resize(n,d,1)
+  gradOutput = gradOutput:view(n,d,1)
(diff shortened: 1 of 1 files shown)


fmassa:
Just did a quick test on the GPU, backward seems to be much faster than the previous version as well.
For an input of dimensions `(32,4096)`, this version is 100x faster than the old one (and the old version used 2GB of temporary buffers).

bamos:
Hi, I ran @ffmpbgrnn's test from https://github.com/torch/nn/pull/341 for some simple profiling of this PR on a Tesla K40 GPU and 3.70GHz CPU. The speedup is amazing!

```Lua
require 'nn'
require 'cutorch'
require 'cunn'

local module = nn.Normalize(2):cuda()
module:fastMode(false)
local input = torch.rand(64, 2400):cuda()
local t = torch.Timer()
for i = 1, 100 do
    module:forward(input)
    module:backward(input, input)
    print(i)
end
print(t:time().real/100)
```

# Current Master
+ Commit: b80bda2ee57354e049cc446136b2e28bd2243c4e
+ md5sum of `Normalize.lua`: 4b1f217a796f00ff8cfe575da8e4a409
+ CPU mean execution time: 3.82 s
+ K40 mean execution time: 0.108 s

# This PR
+ md5sum of `Normalize.lua`: a9c7d3f642c08de361369be201234fca
+ CPU: 0.00634 s
+ K40: 0.000624 s

soumith:
this is super awesome. If unit tests pass, and it's exactly the same implementation as before, why not!!!

PR #429, opened 2015-10-12
Replace plot of SoftPlus function

andreaskoepf:
Created a replacement visualization of the SoftPlus function
for the docs that shows the function and its first derivative.


soumith:
thanks andreas.

PR #432, opened 2015-10-16
Fix modification of output in nn.Normalize

kosklain:
Currently `nn.Normalize` has the following strange behaviour:

```lua
m = nn.Normalize(2)
y = m:forward(torch.randn(3))
> =y
-0.1459
 0.8920
-0.4278
[torch.DoubleTensor of size 3]

> = m:forward(torch.randn(3))
 0.1877
-0.8429
-0.5043
[torch.DoubleTensor of size 3]

> =y
 0.1877 -0.8429 -0.5043
[torch.DoubleTensor of size 1x3]
```

Making it hard to keep track of the output of the module. This PR fixes this.

### DIFF
--- a/Normalize.lua
+++ b/Normalize.lua
@@ -6,17 +6,18 @@ function Normalize:__init(p,eps)
   assert(p > 0, p..'-norm not supported')
   self.p = p
   self.eps = eps or 1e-10
+  self._output = torch.Tensor()
+  self._gradInput = torch.Tensor()
 end
 
 function Normalize:updateOutput(input)
   assert(input:dim() <= 2, 'only 1d layer supported')
-  local is_batch = true
+  local input_size = input:size()
   if input:dim() == 1 then
     input = input:view(1,-1)
-    is_batch = false
   end
 
-  self.output:resizeAs(input)
+  self._output:resizeAs(input)
 
   self.norm = self.norm or input.new()
   self.normp = self.normp or input.new()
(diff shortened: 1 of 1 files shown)


fmassa:
Indeed, I hadn't thought about this corner case. Thanks !

You could avoid creating the `self._output`  and `self._gradInput` by using instead `self.output:resize(input_size)`, but I don't know which solution is better. (maybe using `resize` is riskier...)

@soumith except the 3 spaces and the camelCase policies, are there other coding styles that would be preferred in `torch` ?

(BTW, I should probably change the strict camelCase that I used in https://github.com/torch/nn/pull/413, as well as fix the same view behaviour as this one)

kosklain:
I usually prefer having `self.output` explicitly avoiding `resize`s `view`s and so on to make things clear to the reader (also conceptually, `self.output` will be 'just one thing' this way), but I can change that if you prefer.

W.r.t. name conventions I prefer camelCase for lua, but I followed the ones in this file (I saw  the `is_batch` variable).

soumith:
Thanks @kosklain

bamos:
Hi, commit d3da8ba breaks my openface code, is it possible to get the the commit reverted until this is further debugged? See a user's report at https://github.com/cmusatyalab/openface/issues/26 showing they had to modify `nn.Normalize` to fix the issue. Will look into myself ASAP.

fmassa:
It happens because `_output` and `_gradInput` are not lazily initialized in the `updateOutput` and `updateGradInput`.
We can either check that they exist in each one of those functions (as usually done, `self._output = self._output or input.new()`, or use the solution proposed by @dominikgrewe https://github.com/torch/nngraph/pull/88

dominikgrewe:
I'd just go with the lazy initialization for now. I'd eventually like to see a cleaner solution, but it'd be good to get this fixed quickly.
@kosklain said he's going to prepare a fix.

soumith:
Reverted, will wait for @kosklain 's backward-compatible version of the same PR.

PR #436, opened 2015-10-20
Fix modification of output in nn.Normalize (with lazy-loading of internal tensors)

kosklain:
This is the backwards-compatible version of #432


bamos:
Hi, this fixes the backwards-incompatible change from #432 in my project.

PR #437, opened 2015-10-21
Add randomized leaky rectified linear unit (RReLU)

andreaskoepf:
In RReLUs, the slopes of negative parts are randomized in
a given range during training, and fixed during testing.
Participants of the Kaggle National Data Science Bowl (NDSB) 
competition reported that RReLUs could reduce overftting due to 
their randomized nature. Informally they were called 'Insanity Layer'.
Paper: "Empirical Evaluation of Rectified Activations in
Convolutional Network" http://arxiv.org/abs/1505.00853

TODO:
- [x] lua impl prototype
- [x] c cunn impl
- [x] c impl for nn
- [x] documentation
- [x] tests for inplace and train code


andreaskoepf:
Obviously the travis build fails with a seg fault while running the tests. On my machine I can execute all tests without problems. Even when I run my new RReLU test several thousand times in a loop (which tests a wide range of random input tensor sizes) I do never see a seg fault. I would appreciate it If somebody could do a quick code review and let me know what might be wrong!

soumith:
dont worry about the segfault for now. I looked at your code and nothing looks wrong.

I'll also transition Travis to their new build infra, their legacy infra is a bit buggy overall.

andreaskoepf:
For the records: I found my bug - it is a bit embarrassing - but I somehow forgot the return statements in RReLU_updateOutput() and RReLU_updateGradInput().

soumith:
Thanks a lot Andreas. Very nicely done.

PR #443, opened 2015-10-24
Fixed Segmentation fault on very big data sizes

anastasiuspernat:
When data size exceeds dozens of GB int variables can’t store data size anymore so memset and memcpy fail with Segmentation fault errors - reproduced when using convolutional neural networks for image processing - in neuralstyle and deepstyle implementation, with images larger than 2000x2000.

You can find that using gdb on lua.

For example [this bug](https://github.com/jcjohnson/neural-style/issues/24) has been closed because of this issue. [This one](https://github.com/kaishengtai/neuralart/issues/11) has not been resolved as well.


soumith:
Thanks a lot for the fix

anastasiuspernat:
Thanks for such a quick response!

lolz0r:
Nice fix!

Issue #440, opened 2015-10-23
division by zero in normalize.lua (L60)


rlebret:
https://github.com/torch/nn/blob/master/Normalize.lua#L60
can divide by zero. pow(p-2) divides for p<2.
It could be changed to 
  self.buffer:abs(input):add(self.eps):pow(self.p-2):cmul(input)
to fix this.

soumith:
if you send a PR for it, happy to accept.

fmassa:
I'll prepare a patch for it

soumith:
should be fixed now.

PR #444, opened 2015-10-24
Fix possible division by zero in Normalize

fmassa:
When `p < 2`, if the input is 0 there was division by zero, as pointed out by @rlebret in https://github.com/torch/nn/issues/440.
Adds a dampening factor `self.eps` to avoid this.
Also, small optimizations for different `p`.


soumith:
merci!

PR #446, opened 2015-10-26
Fix CosineDistance to handle non-contiguous input

hohoCode:
Thanks for the batch version of CosineDistance. Just one quick fix here so it can handle the non-contiguous input. Just for backward-compatibility purpose since the original/old CosineDistance.lua can handle non-contiguous input without problems, and also it is a little bit troublesome for end users to make all input tensors contiguous by themselves. Non-contiguous input cases could happen unfortunately.. for example in this recent [model](http://github.com/hohoCode/textSimilarityConvNet). I tested this fix on my end, it works. 

Thanks again for the batch version update.

### DIFF
--- a/CosineDistance.lua
+++ b/CosineDistance.lua
@@ -4,12 +4,25 @@ function CosineDistance:__init()
    parent.__init(self)
    self.gradInput = {torch.Tensor(), torch.Tensor()}
 end 
- 
+
+local function makeContiguous(self, input1, input2)
+   if not input1:isContiguous() then
+      self._input1 = self._input1 or input1.new()
+      self._input1:resizeAs(input1):copy(input1)
+      input1 = self._input1
+   end
+   if not input2:isContiguous() then
+      self._input2 = self._input2 or input2.new()
+      self._input2:resizeAs(input2):copy(input2)
+      input2 = self._input2
+   end
+   return input1, input2
+end
+
 function CosineDistance:updateOutput(input)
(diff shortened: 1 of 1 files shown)


CosineDistance.lua
```diff
@@ -8,6 +8,9 @@ end
 function CosineDistance:updateOutput(input)
    local input1, input2 = input[1], input[2]
 
+   if not input1:isContiguous() then input1 = input1:clone() end
```
fmassa: Hi,
Could you use a temporary buffer to avoid memory allocations at every forward/backward pass ? Something like
```lua
if not input1:isContiguous() then
  self._input1 = self._input1 or input1.new()
  self._input1:resizeAs(input1):copy(input1)
  input1 = self._input1
end
```
You could reuse this `_input1`/`_input2` buffer in the backward pass as well.
Or maybe wrapping it in a function like https://github.com/torch/nn/blob/master/SpatialConvolution.lua#L65 ?
Thanks !

hohoCode: Thanks @fmassa . Updated, now put a local function wrap there to reuse the buffer for the backward pass as suggested. Tested on my end. See if this is good.

fmassa: I think it's good! Thanks !

soumith:
Thanks Hua!

hohoCode:
You're welcome. Thanks for this great tool! :)

PR #450, opened 2015-10-27
Update doc for MarginCriterion

albanD:
Update the MarginCriterion's doc to follow the implementation introduced by 9176533


soumith:
Thanks!

PR #452, opened 2015-10-30
Added batch mode support in DotProduct

sensout:
Added batch mode support (table of matrices as input) in DotProduct.
Followed logic from CosineDistance.
Updated doc accordingly.

### DIFF
--- a/DotProduct.lua
+++ b/DotProduct.lua
@@ -3,27 +3,53 @@ local DotProduct, parent = torch.class('nn.DotProduct', 'nn.Module')
 function DotProduct:__init()
    parent.__init(self)
    self.gradInput = {torch.Tensor(), torch.Tensor()}
-   self.output=torch.Tensor(1)
 end 
  
-function DotProduct:updateOutput(input,y)
-   self.output[1] = input[1]:dot(input[2])
+function DotProduct:updateOutput(input)
+   local input1, input2 = input[1], input[2]
+   if input1:dim() == 1 then
+      -- convert non batch input to batch input
+      input1 = input1:view(1,-1)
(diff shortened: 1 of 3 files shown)


soumith:
cc: @fmassa for review.

DotProduct.lua
```diff
@@ -3,27 +3,53 @@ local DotProduct, parent = torch.class('nn.DotProduct', 'nn.Module')
...
+
+   if not_batch then
+      -- unbatch gradInput
+      self.gradInput[1] = gw1:select(1,1)
+      self.gradInput[2] = gw2:select(1,1)
+   end
+
+   -- fix for torch bug 
+   -- https://github.com/torch/torch7/issues/289
+   self.buffer:resize()
```
fmassa: I think this is not necessary here.
It's used in `CosineDistance` because I reused `self.buffer` in the backward and forced it to have zero stride at some point. But here it's fine I think.

fmassa:
Could you also add unit tests ? Currently there is no unit test for `DotProduct`, it would be great if we could add some.
Apart from that (and the note I left), it looks great.

sensout:
I removed the useless resize and included UT.

sensout:
I removed the useless `buffer:resize()`and added UTs.

soumith:
@fmassa does this look good now?

fmassa:
@soumith I think it's good to go. Thanks @sensout !

soumith:
Squashed commits and merged via: https://github.com/torch/nn/commit/ea649de80623243dfc047e4a7d37f882eb426f17

soumith:
Thanks a lot Vincent.

PR #453, opened 2015-11-02
Rewording the documentation for `nn.Reshape`

allanqunzi:
Rewording the documentation for `nn.Reshape` to make it more clear.

### DIFF
--- a/doc/simple.md
+++ b/doc/simple.md
@@ -566,7 +566,7 @@ module = nn.Reshape(dimension1, dimension2, ... [, batchMode])
 
 Reshapes an `n x p x q x..`  Tensor into a `dimension1 x dimension2 x...` Tensor, taking the elements column-wise.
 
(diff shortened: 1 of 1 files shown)


doc/simple.md
```diff
@@ -564,9 +564,13 @@ module = nn.Reshape(dimension1, dimension2, ... [, batchMode])
...
+Reshapes an `n x p x q x..`  Tensor into a `dimension1 x dimension2 x...` Tensor, taking the elements column-wise.
 
-The optional last argument `batchMode`, when `true` forces the first dimension of the input to be considered the batch dimension, and thus keep its size fixed. This is necessary when dealing with batch sizes of one. When `false`, it forces the entire input (including the first dimension) to be reshaped to the input size. Default `batchMode=nil`, which means that the module considers inputs with more elements than the produce of provided sizes, i.e. `dimension1xdimension2x...`, to be batches.
+The optional last argument `batchMode`, when `true` forces the first dimension of the input to be considered as the batch dimension (i.e., the size of the first dimension will be treated as the batch size), and thus keep the size of the first dimension fixed and only reshape the other dimensions. When `false`, it forces the entire input (including the first dimension) to be reshaped to the input size. The default for `batchMode` is `nil`, which means that this moudle will reshape all the dimensions only when the two statements below
```
soumith: typo here. moudle -> module.

allanqunzi: Thanks, changed it.

andreaskoepf:
The 'batch-auto-detection-heuristic' of Reshape is IMO indeed a bit strange. I think there should be a warning in the documentation saying that a model with nn.Reshape() cannot correctly detect 'single element batches' as batches if `batchMode` is not explicitely set.

PR #458, opened 2015-11-04
asserts in MultiCriterion and ParallelCriterion add

szagoruyko:
It is not possible to do ```nn.Sequential():add()```, but ```nn.ParallelCriterion():add()``` works without saying anything. Added the asserts.


soumith:
thanks.

PR #463, opened 2015-11-09
Add support for negative dimension and both batch and non batch inputs in nn.Sum

vgire:


soumith:
Thanks Vincent.

szagoruyko:
same has to be done for nn.Mean, nn.Max and nn.Min then.

vgire:
I agree and will do so when I have time.

vgire:
Done https://github.com/torch/nn/pull/464

PR #465, opened 2015-11-10
fix size for parallel container

misko:
I believe the inherited :size() function was overloaded with the size variable used inside one of the functions. I was not able to use :size() on the Parallel container which is not according to the documentation for the container class.

### DIFF
--- a/Parallel.lua
+++ b/Parallel.lua
@@ -10,7 +10,8 @@ end
 function Parallel:updateOutput(input)
    local nModule=input:size(self.inputDimension)
    local outputs = {}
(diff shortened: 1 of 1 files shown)


soumith:
hey @misko . Can you change: 
local totalOutputSize = torch.LongStorage() 

to
self.totalOutputSize = self.totalOutputSize or torch.LongStorage() 
local totalOutputSize = self.totalOutputSize

Because otherwise, there's a small malloc / free happening in every forward call.

soumith:
Thank you

PR #467, opened 2015-11-10
allows to get weight/bias separately

eladhoffer:


soumith:
hey Elad,

There was a PR like this in the past, and it was rejected by @andresy because it's getting very very specific to field names. One can do something like this simply via the generic :apply call.
Considering the history of something similar, closing the PR.

PR #468, opened 2015-11-10
Win: omp parallel for waits for signed parameter

diz-vara:
On Windows (Visual Studio) using 
  size_t nip 
results in
  error C3016: 'nip' : index variable in OpenMP 'for' statement must have signed integral type

Changed to LONG_PTR nip (LONG_PTR defined in windows.h)


soumith:
thanks as always @diz-vara

PR #473, opened 2015-11-16
State variable pointers should never change.

insperatum:
see: https://github.com/torch/nn/blob/master/doc/module.md#state-variables

### DIFF
--- a/Identity.lua
+++ b/Identity.lua
@@ -1,12 +1,12 @@
 local Identity, _ = torch.class('nn.Identity', 'nn.Module')
 
 function Identity:updateOutput(input)
-   self.output = input
+   self.output:set(input:storage(), input:storageOffset(), input:size(), input:stride())
    return self.output
 end
 
 
 function Identity:updateGradInput(input, gradOutput)
-   self.gradInput = gradOutput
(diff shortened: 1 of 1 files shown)


soumith:
Thanks a lot. There's a few more modules where this bad habit has propagated.

koraykv:
What if input to nn.Identity is not a tensor?

soumith:
oh, I didn't know that we are using it in such contexts. Let me revert the PR, sorry about that.

insperatum:
Aah whoops, sorry guys! My first commit and I accidentally run the test suite on the unmodified code :/
Will be more careful next time.

What do you think is the correct behaviour then, in modules which accept both tables and tensors? Hold onto one pointer for each type, or just keep it as it is and forget about preserving the pointer?

koraykv:
I would suggest checking if the input is a tensor and then applying your version. For anything else, I think it is too involved. It might be arbitrary data passed around with hierarchies of tables...

Identity.lua
```diff
@@ -1,12 +1,12 @@
 local Identity, _ = torch.class('nn.Identity', 'nn.Module')
 
 function Identity:updateOutput(input)
-   self.output = input
+   self.output:set(input:storage(), input:storageOffset(), input:size(), input:stride())
```
dominikgrewe: `self.output:set(input)` has the same effect.

PR #475, opened 2015-11-16
Do not change state variables in CosineDistance/CosineEmbeddingCriterion

fmassa:
Previous version was changing the state variables if not in batch mode, similar issue as the one pointed out in https://github.com/torch/nn/pull/432.
A few notes:
* I'm resizing the state variables `output`, `gradInput`. If you prefer, I can create a temporary buffer instead
* I also cleaned a bit the use of the `buffer` variable. Replacing it by a zero-strided tensor was not a good idea (and lead to the bug mentioned in the bottom of the files), so I just expand it when needed.

### DIFF
--- a/CosineDistance.lua
+++ b/CosineDistance.lua
@@ -55,7 +55,7 @@ function CosineDistance:updateOutput(input)
    self.w:sqrt()
 
    self.output:cmul(self.w1,self.w)
-   self.output = self.output:select(2,1)
+   self.output:resize(input1:size(1))
 
    return self.output
 end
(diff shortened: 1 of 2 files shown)


soumith:
Thanks Francisco.

dominikgrewe:
`resize` can be dangerous, because it doesn't give any guarantees on the data. It only happens to work if the data is contiguous. At the very least, we should assert that that's the case. But using a temporary view would be cleaner.

fmassa:
@dominikgrewe Indeed, if the data is not contiguous `resize` won't work as expected.
This is not the case in here though, as both `output` and `gradInput` are contiguous tensors at creation, and thus stays contiguous in the end of the functions because there are no operations affecting the contiguity of those tensors inside the functions.
But I agree with you that it's easy to shoot yourself in the foot by using this approach, and it might be better to avoid it.
I'll prepare a patch for it tomorrow to use a temporary view instead.

PR #476, opened 2015-11-16
Normalize supports infinity (maximum) norm

fmassa:
This PR allows normalization by the maximum absolute-value of each batch element.
I thought it would be better to include this functionality into `Normalize`, instead of creating a new function only for that.
Infinity norm can be used by calling the constructor with `math.huge`, as follows
```lua
normalize = nn.Normalize(math.huge)
```

### DIFF
--- a/Normalize.lua
+++ b/Normalize.lua
@@ -17,18 +17,29 @@ function Normalize:updateOutput(input)
 
   self._output = self._output or input.new()
   self.norm = self.norm or input.new()
-  self.normp = self.normp or input.new()
   self.buffer = self.buffer or input.new()
 
   self._output:resizeAs(input)
 
-  if self.p % 2 ~= 0 then
-    self.buffer:abs(input):pow(self.p)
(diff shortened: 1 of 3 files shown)


Normalize.lua
```diff
@@ -103,3 +128,18 @@ function Normalize:__tostring__()
   end
   return string.format(s,torch.type(self),self.p)
 end
+
+function Normalize:type(type)
```
koraykv: should implement the tensorCache argument (https://github.com/torch/nn/blob/master/Module.lua#L116)

fmassa: I thought about it, but as this module doesn't have learnable parameters I thought it wasn't strictly necessary. I'll fix it (and I'll also fix the same issue with nn.Max and nn.Min)

fmassa: Updated it.

koraykv:
I am fine with this and related two other requests, but leaving it to @soumith  to have final look.

PR #480, opened 2015-11-17
Add sizeAverage to init()

sguada:


soumith:
Thanks Sergio. If you squash the commits, I'll merge it in. Also, if you dont mind, change the constructors in this markdown document: https://github.com/torch/nn/blob/master/doc/criterion.md as it will give correct in-line help etc.

soumith:
committed via https://github.com/torch/nn/commit/41d2b1d1ee8aa12b5195fbd3bd165bc932e0764f
your username / name is correctly attributed.

PR #481, opened 2015-11-17
Added Padding option in Volumetric convolution

kmul00:
**VolumetricConvolution** didn't support ***padding*** before. With this PR ***padding*** can be provided in **VolumetricConvolution** in a way similar to **SpatialConvolution**. That is,

    module = nn.VolumetricConvolution(nInputPlane, nOutputPlane, kT, kW, kH, [dT], [dW], [dH], [padT], [padW], [padH])

By default `padT=padW=padH=0`

To be able to add ***padding*** support I had to write a new module to perform Volumetric Convolution, `VolumetricConvolutionMM.c` which, similar to `SpatialConvolutionMM`, constructs Toeplitz matrix and does matrix multiplications. This is done to keep the current implementation of **VolumetricConvolution** intact (`VolumetricConvolution.c`) and it was difficult to add this functionality in it.

I have added necessary test cases in `test.lua`  and initializations in `init.c` as well.

I will be adding similar functionality to the `cuda` version in `cunn`.

### DIFF
--- a/VolumetricConvolution.lua
+++ b/VolumetricConvolution.lua
@@ -1,6 +1,6 @@
 local VolumetricConvolution, parent = torch.class('nn.VolumetricConvolution', 'nn.Module')
 
-function VolumetricConvolution:__init(nInputPlane, nOutputPlane, kT, kW, kH, dT, dW, dH)
+function VolumetricConvolution:__init(nInputPlane, nOutputPlane, kT, kW, kH, dT, dW, dH, padT, padW, padH)
    parent.__init(self)
 
    dT = dT or 1
@@ -15,6 +15,9 @@ function VolumetricConvolution:__init(nInputPlane, nOutputPlane, kT, kW, kH, dT,
    self.dT = dT
    self.dW = dW
    self.dH = dH
+   self.padT = padT or 0
+   self.padW = padW or self.padT
+   self.padH = padH or self.padW
 
    self.weight = torch.Tensor(nOutputPlane, nInputPlane, kT, kH, kW)
    self.bias = torch.Tensor(nOutputPlane)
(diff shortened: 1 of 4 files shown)


fmassa:
Cool, nice work !
I'm just wondering, this toeplitz implementation is quite memory consuming, right (because of the unfolded tensor) ?
Also, did you considered basing the changes according to VolumetricDeconvolution ? It supports padding and does not need the temporary buffer. I don't know how it performs though compared to the unfolded+MM approach.

kmul00:
Thanks !

Yeah, it is memory consuming, since it has to store the unfolded tensor in a temporary matrix.
And I actually based my changes accordinn to the `SpatialConvolutionMM`, since it does similar thing for Spatial domain. I didn't consider `VolumetricDeconvolution` that much. Let me know if that's a problem.

kmul00:
I have a 6 layer network, with filters of size `3X5X5` in the first two layers, followed by `3X3X3` in the remaining layers. Input diemension is `2X38X64X64`. Epoch size is ~ `3500`.
On CPU, with the previous implementation - `Memory ~ 1 GB, time/epoch ~ 2 hrs`
With this implementtaion - `Memory ~ 4.5 GB, time/epoch ~ 20 mins`.

kmul00:
Hi,

Any suggestion/ comments regarding this ?

fmassa:
I don't use `VolumetricConvolution` at the moment, but I think that the memory requirements are a bit too high. It's already the case for `SpatialConvolutionMM`, but it's necessary to have a parallel implementation wrt the batch size.
Also, as you have removed the OMP parallel in the code, there is no need to have `finput` and `fgradInput` to depend on the batch size I think.

kmul00:
@fmassa, Thanks for pointing it out.

I removed the OMP parallel code because of [this issue] (https://github.com/torch/nn/commit/b071b6afcc4d67f9b0b32c075dfa1bab601ce17a) . 
I agree the memory requirement is on the higher side. My assumption was that since people would be using the `cuda` implementation for large data anyways and the speed gain was massive (would be even more if Open MP is used) as shown above, so it's a kind of trade-off between the speed and memory.

Any suggestion regarding whether I should sway away from doing matrix multiplication and try to add padding capabilities in the existing code (which would be difficult since it calls `conv3dmv` internally to perform convolution) or try to reduce memory consumption in this MM implementation only ?

soumith:
Koustav, it's interesting that you implemented it in terms of Volumetric version.
Looks good so far, squash your commits and I'll merge it in.

Thanks a lot.

soumith:
scratch my comment about "implemented it in terms of..."
Looks good to merge.

fmassa:
I think that adapting [VolumetricDeconvolution](https://github.com/torch/nn/blob/master/VolumetricDeconvolution.lua) to have a `VolumetricConvolution` is not that difficult (it involves switching the forward with the backward, compare for example https://github.com/torch/nn/pull/405 and https://github.com/torch/cunn/blob/master/SpatialConvolutionMM.cu). But that's only my point of view.
If the memory requirements are a problem in the future, we can change it later.
Thanks for PR !

fmassa:
I think it looks good to merge as well

kmul00:
Thank you for your feedback. I can look into adapting `VolumetricDeconvolution` if you want. I will send the PR for the `cuda` version of this soon, I already have it with me.

Also, I would like to know what exactly *"squash your commits"* means. :P

fmassa:
have a look at http://stackoverflow.com/questions/2563632/how-can-i-merge-two-commits-into-one

In your case, something like
```
git rebase --interactive HEAD~4
```
and then repling the pick with squash, as in the link

kmul00:
Thanks a lot.

fmassa:
No need to spend time now adapting `VolumetricDeconvolution`, it's certainly going to be slower than this version (look at https://github.com/torch/nn/blob/master/generic/VolumetricDeconvolution.c#L69-L90)

soumith:
@kmul00 afterwards, you have to do git push --force

kmul00:
@soumith , I squashed all my commits and did a `git push --force`.

@fmassa , Yeah, from the piece of code that you shared, it's clear that it is going to be slower. I'll skip adopting `VolumetricDeconvolution` then for the time being.

soumith:
awesome! thanks a lot.

PR #488, opened 2015-11-20
Retain nn.Identity state pointers if they are tensors

insperatum:
Following through on https://github.com/torch/nn/pull/473

### DIFF
--- a/Identity.lua
+++ b/Identity.lua
@@ -1,6 +1,7 @@
-local Identity, _ = torch.class('nn.Identity', 'nn.Module')
+local Identity, parent = torch.class('nn.Identity', 'nn.Module')
 
 function Identity:__init()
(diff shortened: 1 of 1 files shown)


soumith:
This looks good. Can you squash your commits?

soumith:
committed via https://github.com/torch/nn/commit/b7a64bf1f8062b394ee28679f4d6bf736b2672c6 your name and username are correctly attributed. Thanks a lot.

karpathy:
This seems like a big backwards incompatible change and seems to invalidate all previously created checkpoints in Torch? E.g. if I had a saved checkpoint that has an old version of `nn.Identity` then users of this code will not be able to run them because the field `tensorOutput` does not exist.

PR #489, opened 2015-11-21
Allow ceil_mode in SpatialAveragePooling

jhjin:


jhjin:
duplicated PR
https://github.com/torch/nn/pull/365

PR #490, opened 2015-11-23
nn.View now allows non-contiguous tensors for forward/backward pass

jacobmenick:
The nn.View module currently makes calls to input:view() or gradOutput:view() in the forward and backward passes. This results in an error if the tensors passed as input (or gradOutput) are not contiguous. In this pull request, nn.View makes a contiguous copy of the tensors passed in if they are not contiguous.

### DIFF
--- a/doc/table.md
+++ b/doc/table.md
@@ -736,7 +736,7 @@ starting at index `offset` having `length` elements (defaults to 1 element).
 The elements can be either a `table` or a [`Tensor`](https://github.com/torch/torch7/blob/master/doc/tensor.md#tensor).
 
 The gradients of the elements not included in the subtable are zeroed `Tensor`s of the same size. 
-This is true regardless of the dept of the encapsulated `Tensor` as the function used internally to do so is recursive.
+This is true regardless of the depth of the encapsulated `Tensor` as the function used internally to do so is recursive.
 
 Example:
 ```lua
@@ -844,7 +844,7 @@ prl = nn.ParallelTable()
 prl:add(p1_mlp)
 prl:add(p2_mlp)
 
--- now we define our top level network that takes this parallel table and computes the pairwise distance betweem
+-- now we define our top level network that takes this parallel table and computes the pairwise distance between
 -- the pair of outputs
 mlp= nn.Sequential()
 mlp:add(prl)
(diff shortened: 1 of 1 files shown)


fmassa:
https://github.com/torch/nn/pull/284

View.lua
```diff
@@ -72,16 +74,26 @@ local function batchsize(input, size, numInputDims, numElements)
...
 function View:updateOutput(input)
-   local bsz = batchsize(input, self.size, self.numInputDims, self.numElements)
-   if bsz then
-      self.output = input:view(bsz, table.unpack(self.size:totable()))
-   else
-      self.output = input:view(self.size)
-   end
-   return self.output
+  if not input:isContiguous() then
+    self._input:resizeAs(input):copy(input)
```
fmassa: Saved models doesn't have this `_input` field, so you should lazily initialize it.

test.lua
```diff
@@ -509,7 +509,7 @@ function nntest.Normalize()
    end
 
    -- batch mode
-   for _,p in pairs({1,2,3,4,torch.uniform()*math.random(1,10),math.huge}) do
+   for _,p in pairs({1,2,3,4,torch.uniform()*math.random(1,10)}) do
```
fmassa: This shouldn't be removed

test.lua
```diff
@@ -3783,6 +3783,17 @@ function nntest.View()
       "Error in minibatch generalization nElement")
 end
 
+function nntest.ViewNonContiguous()
+  -- Non-continous input and gradOutput
+  local nonContigInput = torch.randn(2, 2, 2):select(2, 1)
+  local nonContigGradOut = torch.randn(4, 2):select(2, 2)
+  local viewModule = nn.View(4)
+  mytester:assert(nonContigInput:isContiguous() == false)
```
fmassa: I think those asserts should be changed.
What I think you could do is to create a `contigInput`, which is a clone of `nonContigInput`, run both in the View module, and assert that the output tensors are equal.
Same thing for gradInput.

fmassa:
Hi,
I've made some inline comments.
I'm ok with the PR (after the changes I mentioned) if @soumith agrees.

jacobmenick:
Hello,

OK, good. Regarding the removal in test.lua - my mistake. I was out of date with upstream.

PR #493, opened 2015-11-25
allow non-contigous inputs to nn.View.

jacobmenick:
Lazy initialize a contiguous copy of input or gradOutput as necessary.

### DIFF
--- a/View.lua
+++ b/View.lua
@@ -72,16 +72,28 @@ local function batchsize(input, size, numInputDims, numElements)
 end
 
 function View:updateOutput(input)
-   local bsz = batchsize(input, self.size, self.numInputDims, self.numElements)
-   if bsz then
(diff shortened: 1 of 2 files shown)


soumith:
will look at PRs in two days time. Sorry for the delay and non-response.

soumith:
@jacobmenick nn.Reshape does what you want, i.e. it does a memcpy on non-contiguous. 
We are trying to make nn.* operators as close as possible to their torch.* operators, so this is not a good idea. If you want me to add any features to nn.Reshape that are missing, let me know.
As @fmassa pointed out, this PR was rejected in the past. Since there's a second PR that came in, I did discuss it briefly with @andresy and he held the same view, that nn.* should be kept as close to torch.* as possible.

PR #497, opened 2015-11-26
Update VolumetricConvolution.lua

kmul00:
Update `VolumetricConvolution` to support `cuda`

Addresses [this](https://github.com/torch/cunn/pull/141#issuecomment-159900066)


soumith:
Thanks

PR #500, opened 2015-11-27
Added ExponetialLU.

bobbens:
Simple non-inplace pure Lua implementation.


soumith:
added via #508 which had docs and tests (and cuda version).

Issue #415, opened 2015-10-04
Weight sharing not working for some modules after converting the model with :cuda() method


ivankreso:
Defined weight sharing breaks after calling the cuda() method for some modules.
I guess that this is not an expected behavior (for example with nn.Linear it works)?

Example code for reproduction:
```
require 'cunn'
nn1 = nn.Sequential()
nn1:add(nn.SpatialConvolution(1, 1, 4, 4))
nn2 = nn1:clone('weight','bias','gradWeight','gradBias')
model = nn.ParallelTable()
model:add(nn1)
model:add(nn2)
w,g = model:getParameters()
print(w:size())
print(g:size())
model:cuda()
w,g = model:getParameters()
print(w:size())
print(g:size())
```

```
th> net=dofile('test_bug.lua')

 17
[torch.LongStorage of size 1]


 17
[torch.LongStorage of size 1]


 34
[torch.LongStorage of size 1]


 34
[torch.LongStorage of size 1]
```

PR #509, opened 2015-11-30
Cosine similarity

nicholas-leonard:
This PR adds a Euclidean-like module for measuring the cosine similarity between an input `x` and `k` cluster centers. Includes unit test and doc.

### DIFF
--- a/Cosine.lua
+++ b/Cosine.lua
@@ -1,0 +1,163 @@
+local Cosine, parent = torch.class('nn.Cosine', 'nn.Module')
+
+function Cosine:__init(inputSize,outputSize)
+   parent.__init(self)
+
+   self.weight = torch.Tensor(outputSize,inputSize)
+   self.gradWeight = torch.Tensor(outputSize,inputSize)
+   
+   self:reset()
+end
+
+function Cosine:reset(stdv)
+   if stdv then
+      stdv = stdv * math.sqrt(3)
+   else
+      stdv = 1./math.sqrt(self.weight:size(1))
+   end
+   self.weight:uniform(-stdv, stdv)
+end
+
+function Cosine:updateOutput(input)
+   local inputSize = self.weight:size(2)
+   local outputSize = self.weight:size(1)
+   
+   self._weightNorm = self._weightNorm or self.weight.new()
+   self._inputNorm = self._inputNorm or self.weight.new()
+   
+   -- y_j = (w_j * x) / ( || w_j || * || x || )
+   
+   self._weightNorm:norm(self.weight,2,2):add(1e-12)
+   if input:dim() == 1 then
+      self.output:resize(outputSize):zero()
+      self.output:addmv(1, self.weight, input)
+      self.__norm = input:norm()+1e-12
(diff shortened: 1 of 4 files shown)


Cosine.lua
```diff
@@ -0,0 +1,163 @@
...
+   self._weightNorm = self._weightNorm or self.weight.new()
+   self._inputNorm = self._inputNorm or self.weight.new()
+   
+   -- y_j = (w_j * x) / ( || w_j || * || x || )
+   
+   self._weightNorm:norm(self.weight,2,2):add(1e-12)
+   if input:dim() == 1 then
+      self.output:resize(outputSize):zero()
+      self.output:addmv(1, self.weight, input)
+      self.__norm = input:norm()+1e-12
```
soumith: this is blocking, but I guess we cant do much here. good that the mini-batch case doesn't suffer from this issue.

soumith:
Thanks Nick.

Cosine.lua
nicholas-leonard: Yeah. Main use case should be mini-batch.

fmassa: Is there a speed penalty in this case to reshape the input to be batched and use the standard batched version ? This would remove the blocking issue and also the amount of code redundancy.

nicholas-leonard: @fmassa Good question. Personally, I like that the 1D case is different than 2D, as it makes for a better unit test. In anycase, if you want your 1D use case to be fast, you could just make the 1D input into a 2D Tensor *before* it enters this module. But yeah, in any case, the code is redundant. If you want to make the change, feel free, but please transfer the 1D code into the unit test though (without all the buffering of course).

Cosine.lua
```diff
@@ -0,0 +1,163 @@
...
+      self._weight:addmm(-1, self._weight, 1, gradOutput:t(), input_)
+      
+      self._weight:cdiv(self._weightNorm:expandAs(self._weight))
+      self.gradWeight:add(self._weight)
+   else
+      error"1D or 2D input expected"
+   end
+end
+
+function Cosine:type(type)
```
fmassa: Just noticed that this module needs the `tensorCache` parameter to properly share it's weights, as per https://github.com/torch/nn/issues/415

fmassa: I'll prepare a fix

nicholas-leonard: Yes, thank you @fmassa !

Cosine.lua
fmassa: @nicholas-leonard @soumith  I don't have anything against current implementation (except that it's blocking), but I think ultimately we will support only batched operations, right (according to the roadmap) ? Anyway, I'd vote for leaving it as is for the moment.

nicholas-leonard: @fmassa Yeah. Initially, I wanted to do it as you said. But then I remembered how much easier it was for me to start with the 1D case, and then move to the 2D case. I never use 1D really, it was just to make the 2D implementation easier for me to understand, and also happens to make for a nice triple comparison in the unit test :)

PR #510, opened 2015-12-01
refactoring Deconvolution -> FullConvolution

soumith:
Refactored both Spatial and Volumetric versions.
All tests pass.


test.lua
```diff
@@ -1864,7 +1864,7 @@ function nntest.SpatialConvolutionMM()
    mytester:asserteq(0, (gradInput-gradInputc):abs():max(), torch.typename(module) .. ' - contiguous err ')
 end
 
-function nntest.SpatialDeconvolution()
+function nntest.SpatialFullConvolution()
```
fmassa: There is a clash with the previous `SpatialFullConvolution` test, no ?

soumith: will fix. good catch.

PR #512, opened 2015-12-01
Fix type() in Cosine

fmassa:
Fix for problem mentioned in https://github.com/torch/nn/pull/509/files#r46301793


soumith:
thanks

PR #514, opened 2015-12-02
Using sparse implementation of zeroGradParameters for SparseLinear

zhangxiangxiao:
This is similar to #362, implemented but not used. This is necessary to implement sparse updates correctly. Let me know if there is a reason why this was not included.

### DIFF
--- a/SparseLinear.lua
+++ b/SparseLinear.lua
@@ -62,5 +62,17 @@ function SparseLinear:updateGradInput(input, gradOutput)
 end
 
 function SparseLinear:updateParameters(learningRate)
-   self.lastInput.nn.SparseLinear_updateParameters(self, learningRate)
+   if self.lastInput then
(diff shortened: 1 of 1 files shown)


soumith:
Your tests are failing, as a first check. Haven't reviewed it yet.

zhangxiangxiao:
The error is because SparseJacobian uses SparseLinear:zeroGradParameters() before a forward propagation, therefore lastInput is missing.

One simple solution might be to add a guard to test whether lastInput is nil, and if so it calls Module's zeroGradParameters.

On the other hand, SparseJacobian seems to be used only for testing SparseLinear since that's the only sparse module so far. So we could also change SparseJacobian.

Let me know which solution you prefer and I will do it.

zhangxiangxiao:
Tests passed; squashing done :)

soumith:
lgtm :)

PR #518, opened 2015-12-08
Fast SpatialBatchNormalization (Facebook)

szagoruyko:
CPU part of https://github.com/torch/cunn/pull/143


soumith:
fixed via #550

PR #522, opened 2015-12-11
PairwiseDistance sanitization-friendly

mys007:
Output and gradInput should be initialized on use and not in constructor (as common in other modules), this allows proper sanitization.


soumith:
Thanks Martin.

Issue #506, opened 2015-11-27
THNN


andreaskoepf:
@fmassa @apaszke and I started working on a FFI based functional lua-free C-backend for torch/nn. The development takes place in [this fork](https://github.com/fmassa/nn/tree/THNN) in the THNN branch, the THNN lib resides in lib/THNN. Up to now 28 of the 44 C modules in nn have been converted (CPU only). If somebody wants to help, please feel free to contact us via [gitter](https://gitter.im/apaszke/THNN), some [style-guidelines and description of the conversion steps](https://github.com/fmassa/nn/blob/THNN/lib/THNN/README.md) exist.

We came to the conclusion that creating THNN as one big PR for nn/cunn might not be the best solution since it takes some time (maybe until X-mas) to convert all CPU/GPU modules and we want to avoid further project drift to the main nn/cunn repo since all changes to nn/cunn had to be tracked and integrated into THNN as well. Technically it is be possible to do the conversion to FFI at the granularity of one module.

We would like to discuss the following points with the torch owners (@soumith et al.):

1. Do you think it would be acceptable to have the C-backend for torch split into two libraries during a short period of time (1-2 months)?
2. From a testing perspective it would be nice to run cross-checks against the old nn code. But for contributors who want to send in fixes having the old impls flying around could be confusing. Also to keep the git-history of the C impls it would be best to use git-mv from nn/generic to nn/lib/THNN/generic. If we remove the original impl we would have to find a way to run the old C-backend vs. THNN functional API tests externally.
3. We have planned to use a vtable-lookup by a tensor:type() string, creating the target signature dynamically once and caching the function returned by ffi. The 'old' nn impl uses the metatable of the tensors to store the nn-function-table. Should we keep the old metatable-misuse or go for the tensor-type-vtable lookup-by string?
4. Some modules need access to a random generator. Generating random numbers differs for cpu/cuda (e.g. THRandom_uniform vs. curand_uniform). Is it important to be able to pass in a custom random generator (e.g. control rgn-state/seed on a call-by-call basis etc.) or could backend-specific global rngs be part of the THNNState* which is now passed as first arg to each backend-function?

Also there are some more general questions:
A reason for creating THNN is to be able to reference the torch functions from a C/C++ library that would mirror the lua side, e.g. to allow use of torch models in production code without lua dependency (e.g. no need to break out of the torch universe and  use Caffe). 
Which models are the most important? 
[...]

PR #524, opened 2015-12-12
THNN: Functional C backend conversion using ffi

andreaskoepf:
This PR is created to gather feedback from the torch community for some design aspects of the THNN library.

The goal of THNN is a [FFI](http://luajit.org/ext_ffi.html) based functional lua-free C-backend for torch/nn, see #506. This project was started some weeks ago by @apaszke, @fmassa and me. We already prepared the conversion of about half of the modules of nn, see [this branch](https://github.com/fmassa/nn/tree/THNN/lib). In the end together with an additional THNN++ lib it should be possible to evaluate models in pure C without lua dependency (e.g. in production code).

In this PR the simple modules `Abs` and `AbsCriterion` have been moved into a separate libTHNN. The module-functions no longer have a single lua_State* parameter but a full signature, e.g. `void THNN_(Abs_updateOutput)(THNNState *state, THTensor *input, THTensor *output);`.

Feedback on the following points is requested:

1.  The current plan is to use FFI. We had some discussions with @soumith regarding ffi vs. cwrap. A critical point of using ffi is the unclear future of luajit, [see here](http://www.freelists.org/post/luajit/Looking-for-new-LuaJIT-maintainers). But there is a compatible FFI [luaffifb](https://github.com/facebook/luaffifb) provided by facebook. In general we are not sure about the opinion of the torch users regarding ffi: If FFI is a showstopper for you please let us know why.
2. Our current plan is to have separate THNN/THCUNN libs. Should we instead perform the 'functional conversion' inplace within libnn? E.g. the advantage of doing things inplace might be that the generic/*.c files stay where they are. For people not following the THNN conversion this might be a less drastic change... also we do not need CMakeLists for a new library. If on the other hand there will always remain some c part interacting with Lua state it would be better to separate the pure c implementation from the Lua c backend.
3. How should we perform the type dependent THNN calls? Two alternative ways to call the tensor-type dependent functions of THNN are demonstrated in this PR: Abs uses `THNN.runKernel('Abs_updateOutput', ... )` while AbsCriterion uses a tensor-metatable based approach which is very similar to the current impl (`input.THNN.AbsCriterion_updateOutput(..)`).
In general the 'metatable-trick' seems to be a little bit faster as shown in this little [benchmark](https://gist.github.com/andreaskoepf/8c646142e5414427e282).


andreaskoepf:
@fmassa has created an alternative proposal that requires no changes of the lua files at all by extracting the main code parts from the generic/*.c files and moving them into a separate THNN lib, see https://github.com/fmassa/nn/commit/dd6bf1437ba17c3b42afa2823238227049861d9f.

Advantages of his approach:

- no FFI
- the conversion of the CPU impl in [nn](https://github.com/torch/nn) could be done independent of the cuda version in [cunn](https://github.com/torch/nn).

Disadvantage:

- Basically it would mean to convert the existing C files into hand-written cwarp code. Modules authors would have to write this glue-code in addition to their main module-implementation.

Any feedback from the prj owners would be appreciated (e.g. whether to use FFI or not).

hughperkins:
Hi, what is current recommended method of using torch from c/c++?  I suppose I just need the basics, like:
- spaticalconvolutionmm
- relu
- tanh
- linear
- maxpooling
- (ideally) averagepooling
- logsoftmax
- nllclasscriterion

andreaskoepf:
@hughperkins The 'problem' of the current torch/nn C-backend is that all args are passed via lua_State* and lua-state-accesor functions. The main motivation behind THNN is to allow 're-use' of the torch code in C/C++ apps by linking to the THNN(++) lib.

hughperkins:
Yes, going via lua stack doesnt sound terribly convenient, though could wrap this a bit I suppose.  I'm a bit puzzlling over how facebook guys avoid ever wanting to use torch from c/c++ actually?  I can see that 80-90% of use-cases might not need c/c++, but 100% of use cases seems ... surprising?

Abs.lua
```diff
@@ -1,3 +1,5 @@
+local THNN = require('nn.THNN')
```
fmassa: wouldn't it be better to include `THNN` in `nn` namespace, like `nn.THNN` or `nn._THNN` ? This would avoid having to require 'nn.THNN' at every lua file.

andreaskoepf: @fmassa If the tensor metatable-'trick' is used there is no need for `require THNN` in each file (only once in init.lua).

In general I wanted to avoid further global namespace pollution - I think torch has there been a litte bit sloppy in the past.

soumith:
This plan looks good. Please go ahead with this. 
Let's break the future work into a few modules per PR, that I can merge in.

cc: @dominikgrewe @andresy @koraykv @clementfarabet

THNN.lua
```diff
@@ -0,0 +1,133 @@
...
+function THNN.getState()
+  return ffi.NULL or nil
+end
+
+function THNN.optionalTensor(t)
+  return t and t:cdata() or THNN.NULL
+end
+
+local ok,result
+if ffi.os == "OSX" then
```
soumith: this whole section around ffi.load should be:
(1) made this way: https://github.com/torch/tds/blob/735f6cd37e7e3385c8eb229d1c38969f07436542/env.lua#L5
     So, THNN will be installed into cpath, and we know exactly where it will be in all OSes.
(2) No pcall around it. It should be a hard failure.

andreaskoepf:
Latest changes: 

- Abs() & AbsCriterion() both are now using the tensor 'metatable' trick
- CMakeLists.txt for THNN allows stand-alone build (e.g. does an optional FIND_PACKAGE(Torch))
- building libTHNN as library type 'MODULE' to create an .so file on OSX
- added package.searchpath polyfill for LUA 5.1

@soumith I hope this resembles what you had in mind.

soumith:
Rebased and merged via a series of commits ending with: https://github.com/torch/nn/commit/baf9255c6cf6ea6fb5b5b804d264d49f16d5ce81

soumith:
Andreas, this is fantastic. Thanks a lot.

hughperkins:
Well... speaking for myself, THNN had the effect, in order, of:
- breaking opencl regularly, and unexpectedly, often
- therefore I ultimately forked distro to distro-cl, which at least gives me a peaceful life
- but on the other hand, I have zero interaction with torch community now
- (and obviously distro-cl becomes increasingly out of date)

Presumably THNN is stable now?  At some point, I guess I should re-port everything across.

hughperkins:
(PS whats the status of hcctorch, with which I have zero interaction with.  is it replacing cltorch?  sitting in the background under development?  a bit of both?)

yuntang:
Is " THNN++" part of the plan?

soumith:
@yuntang not in the near future

PR #525, opened 2015-12-13
fix a bug in conditional expression

xwgeng:
the conditional expression is always false


soumith:
could you squash your commits into 1 commit

xwgeng:
ok

Issue #503, opened 2015-11-27
back compat serialization


soumith:
add tests for back-compat serialization for all nn modules.

andreaskoepf:
Could we introduce a new (optional) method that gets called after deserialization of an object to check and potentially correct the object... For module authors it might not be obvious that there are ways to (re-)create the objects which do not run the ctor.

soumith:
We already have custom read / write that can be specified. I did it this
way in a hurry. More proper fix and unit tests will come over the weekend.

andresy:
If it can be of any help: do you know that :read(file, [version]) takes an optional argument version number (which reflects __version in the metatable), such that you can handle new serialisation variants?

andreaskoepf:
@soumith & @andresy thnx for mentioning read which I did not know.

I created a [mini demo gist](https://gist.github.com/andreaskoepf/8a2df7daea695d98d062#file-torch_obj_versioning-lua) which shows a `restore` function that is called from a custom read impl after simply assigning all deserialized-members without changes.

andreaskoepf:
Maybe it would be possible to have a method similar to `__factory` but that is called during normal class construction and during deserialization. Or maybe we could add a flag in the class-metatable that would trigger execution of the ctor before deserialization. This would at least cover the very common case of additive changes: e.g. replace the need for `backCompatibility()` in  [LookupTable.lua](https://github.com/torch/nn/blob/master/LookupTable.lua).
In general I think it would be better to 'upgrade' an obj once after deserialization than having backwardCompatibility() calls everywhere.

andresy:
you could have a subinit() function called from your custom __factory() as well as the constructor.

andreaskoepf:
@andresy Yes, I was thinking of same kind of 'default ctor' to reduce code duplication in ctor and the custom-deserialization handler - but I forgot that many initializations depend on parameters or state: E.g. a common case would be to add a new tensor field - but it could not be created in a default ctor because its type has to be the same as the other tensors after deserialization (e.g. if the module was  `:cuda()`-converted before serialization): So after sleeping over it I am no longer convinced a default-ctor `subinit()` would be of much help. The `restore()` method solution shown in my gist might be a candidate: e.g. we could add the custom `read` method as `nn.Module:read()` which would call an optional `restore()` method if it was present...

andreaskoepf:
[...]

PR #526, opened 2015-12-13
clearState

szagoruyko:
This introduces sanitization of nn modules so that
```lua
torch.save('model.t7', model:clearState())
```
will save a model without any internal buffers.
`nn.Module` has it's default clearState that clears `output` and `gradInput`.
Any module that uses buffers should have it's own clearState, I went over nn modules and fixed the ones I knew, it is possible that I missed something.

Test added into `Jacobian.testIO` that calls clearState before serializing the module. There might be a better solution because it does not test the old behaviour.

I will add `clearState` to `cudnn`, `inn` and nn.DataParallelTable with `cunn` tests later.

Comments? the original name was `sanitize`, we should discuss if the new name is better.

Was implemented when I was at facebook, it's been in use for a couple of months, pretty stable.

### DIFF
--- a/BatchNormalization.lua
+++ b/BatchNormalization.lua
@@ -154,3 +154,14 @@ function BN:accGradParameters(input, gradOutput, scale)
       self.gradBias:add(scale, self.buffer)
    end
 end
+
+function BN:clearState()
+   nn.utils.clear(self, {
+      'buffer',
+      'buffer2',
+      'centered',
+      'std',
+      'normalized',
+   })
+   return parent.clearState(self)
+end
--- a/CosineDistance.lua
+++ b/CosineDistance.lua
@@ -73,6 +73,11 @@ function CosineDistance:updateGradInput(input, gradOutput)
       not_batch = true
    end
 
+   if #self.gradInput ~= 2 then
+      self.gradInput[1] = self.gradInput[1] or v1.new()
+      self.gradInput[2] = self.gradInput[2] or v1.new()
+   end
+
    local gw1 = self.gradInput[1]
    local gw2 = self.gradInput[2]
    gw1:resizeAs(v1):copy(v2)
@@ -96,4 +101,16 @@ function CosineDistance:updateGradInput(input, gradOutput)
    end
 
    return self.gradInput
+end
+
+function CosineDistance:clearState()
+   nn.utils.clear(self, {
+      'buffer',
+      'w1',
+      'w22',
+      'w',
+      'w32',
+      'ones',
+   })
+   return parent.clearState(self)
 end
--- a/DotProduct.lua
+++ b/DotProduct.lua
@@ -26,6 +26,11 @@ function DotProduct:updateGradInput(input, gradOutput)
    local v2 = input[2]
    local not_batch = false
 
+   if #self.gradInput ~= 2 then
+     self.gradInput[1] = self.gradInput[1] or input[1].new()
+     self.gradInput[2] = self.gradInput[2] or input[2].new()
+   end
+
    if v1:dim() == 1 then
       v1 = v1:view(1,-1)
       v2 = v2:view(1,-1)
@@ -48,4 +53,9 @@ function DotProduct:updateGradInput(input, gradOutput)
    end
 
    return self.gradInput
+end
+
+function DotProduct:clearState()
+   if self.buffer then self.buffer:set() end
+   return parent.clearState(self)
 end
--- a/Module.lua
+++ b/Module.lua
@@ -364,3 +364,7 @@ function Module:listModules()
    end
    return modules
 end
+
+function Module:clearState()
+   return nn.utils.clear(self, 'output', 'gradInput')
+end
--- a/Normalize.lua
+++ b/Normalize.lua
@@ -140,3 +140,16 @@ function Normalize:type(type, tensorCache)
   end
   return self
 end
+
+function Normalize:clearState()
+   nn.utils.clear(self, {
+      '_output',
+      '_indices',
+      '_gradInput',
+      'buffer',
+      'norm',
+      'normp',
+      'cross',
+   })
+   return parent.clearState(self)
+end
--- a/PReLU.lua
+++ b/PReLU.lua
@@ -45,3 +45,8 @@ function PReLU:accGradParameters(input, gradOutput, scale)
    )
    return self.gradWeight
 end
+
+function PReLU:clearState()
+   nn.utils.clear(self, 'gradWeightBuf', 'gradWeightBuf2')
+   return parent.clearState(self)
+end
--- a/SpatialBatchNormalization.lua
+++ b/SpatialBatchNormalization.lua
@@ -140,3 +140,14 @@ function BN:read(file, version)
       end
    end
 end
+
+function BN:clearState()
+   nn.utils.clear(self, {
+      'buffer',
+      'buffer2',
+      'centered',
+      'std',
+      'normalized',
+   })
+   return parent.clearState(self)
+end
--- a/SpatialConvolution.lua
+++ b/SpatialConvolution.lua
@@ -171,3 +171,9 @@ function SpatialConvolution:__tostring__()
    end
    return s .. ')'
 end
+
+function SpatialConvolution:clearState()
+   nn.utils.clear(self, 'finput', 'fgradInput', '_input', '_gradOutput')
+   return parent.clearState(self)
+end
+
--- a/SpatialMaxPooling.lua
+++ b/SpatialMaxPooling.lua
@@ -59,13 +59,9 @@ function SpatialMaxPooling:updateGradInput(input, gradOutput)
    return self.gradInput
 end
 
+-- for backward compat
 function SpatialMaxPooling:empty()
-   self.gradInput:resize()
-   self.gradInput:storage():resize(0)
-   self.output:resize()
-   self.output:storage():resize(0)
-   self.indices:resize()
-   self.indices:storage():resize(0)
+   self:clearState()
 end
 
 function SpatialMaxPooling:__tostring__()
@@ -77,4 +73,11 @@ function SpatialMaxPooling:__tostring__()
    s = s .. ')'
 
    return s
+end
+
+function SpatialMaxPooling:clearState()
+   if self.indices then
+      self.indices:set()
+   end
+   return parent.clearState(self)
 end
--- a/VolumetricConvolution.lua
+++ b/VolumetricConvolution.lua
@@ -25,7 +25,7 @@ function VolumetricConvolution:__init(nInputPlane, nOutputPlane, kT, kW, kH, dT,
    self.gradBias = torch.Tensor(nOutputPlane)
    -- temporary buffers for unfolding (CUDA)
    self.finput = torch.Tensor()
-   self.fgradInput = torch.Tensor()   
+   self.fgradInput = torch.Tensor()
    self:reset()
 end
 
@@ -174,4 +174,9 @@ function VolumetricConvolution:type(type, tensorCache)
    self.finput:set()
    self.fgradInput:set()
    return parent.type(self, type, tensorCache)
+end
+
+function VolumetricConvolution:clearState()
+   nn.utils.clear(self, 'finput', 'fgradInput', '_input', '_gradOutput')
+   return parent.clearState(self)
 end
--- a/utils.lua
+++ b/utils.lua
@@ -162,4 +162,27 @@ function nn.utils.contiguousView(output, input, ...)
   return output
 end
 
+-- go over specified fields and clear them. accepts
+-- nn.utils.clearState(self, {'_buffer', '_buffer2'}) and
+-- nn.utils.clearState(self, '_buffer', '_buffer2')
+function nn.utils.clear(self, ...)
+   local arg = {...}
+   if #arg > 0 and type(arg[1]) == 'table' then
+      arg = arg[1]
+   end
+   local function clear(f)
+      if self[f] then
+         if torch.isTensor(self[f]) then
+            self[f]:set()
+         elseif type(self[f]) == 'table' then
+            self[f] = {}
+         else
+            self[f] = nil
+         end
+      end
+   end
+   for i,v in ipairs(arg) do clear(v) end
+   return self
+end
+
 table.unpack = table.unpack or unpack
--- a/Bilinear.lua
+++ b/Bilinear.lua
@@ -140,3 +140,8 @@ function Bilinear:__tostring__()
          (self.bias == nil and ' without bias' or '')
       )
 end
+
+function Bilinear:clearState()
+   if self.buff then self.buff:set() end
+   return parent.clearState(self)
+end
--- a/CMul.lua
+++ b/CMul.lua
@@ -116,13 +116,20 @@ end
 
 function CMul:type(type, tensorCache)
    if type then
-      self._input = nil
-      self._output = nil
-      self._weight = nil
-      self._gradWeight = nil
-      self._expand = nil
-      self._repeat = nil
-      self._sum = nil
+      self:clearState()
    end
    return parent.type(self, type, tensorCache)
+end
+
+function CMul:clearState()
+   nn.utils.clear(self, {
+      '_input',
+      '_output',
+      '_weight',
+      '_gradWeight',
+      '_expand',
+      '_repeat',
+      '_sum',
+   })
+   return parent.clearState(self)
 end
--- a/CMulTable.lua
+++ b/CMulTable.lua
@@ -48,3 +48,8 @@ function CMulTable:updateGradInput(input, gradOutput)
 
    return self.gradInput
 end
+
+function CMulTable:clearState()
+   if self.tout then self.tout:set() end
+   return parent.clearState(self)
+end
--- a/Container.lua
+++ b/Container.lua
@@ -75,3 +75,26 @@ function Container:parameters()
     end
     return w,gw
 end
+
+function Container:clearState()
+   -- don't call set because it might reset referenced tensors
+   local function clear(f)
+      if self[f] then
+         if torch.isTensor(self[f]) then
+            self[f] = self[f].new()
+         elseif type(self[f]) == 'table' then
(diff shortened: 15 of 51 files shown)


Module.lua
```diff
@@ -348,3 +348,25 @@ function Module:listModules()
    end
    return modules
 end
+
+function Module:clearState()
+    if self.output then
+       if torch.isTensor(self.output) then
+          self.output = self.output.new()
```
dominikgrewe: Rather than creating a new tensor, isn't it better to do `self.output:set()` which resets the tensor? Then all references to it are kept intact.

Normalize.lua
```diff
@@ -143,3 +143,10 @@ function Normalize:type(type, tensorCache)
   end
   return self
 end
+
+function Normalize:clearState()
+  self._output = nil
```
fmassa: there are a few more buffers that need to be cleaned
* [`buffer`](https://github.com/szagoruyko/nn/blob/clearState/Normalize.lua#L20)
* [`norm`](https://github.com/szagoruyko/nn/blob/clearState/Normalize.lua#L19)
* [`normp`](https://github.com/szagoruyko/nn/blob/clearState/Normalize.lua#L34)
* [`cross`](https://github.com/szagoruyko/nn/blob/clearState/Normalize.lua#L62)

SpatialConvolution.lua
```diff
@@ -139,3 +139,16 @@ function SpatialConvolution:__tostring__()
    end
    return s .. ')'
 end
+
+function SpatialConvolution:clearState()
+   self.finput = self.output.new()
```
fmassa: on the same note as the one from @dominikgrewe , wouldn't it be better to do `self.finput:set()` in here ? One can for example share the internal buffers inside the model to reduce memory footprint, wouldn't it be desirable to be kept after `clearState` ?

nicholas-leonard:
:+1:

andreaskoepf:
@szagoruyko would it be possible to use `t:set()` instead of `t = nil` - at least for some of the tensors of `nn.Module`? See my last comment [here](https://github.com/torch/nn/issues/503#issuecomment-160288246). It would also allow to serialize tensor refs.

andreaskoepf:
(sorry, it was already mentioned by @dominikgrewe).

szagoruyko:
@fmassa I fixed Normalize.
@dominikgrewe @andreaskoepf calling `set()` on `output` and `gradInput` is tricky because it might cause referenced tensors in other places to be reseted, which is unexpected. For example `ConcatTable`, `Narrow` or `Identity` will not work as expected after clearState.
I have changed to use set in other places where it is possible.

SpatialMaxPooling.lua
```diff
@@ -33,6 +33,7 @@ function SpatialMaxPooling:updateOutput(input)
    self.ceil_mode = self.ceil_mode or false
    self.padW = self.padW or 0
    self.padH = self.padH or 0
+   self.indices = input.new()
```
soumith: this will always malloc a new self.indices on updateOutput. You want to do: self.indices = self.indices or input.new()

VolumetricConvolution.lua
```diff
@@ -80,6 +77,8 @@ local function unviewWeight(self)
 end
 
 function VolumetricConvolution:updateOutput(input)
+   self.finput = input.new()
```
soumith: same as above. you need to do a: a = a or b

SpatialMaxPooling.lua
szagoruyko: oops.

dominikgrewe:
> @dominikgrewe @andreaskoepf calling set() on output and gradInput is tricky because it might cause referenced tensors in other places to be reseted, which is unexpected. For example ConcatTable, Narrow or Identity will not work as expected after clearState.

That's because those modules take references to tensors that it doesn't own, right? I think this should generally be avoided and instead of things like `self.output = input` we should do `self.output:set(input)`. This also has the advantage that one can keep a reference to the module's output from outside (as is the case for most other modules).

szagoruyko:
@dominikgrewe I'd rather merge this and fix these cases `self.output = input` in a separate PR.

hughperkins:
I'm just a bystander, since not using this, but is there any reason you couldn't have a generic method in `Module` that does something like:

```
function clearState()
   for k,v in pairs(self) do
      if torch.type(k):find('Tensor') then
      self[k] = nil
   end
end
```

fmassa:
@hughperkins this would erase all module weights/bias. Some generic solutions were proposed in the past, like the [sanitize](https://github.com/soumith/imagenet-multiGPU.torch/blob/40b62729b6a243904f8092d40f7de190fafdabd6/train.lua#L129-L139) from @soumith, but it turns out that those generic functions never worked in all cases, so a module-specific solution should be implemented in order for it to always work.

hughperkins:
Ah, I see.  Good point :-)

hughperkins:
Hmmm... is the name misleading in fact?  Should it be perhaps `clearBuffers`, rather than `clearState`?  `clearState` to my mind would imply: clear everything, including weights.

fmassa:
@hughperkins well, it not only clear the internal buffers, but also the state variables `output` and `gradInput`. But I agree that the name should be discussed (it's even written in the first post)

hughperkins:
(By the way, if anyone is wondering, why would one ever want to clear the weights? well... in [char-rnn-er](https://github.com/hughperkins/char-rnn-er), I'm [saving the current network](https://github.com/hughperkins/char-rnn-er/blob/364de5e6486bf1eb2c88a03ecf5eb79be12f70f9/train.lua#L366-L372) , and in order to avoid going down the rabbit hole of figuring out how to [convert between saved cuda-network, cpu-network, cl-network](https://github.com/karpathy/char-rnn/issues/35) , so I save in two parts:
- network architecture, ie how many hidden layers, size of each layer.  Essentially this could be saved as the network, without buffers, and also without weights, though I'm simply saving the network sizes for now, [define netParams](https://github.com/hughperkins/char-rnn-er/blob/364de5e6486bf1eb2c88a03ecf5eb79be12f70f9/train.lua#L93) , [store them](https://github.com/hughperkins/char-rnn-er/blob/364de5e6486bf1eb2c88a03ecf5eb79be12f70f9/train.lua#L368)
- the weights themselves, which I simply clone to Floats, then write to disk [get params](https://github.com/hughperkins/char-rnn-er/blob/364de5e6486bf1eb2c88a03ecf5eb79be12f70f9/train.lua#L216) [store params](https://github.com/hughperkins/char-rnn-er/blob/364de5e6486bf1eb2c88a03ecf5eb79be12f70f9/train.lua#L369)

The first part, saving the network architecture, could plausibly be accomplished by saving the class structure of the network, including the module hyper-parameters (size of each Linear for example), but with neither weights nor buffers (nor hidden state, in case of LSTM module).
)

fmassa:
@hughperkins this issue of converting from cuda/float (with cudnn) is going to be addressed by https://github.com/soumith/cudnn.torch/pull/76 .
The problem with your approach (if you save only the definition, without the buffers, weights, etc) is that you will need sometimes to recreate the sharing when you load the model, which can be painful.

This PR addresses the issue of not having to hack around saving parameters separately from the model definition. You simply call `clearStates` before saving (eventually converting to CPU to make it loadable everywhere), this will keep all the necessary information without the buffers.

Both PRs combined make model definition/saving/loading much simpler, without having to write specific code for converting between frameworks (nn/cudnn), as done in [here](https://github.com/fmassa/object-detection.torch/blob/refactoring/models/frcnn_alexnet.lua) for example

hughperkins:
> The problem with your approach (if you save only the definition, without the buffers, weights, etc) is that you will need sometimes to recreate the sharing when you load the model, which can be painful.

Yes.  Hence, why I dont attempt to save the network, just the hyperparameters used to create it :-)  But, you are right, shared weights means that just storing the network object hierarchy, without weights, is not a generic solution to the problem.  (Hmmm, unless we cached the function calls used to define the shared weights, within each object perhaps, so we can rerun those on demand).

hughperkins:
I do however feel intuitively somehow that separating out the definition of the network, from the weight parameters, is somehow clean, and relatively unpainful.  Especially since we already have the technology to suck all the weights from a network, and drain them to disk, using `getParameters()`.  ie, it'd be nice to be able to do something like:

```
net:writeDefinition('somefile.lua')
$ ls -lh somefile.lua

somefile.lua  2k

$ cat somefile.lua

local function make_net()
   local net = nn.Sequential()
   net:add(nn.Linear(32,16))
   net:add(nn.LogSoftMax())
   net:shareWeights(... whatever we do to do this...)
   return net
end
```

fmassa:
@hughperkins for that, each module need to know how to reconstruct it's own constructor from itself. As we only accept ordered arguments in nn (and some input arguments are not stored in `self`, see for example the [Linear module](https://github.com/torch/nn/blob/master/Linear.lua#L3-L9)), this would require to manually go through all the modules and write a `:writeDefinition` to it, which corresponds to as much work as the current clearState PR.

CosineDistance.lua
```diff
@@ -73,6 +73,10 @@ function CosineDistance:updateGradInput(input, gradOutput)
       not_batch = true
    end
 
+   if #self.gradInput ~= 2 then
```
fmassa: `clearState` is missing here, it should clean the [following fields](https://github.com/szagoruyko/nn/blob/clearState/CosineDistance.lua#L33-L38).

*Not important note*: I'd do in here for consistence something like
```lua
self.gradInput[1] = self.gradInput[1] or v1.new()
self.gradInput[2] = self.gradInput[2] or v1.new()
```

DotProduct.lua
```diff
@@ -26,6 +26,11 @@ function DotProduct:updateGradInput(input, gradOutput)
    local v2 = input[2]
    local not_batch = false
 
+   if #self.gradInput ~= 2 then
```
fmassa: Same comment as the one from `CosineDistance`. Also, the [buffer](https://github.com/szagoruyko/nn/blob/clearState/DotProduct.lua#L16) should also be cleaned

SpatialBatchNormalization.lua
```diff
@@ -182,3 +182,13 @@ function BN:accGradParameters(input, gradOutput, scale)
       self.gradBias:add(scale, self.buffer2) -- sum over mini-batch
    end
 end
+
+function BN:clearState()
+   self.buffer = nil
```
fmassa: The buffers have changed name with https://github.com/torch/nn/pull/550. I think you should clean now [`save_mean` and `save_std`](https://github.com/torch/nn/blob/master/SpatialBatchNormalization.lua#L76-L77)

szagoruyko: ugh, that's why I suggest to merge this PR in its current state and require every new module/PR to handle clearState. will fix the comments when I have a minute, thanks Francisco.

dominikgrewe:
I think we should have some clear guidelines though as to how `clearState` (or `clearBuffers` or maybe `clearTemporaries`?) should behave. How about something like this:

1. After calling the function, all temporary tensors should be in the same state as they were when the module was constructed.
2. Any outside references to temporary tensors should be preserved.

davidsaxton:
Some missing modules / tensors: Cosine, L1Penalty (loss), Linear (addBuffer), LookupTable (_count, _input, copiedInput), Mean (_gradInput), MixtureTable, PairwiseDistance (outExpand, grad, ones), Reshape (_input, _gradOutput), RRelu (noise), SoftMin (mininput), SoftSign, SparseLinear (lastInput), SpatialDivisiveNormalization, SpatialSubtractiveNormalization (don't forget its internal modules).

Also as Dominik says, it would be good to preserve outside references to temporary tensors where possible. So the default behaviour of Module:clearState should be self[f]:set() for tensors (and equivalent for tables). Where this is not appropriate (e.g., nn.Identity()), these modules can override the default behaviour to call new() instead of set().

soumith:
@szagoruyko @dominikgrewe's last comment with the guidelines is pretty reasonable. Can we get this PR to a conclusion soon?

szagoruyko:
@dominikgrewe @soumith it requires big a change in nn, I am not going to do that. We can wait for someone to do it in a separate PR and then return to `clearState`.
@davidsaxton thanks, I will try to find time to add these.

szagoruyko:
alright, addressed @dominikgrewe and @davidsaxton comments. There is now only `set` everywhere, except `Identity.lua` and `Container.lua` where I had to add special `new` clearState. This is not the way I imagined it to do, but it can be fixed later. Tests are passing, rebased on master.

SpatialMaxPooling.lua
```diff
@@ -42,6 +43,7 @@ function SpatialMaxPooling:updateOutput(input)
       self.padW, self.padH,
       self.ceil_mode
    )
+   self.indices = self.indices or input.new()
```
soumith: this line should go before the THNN.*updateOutput call

soumith:
there seem to still be changes needed from comments from me / francisco. is it the github diff tool, or did you not update? for example in CosineDistance and in DotProduct.

szagoruyko:
@soumith thanks, missed that. updated and added clearState for Volumetric and Temporal poolings.

BatchNormalization.lua
```diff
@@ -154,3 +154,16 @@ function BN:accGradParameters(input, gradOutput, scale)
...
+function BN:clearState()
+   for i,v in ipairs{
+      'buffer',
+      'buffer2',
+      'centered',
+      'std',
+      'normalized',
+   } do
+      if self[v] then self[v]:set() end
+   end
```
dominikgrewe: Might be worth factoring that out in to a utils functions, e.g. `nn.utils.clearTensors(self, {...})`

szagoruyko: I had this at some point: https://gist.github.com/szagoruyko/bd39620cdf35964647b2 what do you think?

szagoruyko:
added `nn.util.clear` and doc.

soumith:
i just did a pass on it, and looks good to me. once dominik signs off on it, it's ready for squash and merge. dont need to squash into 1 commit, but rebase so that the bugfix commits and the "set everywhere everywhere" are squashed up.

PReLU.lua
```diff
@@ -6,11 +6,11 @@ function PReLU:__init(nOutputPlane)
    self.nOutputPlane = nOutputPlane or 0
    self.weight = torch.Tensor(nOutputPlane or 1):fill(0.25)
    self.gradWeight = torch.Tensor(nOutputPlane or 1)
-   self.gradWeightBuf = torch.Tensor()
-   self.gradWeightBuf2 = torch.Tensor()
 end
 
 function PReLU:updateOutput(input)
+   self.gradWeightBuf = self.gradWeightBuf or input.new()
+   self.gradWeightBuf2 = self.gradWeightBuf2 or input.new()
```
dominikgrewe: Why is this lazy initialization required?

szagoruyko: 2 reasons
* this buffers are needed just because cutorch reduce in not generic enough and cannot reduce with += to a buffer, so should be removed sometime later
* I want to keep all buffers like this out of constructor for backend switching like `cudnn.convert`

dominikgrewe: So it is unrelated to the introduction of `clearState`?

szagoruyko: yes, this is refactoring

Module.lua
```diff
@@ -364,3 +364,7 @@ function Module:listModules()
    end
    return modules
 end
+
+function Module:clearState()
+   return nn.utils.clear(self, 'output', 'gradOutput')
```
dominikgrewe: gradOutput -> gradInput

PReLU.lua
dominikgrewe: Would you mind moving this to a separate PR in that case? This one is already big enough as it is.

szagoruyko: agree, will factor out now

Module.lua
szagoruyko: oops, fixed now

dominikgrewe:
`nn.Min` and `nn.Max` also need a `clearState` method.

utils.lua
```diff
@@ -162,4 +162,27 @@ function nn.utils.contiguousView(output, input, ...)
...
+   local arg = {...}
+   if #arg > 0 and type(arg[1]) == 'table' then
+      arg = arg[1]
+   end
+   local function clear(f)
+      if self[f] then
+         if torch.isTensor(self[f]) then
+            self[f]:set()
+         elseif type(self[f]) == 'table' then
+            self[f] = {}
```
GeorgOstrovski: for tables, this should probably recurse into the table - that way external references to that table or the tensors in it remain valid after the clearState() call.

szagoruyko:
factored out lazy inits and added Min and Max.

utils.lua
szagoruyko: yes, that's true, but I am a little bit afraid of doing that, who knows what's referenced there.

GeorgOstrovski: Not sure I understand - as long as nn.utils.clear is only applied to fields of the module such as output and gradInput, or internal buffers, what could be referenced there that would be a problem?

szagoruyko: for example you want to call clearState on one part of a network and it's connected to another part through SelectTable or NarrowTable like modules, then set() clearState will destroy tensors it points to, and everything else that was referenced there. Normally we should add new() clearState to all modules that reference tensors in tables like this before enabling set() in nn.Module I think.

andreaskoepf: What do you think about calling `collectgarbage()` at he end of `nn.utils.clear()` to free tensors that are no longer referenced (e.g. stuff built by [zeroTableCopy()](https://github.com/torch/nn/blob/0c77df0fcae2fdb0f2902fc88a57d30f3fa75cc2/SelectTable.lua#L20) in SelectTable) after the `= {}` replacement?

andreaskoepf:
For a v1 of clearState() this PR looks ready to me. +1 for merging this soon.

soumith:
all hail the sergey! thanks for the huge PR

Atcold:
I'm having some troubles with `:clearState()` and `nngraph`... it looks like not much is cleared. Perhaps I'm doing something wrong...

PR #532, opened 2015-12-15
the statements are redundant

xwgeng:
error messages statements are redundant


soumith:
thanks. nice that you're catching all these.

PR #535, opened 2015-12-15
criterion.md: Fix typos

tindzk:


doc/criterion.md
```diff
@@ -108,7 +108,7 @@ The loss can be described as:
 loss(x, class) = -x[class]
 ```
 
-or in the case of the `weights` argument being specified:
+or in the case of the `weights` argument may be specified:
```
fmassa: English is not my mother language, but wasn't "being" ok in this situation ?

PR #537, opened 2015-12-17
Add SpatialMaxUnpooling

clement-masson:
*I messed a little with the branch so I've cleaned everything and do a new PR. 
There is now just 1 commit*

Hi 

Those days, several people including @fmassa, @andreaskoepf and I, have been discussing on gitter about Convolutional AutoEncoder and max-unpooling in particular. This PR proposes a C implementation of max-unpooling reusing the indices of max values previously computed by a max-pooling module. A CUDA implementation has already been made and can be proposed if this proposition gets accepted.

#### Origins
The first paper (I read) using unpooling as it is proposed here is by Matthew Zeiler and Rob Fergus ["Visualizing and Understanding Convolutional Networks"](https://www.cs.nyu.edu/~fergus/papers/zeilerECCV2014.pdf) (2014) though they had already introduced another type of unpooling in 2011 in ["Adaptive Deconvolutional Networks for Mid and High Level Feature Learning"] (https://www.cs.nyu.edu/~gwtaylor/publications/zeilertaylorfergus_iccv2011.pdf).
It has also been recently used in ["Learning Deconvolution Network for Semantic Segmentation"](http://arxiv.org/pdf/1505.04366v1.pdf) by Noh et al.

Unpooling, from Zeiler & Fergus (2014)')

#### What it actually does
+ First, this SpatialMaxUnpooling module's constructor only takes a nn.SpatialMaxPooling module, say "pooler", as argument and store it in a member variable. It slighty changes pooler's `updatOutput(input)`: the latter now stores in two pooler's member variables the sizes (width and height) of input. This is necessary for the "unpooler" to be able to correctly place the max value back to their orignal position since the relation inputSize -> outputSize is not bijective for a pooling module, due to the ceil or floor.

+ when `unpooler:updateOutput(input)` is called, it resizes its output according to the sizes stored in its pooling module "pooler", then goes through input's elements and puts their value in output at the position given by the corresponding index stored in pooler.indices. (see next section for corner cases)

+ `unpooler:updateGradInput(input,gradOutput)` simply puts in gradInput the values in gradOutput located at the positions corresponding to pooler.indices.

#### Special -undefined- cases
In the case where the pooling module has a filter size (let's say kH) larger than its stride (dH) then the same input element in an input map (at the particular index n) can be the max-value associated to 2 (or more) neighboring elements in the pooled map (let note p1 and p2 their index in the pooled map). Then, if an SpatialMaxUnpooling is attached to that pooling module, when it will receive an input to unpool, it is not clear if the element at position n in the "unpooled" map should have the value of the element figuring at position p1 or p2 in the input.
The solutions we saw with @andreaskoepf  are:
- compelling unpooling modules to be associated to poolingmodules such that stride = filterSize (this is what is enforced in the current version of the code)
- last-write-win: scatter the max values from the input to the unpooled map in a deterministic order and the last input element associated with index n wins
- randomized last-write-win: same as previous one but with a randomized order
- making an average of every input element associated with the same element in the unpooled map.

*It should be noted that the last 3 solutions do not pass the test "nn.Jacobian.testJacobian" which tests that the Jacobian given by backward is similar to the one computed by calling foward with two slightly different inputs.*

#### Code efficiency
I am not a C expert, especially when it comes to performance. Much of this code actually comes from nn.SpatialMaxPooling so there should not be any massive issues. Though, I wonder if the test I perform in `updateOutput_frame` and `updateGradInput_frame` to check whether `(maxp<0 || maxp>=owidth*oheight)` is the optimal soution. Does this simple check (which, if everithing is normal, should never be true) have a real influence on the performance, is there a smarter way to check ?


soumith:
I had a question about serialization.
After the model is saved to disk (torch.save), and loaded again (torch.load), will this Unpooling module still work correctly?

clement-masson:
@soumith Good remark.
I did not think about this when I chose to store a ref to the pooling module in the unpooling one.

I just performed a quick test ([here is the gist](https://gist.github.com/clement-masson/afb61769c3e2a9450cea)) and it seems it works quite well.
As you can see, you only need saving the unpooler. To load from disk, you just have to load the unpooler, and you can retrieve the pooler by doing `pooler = unpooler.pooling`

clement-masson:
I also did a test in which I build a bigger net including a pair pooler/unpooler to see how it behaves after serialization: [cf test2 of the same gist](https://gist.github.com/clement-masson/afb61769c3e2a9450cea)

It seems to be ok. The link between the two modules is preserved, and even the indices before saving are kept.

albanD:
@clement-masson have you tried the same test with the input [here](https://gist.github.com/clement-masson/afb61769c3e2a9450cea#file-test2-lua-L28) being bigger than the one used during the [first forward](https://gist.github.com/clement-masson/afb61769c3e2a9450cea#file-test2-lua-L11)

The problem pointed out by @soumith I think is that the `updateOutput` method of the `pooler` after loading will be the original one and not your patched one.

clement-masson:
@albanD Indeed ...

+ In the case of test1, it's ok: the updateOutput of the loaded pooler seems to be mine.
When I add 
```
print(pooler2.iwidth)
print(pooler2.iheight)
```
then the printed width and height correspond to the actual input X given after loading from disk, even if it is larger/smaller than the one given before saving.

+ However, in the case of test2 when the unpooler is not directly saved, the `pooler2.iwidth` and `pooler2.iheight` correspond to the sizes of the input that was given before saving, which suggest that the updateOutput method of the loaded pooler is the orignal one, not mine. Yet the loaded unpooler module has its member variable 'pooling' that refers to the loaded pooler ....

How can you explain the difference bewteen the two cases ?
I've tried to have a look at the readObject method of File.lua but I did not understand everything.
Do you know how we can preserve the method of an object when saving it ?
How could we mimic the execution of the SpatialMaxUnpooling:__init() method in order to force the loaded pooler has the patched version of updateOutput, not the original ?

Otherwise, simple but not elegant solutions would be:
- replace current nn.SpatialMaxPooling:updateOutput by the patched one
- create a tiny class that inherit nn.SpatialMaxPooling and that overloads updateOutput.

andreaskoepf:
@clement-masson It might be possible to do the pataching in a custom `:read(file, version)` member-function, see https://github.com/torch/nn/issues/503.

clement-masson:
@andreaskoepf Thanks ! I have quickly looked at #503 and I think writing a custom read or a restore function as you did can be the solution !
I'll give it a try.

albanD:
@clement-masson 
tl-dr: after having a better look: my bad, your module should work as is.
I think you made a type in your test script [ here](https://gist.github.com/clement-masson/afb61769c3e2a9450cea#file-test2-lua-L29), it should be `net ` (thats why your second test was failing)

for the serialization of torch classes:
* Torch classes are regular table + a metatable (the content of this metatable is what you access when you do `nn.SpatialMaxPooling.updateOutput`)
* When serializing, it saves the class name and the table part
* When deserializing, it creates a new empty table, associate to it the metatable for the class name and then add back the content of the table

After looking more in details, it seems that when you do `poolingModule.updateOutput = function(pool, input)`  you actually change the table part of the pooling module. In your example, after loading the net, you can print
```lua
print(net:get(2).updateOutput)
print(nn.SpatialMaxPooling.updateOutput)
```
They are different meaning that it still contains your patched version

andreaskoepf:
@albanD thanks for mentioning that you can do stuff like torch.serialize(function(a,b) return a+b end) - even with upvals.

clement-masson:
@albanD You were totally right, I did a stupid copy/paste mistake :( 
It indeed works in case 2 too !

When printing the two updateOutput refs, they are indeed different. (I updated the gist).
So there is no serialization problem it seems ... !

soumith:
Thanks a lot Clement!

PR #542, opened 2015-12-23
Compute Normalize without bmm and baddbmm

adamlerer:
Remove bmm and baddbmm from Normalize, because they allocate memory, causing sync on CUDA. This version is much faster at least in cutorch.

### DIFF
--- a/Normalize.lua
+++ b/Normalize.lua
@@ -61,9 +61,7 @@ function Normalize:updateGradInput(input, gradOutput)
   self._gradInput = self._gradInput or input.new()
   self.cross = self.cross or input.new()
   -- compute diagonal term with gradOutput
-  self._gradInput:resize(n,d,1)
-  gradOutput = gradOutput:view(n,d,1)
-
+  self._gradInput:resize(n,d)
   if self.p == math.huge then
     -- specialization for the inf case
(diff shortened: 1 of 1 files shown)


soumith:
Thanks Adam.

fmassa:
@adamlerer @soumith @dominikgrewe  does that means that `bmm` and `baddbmm` should be avoided if we want the code to be non-blocking ?
The memory allocation is done [here](https://github.com/torch/cutorch/blob/master/lib/THC/THCTensorMathBlas.cu#L329-L353), and I think the reason for it is that [`cublasSgemmBatched` expects `float **`](http://docs.nvidia.com/cuda/cublas/index.html#cublas-lt-t-gt-gemmbatched), and that forcing a cast from `float *` (the tensor data) to `float **` is [dangerous](https://isocpp.org/wiki/faq/const-correctness#constptrptr-conversion).

Is there a way avoiding this memory allocation/deallocation in `baddbmm`, or to make them non-blocking ? How *dangerous* would be doing a casting in this situation ?

soumith:
@fmassa wickedfoo is working on a patch for this, to avoid alloc/dealloc.

fmassa:
Nice ! Thanks @soumith !

PR #544, opened 2015-12-24
Add "Deconvolution" note to SpatialFullConvolution docs

gcr:
This commit simply adds a greppable note to the docs for `SpatialFullConvolution` to let people find it.

I agree that the terminology used by other frameworks is incorrect (particularly the use of the word "deconvolution"), but this layer is impossible to understand without it. After reading so many papers that refer to "deconvolution" or "upconvolution" layers, I strongly feel that a note like this in the docs is necessary. When I first read the docs for `SpatialFullConvolution`, I thought that it was just convolution with a full connection table since it comes right after the connection-table convolution functions. I wouldn't have ever guessed that this was really deconvolution with a different name.

Other discussions:
https://github.com/torch/nn/pull/405
https://github.com/facebook/eyescream/issues/7


soumith:
Thanks, that was probably needed.

PR #545, opened 2015-12-25
nn.Contiguous

szagoruyko:
Very simple module. Can be used instead of reshape to save memory when either `gradOutput` or `input` is not contiguous, and does not require size arguments.


Contiguous.lua
```diff
@@ -0,0 +1,19 @@
+local Contiguous, parent = torch.class('nn.Contiguous', 'nn.Module')
+
+function Contiguous:updateOutput(input)
+   if not input:isContiguous() then
+      self.output:resizeAs(input):copy(input)
+   else
+      self.output = input
```
soumith: can you make this to be :set, instead of =. Also, for gradInput below.

soumith: ping!

szagoruyko: fixed

fmassa:
Is there some need for unit tests for this module ?

PR #546, opened 2015-12-30
add local connected layer: SpatialConvolutionLocal.lua

yangky11:
This PR adds a local-connected layer called SpatialConvolutionLocal, which is similar to convolutional layer but without weight-sharing.
This layer has the same name and functionality as its conterpart in the [ccn2 package](https://github.com/soumith/cuda-convnet2.torch).  
However:

1. ccn2 cannot run on CPU  
2. ccn2 requires the batch size to be a multiple of 32, which brings too much restriction. (Once I wanted to use local-connected layer but my network was so large that a batch size of 10 would fill the entire 12GB GPU memory)

The implementation is adapted from SpatialConvolutionMM and has passed [some tests](https://gist.github.com/yangky11/5192576ad6717797de80). I'm still working on testing, documentation, and GPU implementation.


generic/SpatialConvolutionLocal.c
```diff
@@ -0,0 +1,414 @@
...
+#else
+
+#ifdef _WIN32
+# include <windows.h>
+#endif
+
+
+
+/* note: due to write issues, this one cannot be parallelized as well as unfolded_copy */
+static void nn_(spatialconvolutionlocal_unfolded_acc)(THTensor *finput, THTensor *input,
```
yangky11: This function is copied from SpatialConvolutionMM.c without modification

generic/SpatialConvolutionLocal.c
```diff
@@ -0,0 +1,414 @@
...
+                THVector_(add)(dst+(size_t)(iy*inputWidth+ix+x*dW), src+(size_t)(y*outputWidth+x), 1, 1);
+            }
+          }
+        }
+      }
+    }
+  }
+}
+
+static void nn_(spatialconvolutionlocal_unfolded_copy)(THTensor *finput, THTensor *input,
```
yangky11: This function is copied from SpatialConvolutionMM.c without modification

generic/SpatialConvolutionLocal.c
```diff
@@ -0,0 +1,414 @@
...
+        else{
+          for (x=0; x<outputWidth; x++)
+             memcpy(dst+(size_t)(y*outputWidth+x), src+(size_t)(iy*inputWidth+ix+x*dW), sizeof(real)*(1));
+         }
+      }
+    }
+  }
+}
+
+static void nn_(SpatialConvolutionLocal_updateOutput_frame)(THTensor *input, THTensor *output, THTensor *weight, THTensor *bias, THTensor *finput,
```
yangky11: change the `addmm` in SpatialConvolutionMM.c to `baddbmm`

generic/SpatialConvolutionLocal.c
```diff
@@ -0,0 +1,414 @@
...
+      THTensor_(free)(output_t);
+      THTensor_(free)(finput_t);
+    }
+  }
+
+  return 1;
+}
+
+
+static void nn_(SpatialConvolutionLocal_updateGradInput_frame)(THTensor *gradInput, THTensor *gradOutput, THTensor *weight, THTensor *fgradInput,
```
yangky11: change the `addmm` in SpatialConvolutionMM.c to `baddbmm`

generic/SpatialConvolutionLocal.c
```diff
@@ -0,0 +1,414 @@
...
+  THTensor_(free)(gradOutput3d);
+  THTensor_(free)(fgradInput3d);
+  
+  THTensor_(zero)(gradInput);
+
+  nn_(spatialconvolutionlocal_unfolded_acc)(fgradInput, gradInput, kW, kH, dW, dH, padW, padH, 
+                                            nInputPlane, inputWidth, inputHeight, outputWidth, outputHeight);
+}
+
+static int nn_(SpatialConvolutionLocal_updateGradInput)(lua_State *L)
```
yangky11: Nothing substantially different from its SpatialConvolutionMM conterpart.

generic/SpatialConvolutionLocal.c
```diff
@@ -0,0 +1,414 @@
...
+      THTensor_(free)(fgradInput_t);
+    }
+  }
+
+  THTensor_(transpose)(weight, weight, 1, 2);
+
+  return 1;
+}
+
+static void nn_(SpatialConvolutionLocal_accGradParameters_frame)(THTensor *gradOutput, THTensor *gradWeight, THTensor *gradBias, THTensor *finput, real scale,
```
yangky11: change the addmm in SpatialConvolutionMM.c to baddbmm

generic/SpatialConvolutionLocal.c
```diff
@@ -0,0 +1,414 @@
...
+  THTensor_(baddbmm)(gradWeight, 1.0, gradWeight, scale, gradOutput3d, finput3d);
+  // gradWeight:    oH*oW x nOutputPlane x kW*kH*nInputPlane
+
+  THTensor_(cadd)(gradBias, gradBias, 1.0, gradOutput);
+
+  THTensor_(free)(gradOutput3d);
+  THTensor_(free)(finput3d);
+}
+
+static int nn_(SpatialConvolutionLocal_accGradParameters)(lua_State *L)
```
yangky11: Nothing substantially different from its SpatialConvolutionMM conterpart.

generic/SpatialConvolutionLocal.c
```diff
@@ -0,0 +1,414 @@
...
+  // weight:    oH*oW x nOutputPlane x nInputPlane*kH*kW
+  // finput3d:  oH*oW x nInputPlane*kH*kW x 1  
+  THTensor_(baddbmm)(output3d, 1.0, output3d, 1.0, weight, finput3d);
+  // output3d:  oH*oW x nOutputPlane x 1
+  
+  THTensor_(free)(output3d);
+  THTensor_(free)(finput3d);
+}
+
+static int nn_(SpatialConvolutionLocal_updateOutput)(lua_State *L)
```
yangky11: Nothing substantially different from its SpatialConvolutionMM conterpart.

SpatialConvolutionLocal.lua
```diff
@@ -0,0 +1,166 @@
+local SpatialConvolutionLocal, parent = torch.class('nn.SpatialConvolutionLocal', 'nn.Module')
```
yangky11: Nothing substantially different from `SpatialConvolution.lua`

fmassa:
Nice addition !
Just wondering, is SpatialConvolutionLocal a good name for it ? It's not really a convolution, but a locally connected layer, as you mention in the description.

yangky11:
@fmassa  thanks for the feedback!
Yes, the name is a little confusing. I just adopted the name in the cc2 package and didn't think much about it.
Any suggestion?  Maybe `SpatialLocalConnected` or something else?

yangky11:
I just made a copy of the tests for `SpatialConvolution` in test.lua and adpated them for `SpatialConvolutionLocal`.
All tests are OK.

test.lua
```diff
@@ -1935,6 +1935,112 @@ function nntest.SpatialConvolutionMM()
    mytester:asserteq(0, (gradInput-gradInputc):abs():max(), torch.typename(module) .. ' - contiguous err ')
 end
 
+function nntest.SpatialConvolutionLocal()
```
yangky11: Almost the same as `nntest.SpatialConvolution`

generic/SpatialConvolutionLocal.c
fmassa: I think it would be better to factor out `unfolded_acc` and `unfolded_copy` in a separate file, as done in the GPU version, to avoid redundant code.

SpatialConvolutionLocal.lua
```diff
@@ -0,0 +1,166 @@
...
+   self.kW = kW
+   self.kH = kH
+   self.iW = iW
+   self.iH = iH
+
+   self.dW = dW
+   self.dH = dH
+   self.padW = padW or 0
+   self.padH = padH or self.padW
+   self.oW = math.ceil((self.padW * 2 + iW - self.kW) / self.dW + 1)
```
fmassa: in both `SpatialConvolution` and `SpatialConvolutionMM` the output dimensions are [computed using `floor`](https://github.com/torch/nn/blob/master/generic/SpatialConvolutionMM.c#L207-L208). Wouldn't it be better to keep it consistent here as well ?

SpatialConvolutionLocal.lua
```diff
@@ -0,0 +1,166 @@
...
+local function unviewWeight(self)
+   self.weight = self.weight:view(self.oH, self.oW, self.nOutputPlane, self.nInputPlane, self.kH, self.kW)
+   if self.gradWeight and self.gradWeight:dim() > 0 then 
+      self.gradWeight = self.gradWeight:view(self.oH, self.oW, self.nOutputPlane, self.nInputPlane, self.kH, self.kW)
+   end
+end
+
+local function checkInputSize(self, input)
+   if input:nDimension() == 3 then
+      if not input:isSize(torch.LongStorage{self.nInputPlane, self.iH, self.iW}) then
```
fmassa: you are allocating a new tensor at every forward pass. I think it would be better to just do
```lua
input:size(1) == self.nInputPlane and input:size(2) == self.iH and input:size(3) == self.iW
```
Same thing for the `checkOutputSize` function

fmassa:
Will finish reviewing this PR this weekend.

SpatialConvolutionLocal.lua
yangky11: Sure

yangky11:
Thanks Francisco!
GPU version here https://github.com/torch/cunn/pull/183
Will look into the reviews in the weekend

SpatialConvolutionLocal.lua
yangky11: yes

generic/SpatialConvolutionLocal.c
yangky11: I thought about it before but the problem is that, previously each file in `nn/generic` corresponds to a module. If I add something like `unfold.c`, `unfold.h`, it may looks a little inconsistent.  
Anyway, I've just made a commit with `unfolded_copy` and `unfolded_acc` in a separate file. You can see if it's better.

fmassa:
I had another pass on it and I think it's good.

Could you add another test checking `SpatialConvolutionLocal` against `SpatialConvolution` ? Something similar to what you have in your gist, but with non-constant weight matrix ? Simple example
```lua
-- define mConvLocal and mConv
mConv.weight:zero(); mConvLocal.weight:zero();
mConv.bias:zero();   mConvLocal.bias:zero();
mConv.weight[1][1][1][3]=1
mConvLocal.weight[{{},{},1,1,1,3}]=1
```
This would be helpful to check correctness against standard convolution. For example, I initially thought that you would have transposed outputs (wrt standard convolution), but then I noticed that the matrices you feed to the `baddbmm` [are not contiguous](https://github.com/torch/nn/pull/546/files#diff-dadde21317b8022e47d4d48adce947cbR25), making it work as expected.

One drawback is that memory is allocated at every `updateOutput`,`updateGradInput` and `accGradParameters` by `baddbmm` (as it expects contiguous tensors), but I think that's fine.

Also, could you fix the merge issues, rebase and squash the commits ?

SpatialConvolutionLocal.lua
```diff
@@ -0,0 +1,166 @@
...
+   self.kW = kW
+   self.kH = kH
+   self.iW = iW
+   self.iH = iH
+
+   self.dW = dW
+   self.dH = dH
+   self.padW = padW or 0
+   self.padH = padH or self.padW
+   self.oW = (self.padW * 2 + iW - self.kW) / self.dW + 1
```
fmassa: This needs to be fixed by adding a `floor`.
If not, you can have wrong results when viewing the weights, as in https://github.com/torch/nn/pull/546/files#diff-be444d1c4584d74f86d51ed473de5751R71

yangky11:
All set!
The tests against nn.SpatialConvolution are in `test.lua`

SpatialConvolutionLocal.lua
```diff
@@ -0,0 +1,166 @@
...
+   self.kW = kW
+   self.kH = kH
+   self.iW = iW
+   self.iH = iH
+
+   self.dW = dW
+   self.dH = dH
+   self.padW = padW or 0
+   self.padH = padH or self.padW
+   self.oW = math.floor(self.padW * 2 + iW - self.kW) / self.dW + 1
```
fmassa: There are missing parenthesis, it should be instead
```lua
self.oW = math.floor((self.padW * 2 + iW - self.kW) / self.dW) + 1
```

yangky11:
missing parenthesis fixed

fmassa:
I think it's good now, thanks !
Could someone else have a look ? cc @soumith

soumith:
The CPU version looks good. I have a minor change needed in the CUDA version, I'll comment there.

soumith:
Thanks a lot Kaiyu!

PR #547, opened 2015-12-30
THNN functional module conversions batch 1

andreaskoepf:
- [x] ClassNLLCriterion
- [x] DistKLDivCriterion
- [x] HardShrink
- [x] HardTanh
- [x] L1Cost
- [x] final cross-check test with old implementation


hughperkins:
For some reason https://github.com/torch/nn/commit/ad1efeed343a6f82593a15e78ea7e7bd6ceb041c has broken clnn.  It would be less of a mystery if it seemed not to break either nn or cunn at the same time :-)  It seems like nn/Abs.lua expects input to contain input.THNN now, though how THNN gets into input is a mystery to me, for now.

soumith:
oh right. Totally forgot that this would break clnn Hugh! Sorry about that.

hughperkins:
:-D

ClassNLLCriterion.lua
```diff
@@ -1,18 +1,16 @@
...
-    'nn.Criterion'
-)
+local THNN = require 'nn.THNN'
+local ClassNLLCriterion, parent = torch.class('nn.ClassNLLCriterion', 'nn.Criterion')
 
 function ClassNLLCriterion:__init(weights, sizeAverage)
     parent.__init(self)
     if sizeAverage ~= nil then
-        self.sizeAverage = sizeAverage
+       self.sizeAverage = sizeAverage
```
hughperkins: some of this is 3-indented, and some is 4-indented :-P

I really hate 3-indentation by the way.   2-indentation is quite nice.  4-indentation is quite standard.  3-indentation is one of the things that makes me run away and write python wrappers instead :-P

andreaskoepf:
@hughperkins Sorry for breaking clnn, actually I wanted to ask @soumith about other backends two days ago but after I remembered that cudnn uses completely different modules I felt this might be a stupid question to ask: All backends that work by keeping the existing modules but a different tensor type are affected by the THNN change.

We are using the same 'metatable' trick to pass in the tensor-type specific C-backend functions that was done previously in C, e.g. we register under the name 'THNN' in the specific tensor-metatables (e.g. the metatables of DoubleTensor, FloatTensor, CudaTensor) a table that contains the new ffi C-functions.

Great to see that you already began working on mirroring the THNN functions in CL. btw you can reach the THNN team via gitter [here](https://gitter.im/apaszke/THNN).

BTW sorry for the indention-errors, I will fix these. I am using 2-space indention for my normal lua and c code and I am using an [editor](http://codelite.org/) which afiak does not allow to set project specific indentation rules .. I was just too lazy to change the setting and did the indentation by hand which is obviously not 100% ;-). Regarding 3-space indention I like the [rumour](http://lua-users.org/lists/lua-l/2014-01/msg00119.html) that Mike Pall once said he he would refuse to take anyone seriously who used three spaces.

andreaskoepf:
Compared against non-THNN versions the converted modules via [xtest.lua](https://github.com/andreaskoepf/nn/blob/nn2/xtest.lua). nn2 is based on the last commit before THNN was added.

soumith:
Thanks Andreas. This is coming along really well.

hughperkins:
Hmmm, seems by some miracle doesnt break clnn build https://travis-ci.org/hughperkins/clnn/builds/99973441 . I'm not quite sure why that is ....

PR #549, opened 2016-01-05
THNN functional module conversions batch 2

andreaskoepf:
- [x] ELU
- [x] LeakyReLU
- [x] LogSigmoid
- [x] LogSoftMax
- [x] LookupTable
- [x] Cross-check with old module implementation


lib/THNN/generic/THNN.h
```diff
@@ -91,4 +104,50 @@ TH_API void THNN_(L1Cost_updateGradInput)(
...
+          THTensor *output);
+
+TH_API void THNN_(LookupTable_accGradParameters)(
+          THNNState *state,
+          THLongTensor *input,
+          THTensor *gradOutput,
+          THTensor *gradWeight,
+          real lr,
+          bool shouldScaleGradByFreq,
+          THLongTensor* count);
```
fmassa: Wouldn't it be better to use `THIndexTensor` instead ? You have changed everywhere from `int` to `long`, I thought you would want to make it consistent with `THIndexTensor`

lib/THNN/generic/LookupTable.c
```diff
@@ -20,28 +20,23 @@ static void nn_(LookupTable_resetCount)(int *count_data, THLongTensor *input)
...
-
-  if (luaT_getfieldcheckboolean(L, 1, "shouldScaleGradByFreq"))
+  long *count_data = NULL;
+  
+  if (shouldScaleGradByFreq)
   {
-    THIntTensor *count = luaT_getfieldcheckudata(L, 1, "_count", "torch.IntTensor");
-    THIntTensor_resize1d(count, gradWeight->size[0]);
-    count_data = THIntTensor_data(count);
+    THLongTensor_resize1d(count, gradWeight->size[0]);
```
fmassa: Same thing here, wouldn't it be better to use `THIndexTensor_(resize1d)` ?

fmassa:
There are a few places where you explicitely used `THLongTensor`, but I think it would be better to use `THIndexTensor`. Same thing for the `long` type, instead of `TH_index_t`

lib/THNN/generic/LookupTable.c
```diff
@@ -2,7 +2,7 @@
 #define TH_GENERIC_FILE "generic/LookupTable.c"
 #else
 
-static void nn_(LookupTable_resetCount)(int *count_data, THLongTensor *input)
+void THNN_(LookupTable_resetCount)(long *count_data, THLongTensor *input)
```
andreaskoepf: Helper functions should be static.

LookupTable.lua
```diff
@@ -82,7 +95,7 @@ function LookupTable:type(type, tensorCache)
       self._input = self.weight.new()
    else
       -- self._count and self._input should only be converted if using Cuda
-      self._count = torch.IntTensor()
+      self._count = torch.LongTensor()
```
szagoruyko: I think it's a bad idea to change Int to Long here, might break something.

andreaskoepf:
@szagoruyko OK, will restore IntTensor. I already discussed this with @fmassa. Ideally the THNN.h file would be usable for the CPU and GPU backend .. but that would mean we need some additional macro beside THIndexTensor that expands to IntTensor on CPU and CudaFloatTensor (or later CudaIntTensor) in GPU .. maybe something like THIntTensor?

soumith:
Thanks Andreas!

hughperkins:
Hmmm, this breaks `clnn` :-(

hughperkins:
(breaks to clnn following this pull are fixed now, I think.  at least, tests pass now).

PR #550, opened 2016-01-05
Add C implementation of SpatialBatchNormalization

colesbury:
This is primarily to support the fast, memory-efficient CUDA
implementation. Some other changes include making weight and bias each
individually optional and averaging the variances instead of the
inverse standard deviation.


SpatialBatchNormalization.lua
```diff
@@ -58,127 +60,88 @@ function BN:__init(nFeature, eps, momentum, affine)
...
+   for k,v in pairs(var) do
+      if version < 2 and k == 'running_std' then
+         k = 'running_var'
+         v = v:cmul(v):pow(-1)
+      end
+      self[k] = v
    end
 end
+
+function BN:sanitize()
```
soumith: remove this

PR #551, opened 2016-01-06
THNN functional module conversion batch 3

fmassa:
Another set of conversions.
- [x] SpatialConvolutionMM
- [x] SpatialAdaptiveMaxPooling
- [x] SpatialAveragePooling
- [x] SpatialMaxPooling
- [x] cross-check with old module implementation
- [x] add cuda conversion
- [x] fix difference with cuda SpatialConvolutionMM interface


SpatialConvolutionMM.lua
```diff
@@ -70,19 +70,49 @@ function SpatialConvolutionMM:updateOutput(input)
...
+   input.THNN.SpatialConvolutionMM_updateOutput(
+      input:cdata(),
+      self.output:cdata(),
+      self.weight:cdata(),
+      self.bias:cdata(),
+      self.finput:cdata(),
+      self.kW, self.kH,
+      self.dW, self.dH,
+      self.padW, self.padH
+   )
```
szagoruyko: how are you going to add `groups` here, by adding a for loop in lua or in C side?

fmassa: I'm thinking about adding a loop in lua side iterating, like in cudnn. Do you think we need to do it in C (to make better use of streams in cuda, etc) ?

hughperkins:
Hi Andreas.  I'm just about to start a new job, and probably am not going to be able to devote quite as much time to supporting clnn at that point.  Dont suppose... would you be able to take ownership of ensuring that the THNN changes do not break clnn going forward?

hughperkins:
Basically, what you'd need to do is:
- `luarocks make` on the nn changes
- `luajit -l cltorch -e 'cltorch.test()'` => check runs ok, no failures
- `luajit -l clnn -e 'clnn.test()'` => check runs ok, no failures
- install [char-rnn](https://github.com/karpathy/char-rnn), cd into char-rnn directoyr, run `th train.lua`, let run for 30 seconds.  Check no errors, and loss is non-NaN.

hughperkins:
I've fixed clnn for this particular pull.  I can fix it pre-emptively in the future if need-be.  I would prefer ideally please:
- two weeks notice prior to merge
- add a checkpoint to the pull 'clnn updated for batch'

For completeness, it seems that fixing clnn is a fairly straightforward process.  What I did, for each broken module, is, taking 'SpatialAveragePooling' as an example:
- create a new file 'SpatialAveragePooling.lua'
- copy contents of ELU.lua into SpatialAveragePooling.lua
- search and replace 'ELU' with 'SpatialAveragePooling', case-sensitive
- add SpatialAveragePooling.lua to CMakeLists.txt , line 34, for `luasrc`
- add `include 'SpatialAveragePooling.lua'` to `init.lua`, around line 18
- that's it :-)
Rebuild, rerun tests, check pass.

Run tests twice:
- once after building from the new nn branch, check all pass
- once after building the current nn master, check all pass

Note that for SpatialConvolutionMM, slight nuance: need to also monkey patch AccGradParameters, just same approach as for updateOutput though, [patch in our new method](https://github.com/hughperkins/clnn/blob/fb0e6e6839526def42f46793d2dcc2a7fd8775a9/SpatialConvolutionMM.lua#L5) [write new method](https://github.com/hughperkins/clnn/blob/fb0e6e6839526def42f46793d2dcc2a7fd8775a9/SpatialConvolutionMM.lua#L25-L32)

andreaskoepf:
@hughperkins I think we will have converted all modules that have a corresponding clnn implementation soon. Since I have not yet had time to deeply look into clnn I will have to find some time to really check it out. I think we should in some way be able to take care of clnn as well.

lib/THNN/generic/SpatialAveragePooling.c
```diff
@@ -74,7 +62,7 @@ static int nn_(SpatialAveragePooling_updateOutput)(lua_State *L)
     THTensor_(resize4d)(output, input->size[0], nInputPlane, outputHeight, outputWidth);
   
   input = THTensor_(newContiguous)(input);
-  luaL_argcheck(L, THTensor_(isContiguous)(output), 1, "");
+  THArgCheck(THTensor_(isContiguous)(output), 1, "");
```
andreaskoepf: This was already the case in the old impl? Do you have an idea why no error message is provide and argument index is 1 (e.g. output was a field before but now is a real arg)? Same goes for ln. 177.

lib/THNN/generic/SpatialMaxPooling.c
```diff
@@ -61,18 +61,8 @@ static void nn_(SpatialMaxPooling_updateOutput_frame)(real *input_p, real *outpu
   }
 }
 
-static int nn_(SpatialMaxPooling_updateOutput)(lua_State *L)
+void THNN_(SpatialMaxPooling_updateOutput)(THNNState *state, THTensor *input, THTensor *output, THTensor *indices, int kW, int kH, int dW, int dH, int padW, int padH, int ceil_mode)
```
andreaskoepf: Is ceil_mode a boolean value? Maybe we could use type bool? It is btw interesting that indices historically is a floating point tensor (may become a problem for indices > 2^23 (size of float mantissa)).

lib/THNN/generic/SpatialAveragePooling.c
```diff
@@ -129,25 +117,10 @@ static int nn_(SpatialAveragePooling_updateOutput)(lua_State *L)
     }
   }
   THTensor_(free)(input);
-
-  return 1;
 }
 
-static int nn_(SpatialAveragePooling_updateGradInput)(lua_State *L)
+void THNN_(SpatialAveragePooling_updateGradInput)(THNNState *state, THTensor *input, THTensor *gradOutput, THTensor *gradInput, int kW, int kH, int dW, int dH, int padW, int padH, int ceil_mode, int count_include_pad)
```
andreaskoepf: same bool comment for count_include_pad

lib/THNN/generic/SpatialAveragePooling.c
fmassa: Good catch.
Will fix the error message (but could eventually call `newContiguous` as well) and the `THArgCheck` index (it was 1 because the `self` was the first argument before).
BTW, it's sometimes difficult to specify an index when several arguments are involved (like comparing the input with `kW`).

lib/THNN/generic/SpatialMaxPooling.c
fmassa: Historically, [`nn` was C89](https://github.com/torch/nn/pull/2#issuecomment-31919246) and there is no `bool` in C89, but that has changed and [since this commit](https://github.com/torch/nn/pull/550) `nn` is officially C99.
Plus, `THNN` already support the [`bool` macro](https://github.com/torch/nn/blob/master/lib/THNN/THNN.h#L4), so I should probably update it.

Regarding the indices being float, long ago [they were actually `long` but were moved to `float` to work on the gpu](https://github.com/torch/nn/pull/46#issuecomment-49786258), but with the addition of other tensor types to cutorch, we can soon start thinking about updating it.

fmassa:
Cross check tests added https://github.com/andreaskoepf/nn/commit/a7296ca774b4433e73ea783458c4856baab4bcb5

soumith:
cc: @hughperkins

hughperkins:
Hmmm, so, after `luarocks make`ing this branch:
- `clnn.test()` continues to work
- char-rnn continues to train.
- on the other hand, when I run neural-style, with clnn, I realize I get an error:

```
torch/install/share/lua/5.1/nn/SpatialConvolution.lua:100: attempt to call field 'SpatialConvolutionMM_updateOutput' (a nil value)
stack traceback:
	...tu/torch/install/share/lua/5.1/nn/SpatialConvolution.lua:100: in function 'updateOutput'
	/home/ubuntu/torch/install/share/lua/5.1/nn/Sequential.lua:44: in function 'forward'
	neural_style.lua:204: in function 'main'
	neural_style.lua:500: in main chunk
	[C]: in function 'dofile'
	...untu/torch/install/lib/luarocks/rocks/trepl/scm-1/bin/th:131: in main chunk
```
Unfortunately it's 1am on Monday morning, and I probably wont be able to address this until next weekend.

hughperkins:
Ok, I've fixed it, in branch, https://github.com/hughperkins/clnn/tree/thnn3  It looks like I cannot merge until the pull request is merged, since the changes are backwards-incompatible.  Can we arrange a day/time for the merge in advance, maybe sometime over a weekend?

soumith:
@hughperkins merged now.

hughperkins:
:-P  Ok, fine :-P

PR #552, opened 2016-01-06
Add adjustment term to SpatialFullConvolution to control output size

fbesse:
There are multiple valid output sizes when using stride that are greater than 1, since there are multiple input sizes that result in the same output size for the regular SpatialConvolution. This term allows the user to control the size of the output by expanding it, up to a value of (stride-1) in each dimension.

### DIFF
--- a/SpatialFullConvolution.lua
+++ b/SpatialFullConvolution.lua
@@ -1,7 +1,7 @@
 local SpatialFullConvolution, parent = torch.class('nn.SpatialFullConvolution','nn.Module')
 
 function SpatialFullConvolution:__init(nInputPlane, nOutputPlane,
-                                       kW, kH, dW, dH, padW, padH)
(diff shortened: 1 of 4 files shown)


soumith:
Looks good to me, wanted to do it myself.
Two additional things. Can you document this in doc/convolution.md and, can we have this for cunn as well...

fbesse:
Updated the docs. Will add the same functionality to cunn shortly.

Travis seems to have problems fetching files from github though:
Error: Failed installing dependency: https://raw.githubusercontent.com/torch/rocks/master/luaffi-scm-1.rockspec - Error fetching file: Failed downloading https://raw.githubusercontent.com/torch/rocks/master/luaffi-scm-1.rockspec - luaffi-scm-1.rockspec

soumith:
Thanks a lot!

PR #553, opened 2016-01-07
self.train defaults explicitly to 'true'

hughperkins:
self.train defaults explicitly to 'true'

(means we can avoid having to deal with the possibility that it might be `nil`, eg see https://github.com/Element-Research/rnn/blob/master/Recursor.lua#L22 )

### DIFF
--- a/Module.lua
+++ b/Module.lua
@@ -3,6 +3,7 @@ local Module = torch.class('nn.Module')
 function Module:__init()
    self.gradInput = torch.Tensor()
    self.output = torch.Tensor()
+   self.train = true
 end
 
 function Module:parameters()


nicholas-leonard:
:+1:

fmassa:
To keep backward compatibility with old code, shouldn't it default to `false` instead ?

hughperkins:
> To keep backward compatibility with old code, shouldn't it default to false instead ?

Well... which modules are using it?  I know the following:
- Dropout.lua => explicitly assigns training to true, https://github.com/torch/nn/blob/master/Dropout.lua#L6
- Recursor.lua => treats `nil` as `true` https://github.com/Element-Research/rnn/blob/master/Recursor.lua#L22

hughperkins:
```
git/nn$ grep 'self.train' *.lua
BatchNormalization.lua:   self.train = true
BatchNormalization.lua:   if self.train == false then
BatchNormalization.lua:   assert(self.train == true, 'should be in training mode when self.train is true')
Dropout.lua:   self.train = true
Dropout.lua:   if self.train then
Dropout.lua:   if self.train then
Module.lua:   self.train = true
Module.lua:   self.train = false
RReLU.lua:   self.train = true
SpatialBatchNormalization.lua:   self.train = true
SpatialBatchNormalization.lua:      self.train,
SpatialBatchNormalization.lua:   assert(self.train == true, 'should be in training mode when self.train is true')
SpatialDropout.lua:   self.train = true
SpatialDropout.lua:   if self.train then
SpatialDropout.lua:   if self.train then
```

Going through these:
- BatchNormalization.lua => assigns `self.train = true` in `__init`
- SpatialDropout.lua => assigns `self.train = true` in `__init`
- SpatialBatchNormalization.lua => assigns `self.train = true` in `__init`
- RReLU.lua => assigns `self.train = true` in `__init`

fmassa:
Indeed, it should be `true` by default. My bad :-)

szagoruyko:
why is this needed?

soumith:
I dont see why this PR is good. To solve something minor in a module downstream, you are introducing a property to all modules ever. I'd vote for this to be unnecessary

szagoruyko:
ah sorry, missed the nil section. I think this field is redundant and only the modules that need it should have it. Also it might break self-written modules.

hughperkins:
> I dont see why this PR is good. To solve something minor in a module downstream, you are introducing a property to all modules ever. I'd vote for this to be unnecessary

I'm in two minds about this.  On the one hand, I kind of agree.  On the other hand, Module already owns `:training()` and `:evaluate()`, which set the value of `self.train`.  Therefore Module is aware of `self.train`, and arguably kind of 'owns' it.

One other possible solution would be to create an intermediate class, such as `Trainable` (or some better name), which contains `self.train`, `self:training()`, and `self:evaluate()`.

hughperkins:
Note that, in the interest of checking how other libraries handle it, for consistency, looks like Caffe stores this in Layer:

https://github.com/BVLC/caffe/blob/master/include/caffe/layer.hpp#L323-L324
```
  /** The phase: TRAIN or TEST */
  Phase phase_;
```

DeepCL too: https://github.com/hughperkins/DeepCL/blob/master/src/layer/Layer.h#L33
```
 bool training;
```

mxnet too :-)
https://github.com/dmlc/mxnet/blob/master/include/mxnet/operator.h#L46
```

struct OpContext {
  /*! \brief whether it is training phase */
  int is_train;
```

soumith:
I've discussed this offline, this seems unnecessary.

PR #554, opened 2016-01-07
Dropout optimized for p == 0

ivpopov:
Dropout optimized for p == 0 not to perform any operations

### DIFF
--- a/Dropout.lua
+++ b/Dropout.lua
@@ -14,34 +14,42 @@ function Dropout:__init(p,v1,inplace)
 end
 
 function Dropout:updateOutput(input)
-   if self.inplace then 
-      self.output = input
-   else
-      self.output:resizeAs(input):copy(input)
-   end
-   if self.train then
-      self.noise:resizeAs(input)
-      self.noise:bernoulli(1-self.p)
-      if self.v2 then
-         self.noise:div(1-self.p)
+   if self.p > 0 then
+      if self.inplace then
+          self.output = input
+      else
+          self.output:resizeAs(input):copy(input)
       end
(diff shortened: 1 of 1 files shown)


Dropout.lua
```diff
@@ -14,34 +14,42 @@ function Dropout:__init(p,v1,inplace)
...
+          self.noise:bernoulli(1-self.p)
+          if self.v2 then
+             self.noise:div(1-self.p)
+          end
+          self.output:cmul(self.noise)
+      elseif not self.v2 then
+          self.output:mul(1-self.p)
+      end
+   else
+      self.output = input
```
soumith: this does not respect in-place / out-of-place.
Instead, remove the if condition in line 17, remove this else condition and add:
if self.p >= 0 then
... line 18 to 22
end

Dropout.lua
```diff
@@ -14,34 +14,42 @@ function Dropout:__init(p,v1,inplace)
...
+         end
+         self.gradInput:cmul(self.noise) -- simply mask the gradients with the noise vector
       else
-         self.gradInput:resizeAs(gradOutput):copy(gradOutput)
+         error('backprop only defined while training')
       end
-      self.gradInput:cmul(self.noise) -- simply mask the gradients with the noise vector
    else
-      error('backprop only defined while training')
+      self.gradInput = gradOutput
```
soumith: same comments as in updateOutput. Doesn't respect in-place / out-of-place. instead reuse the logic in line 42-46

ivpopov:
Changed

soumith:
Thanks Ivan.

PR #555, opened 2016-01-07
Fix docs for BCECriterion and DistKLDivCriterion

avrahamruderman:
The current docs suggest a sum is computed, but in the code, an average is computed by default. In addition, the order of `o` and `t` for BCE does not match the code.

In addition, added documentation of default behavior of criteria with respect to averaging vs summing of losses.


soumith:
Thanks Avi!

Issue #557, opened 2016-01-08
Error in ParallelCriterion


elbamos:
With a batch size greater than one and ParallelCriterion backed by MarginRankingCriterion, MSECriterion, and CrossEntropyCriterion:
```
/usr/local/share/lua/5.1/nn/ParallelCriterion.lua:23: internal error in __mul: no metatable
stack traceback:
	[C]: in function '__mul'
	/usr/local/share/lua/5.1/nn/ParallelCriterion.lua:23: in function 'forward'
	[string "loss = all_loss:forward({{predictions[1], pre..."]:1: in main chunk
``` 

I believe what's happening is that one of the criteria wants to report loss as a vector while the others want to report is as a scalar.

Running with a batch size of one:
```
cuda runtime error (77) : an illegal memory access was encountered at /tmp/luarocks_cutorch-scm-1-5840/cutorch/lib/THC/generic/THCStorage.c:30
stack traceback:
	[C]: at 0x0856aca0
	[C]: in function '__index'
	/usr/local/share/lua/5.1/nn/ClassNLLCriterion.lua:49: in function 'updateOutput'
	/usr/local/share/lua/5.1/nn/CrossEntropyCriterion.lua:13: in function 'updateOutput'
	/usr/local/share/lua/5.1/nn/ParallelCriterion.lua:23: in function 'forward'
```

fmassa:
Indeed, MarginRankinhCriterion outputs a tensor in the batch case. This should be fixed

fmassa:
I'll prepare a PR for it later on today

PR #558, opened 2016-01-08
Fix batch mode in MarginRankingCriterion

fmassa:
Fixes https://github.com/torch/nn/issues/557
It seems that the batch version of this module was buggy since 2013 !


soumith:
Hehe the classic modules....

Issue #548, opened 2015-12-31
SoftMax and LogSoftMax behave differently with multidimensional input


dpfau:
SoftMax will treat 4D input as a minibatch and sum along the second dimension, while LogSoftMax will throw an error, claiming it was expecting a vector or matrix. It seems like the two should behave similarly.

fmassa:
They used to behave similarly before commit https://github.com/torch/nn/pull/370 , which extended SoftMax to work in the spatial domain.
LogSoftMax could be extended similarly as well, but no one needed it.

fmassa:
I was wrong when I said that no one needed SpatialLogSoftMax. It's present in cudnn actually.
I'll prepare a PR extending LogSoftMax to work in the spatial domain

PR #560, opened 2016-01-09
Update LogSoftMax to work in spatial domain

fmassa:
Extends `LogSoftMax` to work with 3D and 4D tensors.

I did the same thing as in https://github.com/torch/nn/pull/370 , so currently both `LogSoftMax` and `SpatialLogSoftMax` shares the same C code and are by all means equivalent. I wonder if we should only have a `LogSoftMax` which handles 1D/2D/3D/4D cases (which is already the case btw).
I was going to add some asserts to limit the use cases of each function, but then I saw that `cudnn` `SpatialLogSoftMax` [also supports 1D/2D tensors](https://github.com/soumith/cudnn.torch/blob/master/SpatialSoftMax.lua#L18-L28), so I left it like that.

I also slightly increased the tolerance `expprecision` in the tests, as they were sometimes failing.

Addresses the remarks presented https://github.com/torch/nn/issues/548.

I'll prepare a CUDA version later.

### DIFF
--- a/SpatialLogSoftMax.lua
+++ b/SpatialLogSoftMax.lua
@@ -1,0 +1,19 @@
+local SpatialLogSoftMax = torch.class('nn.SpatialLogSoftMax', 'nn.Module')
+
+function SpatialLogSoftMax:updateOutput(input)
+   input.THNN.LogSoftMax_updateOutput(
+      input:cdata(),
+      self.output:cdata()
+   )
+   return self.output
+end
+
+function SpatialLogSoftMax:updateGradInput(input, gradOutput)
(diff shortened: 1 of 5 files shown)


lioutasb:
Can you please sir fix the issues? Because still LogSoftMax does not support 3D tensors.

apaszke:
@lioutasb If you're using CUDA maybe `cudnn.SpatialLogSoftMax` can help for now.

fmassa:
I think this was working fine. The CUDA PR is also ready and was working.

soumith:
rebased and merged via https://github.com/torch/nn/commit/eb7656e621347e3d9150ed593279bae63dd004f4

PR #562, opened 2016-01-11
Improve error message in SpatialConvolutionMM

fmassa:
Better error message when number of input channels does not match the expected value


fmassa:
Travis is failing with an unrelated error.
FYI, just noticed that `cunn` already [asserts for the input channels](https://github.com/torch/cunn/blob/master/SpatialConvolutionMM.cu#L36). Shall I make both error messages the same ?

Issue #556, opened 2016-01-07
calling getParameters() twice on same module


nicholas-leonard:
Calling getParameters() twice on the same module should return the same tensors. 
Basically, it would be nice if this test passed : 

```lua
local mlp1 = nn.Linear(3, 4)
local p1, gp1 = mlp1:getParameters()
local p2, gp2 = mlp1:getParameters()
p1:add(1); gp1:add(1)
assert(math.abs(p1:sum() - p2:sum()) < 0.000001)
```

fmassa:
I think this is addressed in https://github.com/torch/nn/pull/563

nicholas-leonard:
@fmassa great can't wait to see this fixed :)

PR #563, opened 2016-01-11
Don't re-flatten parameters if they are already flattened

colesbury:


soumith:
Thanks.

dominikgrewe:
This doesn't work for networks with shared weights, does it? I'm a bit worried about having getParameters() behave differently depending on the network it is called on. People will start relying on getParameters returning the same tensor when calling it twice on an unchanged network, but this isn't always the case.

I think changes like this should really be discussed a bit before we merge them.

soumith:
Sam, discuss it with Dominik and others. Im away for 2 weekss.

dominikgrewe:
That was quick! Thanks Soumith

colesbury:
I think we should make it work for networks with shared weights too. I can submit another PR for that.

Here's my proposal:

In general, I think getParameters should provide some guarantees which we haven't made explicit:
1) weight and gradWeight should align (i.e. a M.weight should should be at the same offset as M.gradWeight in their respective tensors); this is necessary for optim to work properly
2) getParameters should be idempotent if your network structure hasn't changed

I also think we should support shared weights, but not partially overlapping weight tensors. Our partial support for overlapping weight tensors overly complicates the code, and I have not seen a network that makes use of this

Let me know what you think.

cc/ @adamlerer

nicholas-leonard:
@colesbury I am pretty sure it already supports tensors that share storage at the moment, right?

colesbury:
I meant that we should stop trying to support partially overlapping tensors. i.e we should *not* try to support something like the below code in which the weights are interleaved:

```lua
local w = torch.Tensor(100, 2)
module1.weight = w:select(2, 1)
module2.weight = w:select(2, 2)
```

It complicates the code and makes it harder to achieve more important goals like (1). (If people actually have networks like that then I guess we should support it, but I don't think that's the case.)

nicholas-leonard:
@colesbury Yeah that kind of overlapping is used sometimes. I know I have used it to share an input with an output layer in a language model where the first row of the lookuptable is for dummy words : 
```lua
Linear.weight:set(LookupTable.weight:narrow(2,LookupTable.weight.size(1)))
```
I think we should support it, even though the use-case is indeed rare.
If we choose not to support it, we should at least detect the issue and then call `error` to let the user know. Up to @soumith to decide.

dominikgrewe:
How about we leave `getParameters` as it is but deprecate it and replace it with something saner? This way we don't break old code in unexpected ways.

Here's an idea that came up in a discussion we had:

We provide two methods, say `flattenParameters` and `getFlattenedParameters`. Each time `flattenParameters` is called, it creates a new flattened tensor of the module's parameters and saves it alongside the module. A call to `getFlattenedParameters` returns a reference to that tensor (and raises an error if `flattenParameters` hasn't been called yet).

`flattenParameters` could also invalidate a previously created tensor, e.g. by setting it to an empty tensor. This would prevent you from accidentally using out-of-date parameters.

With this approach, the responsibility of creating a new flattened parameter tensor has been shifted to the user. I think that's more transparent and the user typically knows when the network has changed.

Any thoughts?

nicholas-leonard:
@dominikgrewe I think the flattenParameters() idea is great. But couldn't we just change `getParameters()` so that it does what your proposed `getFlattenedParameters()` method would do? Actually, a call to `getParameters([overwrite])` should automatically call `flattenParameters()` if the module doesn't have the `flatParam, flatGradParam` attributes set by `flattenParameters()` and call it in any case when optional argument to `overwrite=true`. Default value for `overwrite` argument would be `false`. 

I also like the idea that it would invalidate the previous flattened tensors.

colesbury:
@dominikgrewe @nicholas-leonard 
I'm going to submit another PR for this issue (#556) with support for shared tensors. I think Dominik's idea of separating the flattening from the getting of the parameters (and the better name) makes sense, but that can be done independently of this issue.

Avoiding re-flattening is particularly important because it often breaks the *current* `DataParallelTable` (the tests are currently failing).

dominikgrewe:
I'm just a bit worried about changing the semantics of such as fundamental function as `getParameters`. There might be plenty of code out there relying on the fact that it always returns a new tensor. Hence the suggestion to leave `getParameters` as-is and add better-named functions.

colesbury:
I think this is worth that risk. There is code out there that breaks because of the current behavior. As I mentioned before, nn.DataParallelTable is currently broken.

koraykv:
I don't think one class that breaks sometimes is enough to justify breaking backwards compatibility and risking many hidden bugs. Let's be a little more careful when we make changes to functions that are used by every training script out there and has been around for more than 3 years with accepted behaviour and interface.

nicholas-leonard:
I am curious, has anyone ever used the functionality where a new tensor/storage is created every time? I don't see how any one would actually need that functionality. 

Yes, yes, it is possible that someone does indeed rely on this behaviour. I am just curious to know why anyone would call getParameters() twice on purpose to delete the old storage...

adamlerer:
@koraykv I typically have a high bar for API-breaking changes, and tend to subscribe to the view "you have to live with your mistakes".

However, I think that :getParameters() really needs to be fixed. Besides suffering from really bad naming (it sounds like a const function but in fact it's a dangerously-non-const function), it's been buggy forever and had multiple (behavior-modifying) attempts to fix up the bugs:

https://github.com/torch/nn/commit/7105419c8382c3a9c5c257661c8ab49872fdcddc
https://github.com/torch/nn/commit/8581f84f45c2baef72dd32f6336410480b617ff7
https://github.com/torch/nn/commit/0511421f2df3d2a015986de292b4e39af65ea298
etc.

none of which (including mine) seems to have resolved all the bugs. Unfortunately, the implicit spec of `:getParameters()` -- compact all the tensors' data into a single contiguous storage, while preserving sharing, etc. -- is impossible in certain cases, which is part of the problem. 

To the extent that we can write down an explicit spec and write a correct implementation of that spec, I think it's worth breaking certain undocumented assumptions that may be baked into scripts.

@soumith ?

adamlerer:
FYI we do internally suffer from a number of problems/mistakes related to the behavior of `getParameters`.

* we had a bug where a user would delete the `gradWeight` field from their model before saving, then recreate these tensors and call `getParameters` again. This is a silent bug that causes your model to not train correctly.
* users often call `getParameters` not realizing it's non-idempotent, and end up breaking their models in ways they don't understand, e.g.
```
local nParams = model:getParameters():nElement()
```

koraykv:
I agree that its behavior is bad and naming as well, so creating a new function that behaves and is named properly sounds like a good idea to me.

colesbury:
Do we have any deprecation strategy? (Internally, I can just replace all uses of getParameters with the new function)

The semantic change seems very unlikely to break existing code. The the cost of adding a new function with slightly different behavior may be worse than changing the existing implementation. I'm worried that people will still use the existing broken function because it mostly does what they want, but breaks in unexpected ways. (Such as when retraining a model with getParameters)

soumith:
Okay fine, let's do this:
- create a function called flattenParameters() that does what @colesbury proposes.
- add a big WARNING to getParamters stating the depreceation. We can publish a depreceation date of July 2016.
- In August 2016, we move getParameters to silently call flattenParameters.

Internally, immediately move all getParameters to flattenParameters.

nicholas-leonard:
@soumith :+1:

jmrf:
Any progress with this issue?

mbchang:
Checking to see if there is any progress with this issue. Support for partially overlapping tensors would also be useful for [https://github.com/torch/torch7/issues/655#issuecomment-212919156](url)

PR #564, opened 2016-01-12
In-place option for CAddTable

colesbury:


soumith:
add doc to doc/XXX.md

PR #569, opened 2016-01-16
Unpooling with stored Indices for Volumetric Max Pool

kmul00:
Added `Unpooling` option to **VolumetricMaxPooling**. Works with the current `VolumetricMaxPooling` module.
This is the `CPU` version. I will update the `CUDA` version soon.

```lua
    require 'nn'

    max_module = nn.VoulmetricMaxPooling(kT, kW, kH)
    
    model = nn.Sequential()
    model:add(max_module)
    model:add(nn.VolumetricMaxUnpooling(max_module))
```
It requires `kT==dT`, `kW==dW`, `kH==dH`.

Updated `init.lua`, `init.c` and `test.lua`.

### DIFF
--- a/generic/VolumetricMaxUnpooling.c
+++ b/generic/VolumetricMaxUnpooling.c
@@ -1,0 +1,274 @@
+#ifndef TH_GENERIC_FILE
+#define TH_GENERIC_FILE "generic/VolumetricMaxUnpooling.c"
+#else
+
+static void nn_(VolumetricMaxUnpooling_updateOutput_frame)(real *input_p, real *output_p,
+                                                      real *ind_p,
+                                                      long nslices,
+                                                      long itime, long iwidth, long iheight,
+                                                      long otime, long owidth, long oheight,
+                                                      int dT, int dW, int dH)
+{
+  long k;
+#pragma omp parallel for private(k)
+  for (k = 0; k < nslices; k++)
+  {    
+    long ti, i, j, maxz, maxy, maxx;
+    for(ti = 0; ti < itime; ti++)
+    {
+      for(i = 0; i < iheight; i++)
+      {
+        for(j = 0; j < iwidth; j++)
+        {
+          real *output_p_k = output_p + k*otime*owidth*oheight + ti*owidth*oheight*dT + i*owidth*dH + j*dW;
+          real *input_p_k = input_p + k*itime*iwidth*iheight + ti*iwidth*iheight + i*iwidth + j;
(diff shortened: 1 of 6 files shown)


soumith:
@kmul00 can you add documentation in doc/

kmul00:
@soumith Done !

generic/VolumetricMaxUnpooling.c
```diff
@@ -0,0 +1,272 @@
...
+                                                indices_data+p*nslices*itime*iwidth*iheight,
+                                                nslices,
+                                                itime, iwidth, iheight,
+                                                otime, owidth, oheight,
+                                                dT, dW, dH);
+    }
+  }
+
+  /* cleanup */
+  THTensor_(free)(input);
```
soumith: memory leak here. you need to free indices too.

generic/VolumetricMaxUnpooling.c
```diff
@@ -0,0 +1,272 @@
...
+                                                   indices_data+p*nslices*itime*iwidth*iheight,
+                                                   nslices,
+                                                   itime, iwidth, iheight,
+                                                   otime, owidth, oheight,
+                                                   dT, dW, dH);
+    }
+  }
+
+  /* cleanup */
+  THTensor_(free)(gradOutput);
```
soumith: another memory leak, you need to free indices

generic/VolumetricMaxUnpooling.c
fmassa: Indeed there is a leak. The leak is also present in the spatial version of this module, which should be fixed. cc @clement-masson 
There is no "real" leak in fact because `indices` is contiguous and kept forever.

kmul00:
Something went wrong with the travis check. One of the Jobs gave the error `No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself` and stopped terminated.

This happened while cloning into `extra/nn`.

Details [here](https://travis-ci.org/torch/nn/jobs/103652072)

kmul00:
All builds are successful finally. Might have been some problem on Travis side. I just closed and reopened the PR and the builds started once again, and this time all of them went through.

soumith:
Thanks, looks good. Can you squash your commits into 1 commit (for example using git rebase -i ), I'll merge after that.

kmul00:
Yeah done ! :)

soumith:
Thanks :)

PR #575, opened 2016-01-21
Supplementary pull request for argmax/argmin as second argument on max/min calls in torch7

mevGDM:
This makes a minor change to MultiLabelMarginCriterion.c to adhere to the new interface for minall/maxall functions which are changed in the pull request for argmax/argmin as second argument on max/min calls. This will not build until the https://github.com/torch/torch7/pull/489 pull is finished.


soumith:
fixed via https://github.com/torch/nn/commit/b2af7eaddfc1de72661b6861115d3fdb97403bf3

dominikgrewe:
@soumith Can you revert the change? See torch/torch7#523

soumith:
sure

PR #576, opened 2016-01-21
read and write methods for nn.Module.

dominikgrewe:
Allows modules that want to implement custom `read` and `write` methods to
call `parent.read/write`.

### DIFF
--- a/Module.lua
+++ b/Module.lua
@@ -141,6 +141,22 @@ end
 function Module:reset()
 end
 
+function Module:write(file)
+  -- Write all values in the object as a table.
+  local object = {}
+  for k, v in pairs(self) do
+    object[k] = v
+  end
+  file:writeObject(object)
+end
+
+function Module:read(file)
+  local object = file:readObject()
+  for k, v in pairs(object) do
+    self[k] = v
+  end
+end
+
 -- This function is not easy to understand. It works as follows:
 --
 -- - gather all parameter tensors for this module (and children);


nicholas-leonard:
@dominikgrewe This won't break anything, but now all modules will be serialized through those methods you just created.

dominikgrewe:
Yes, but they're just doing what torch serialization would have done anyway (apart from filtering out non-writable objects).

soumith:
looks pretty useful! Thanks.

soumith:
@nicholas-leonard this is useful when implementing custom read-write, you can change some fields and then just call parent.read /write

eladhoffer:
@soumith can't we use this to avoid the need for #526 by adding an option to write just weight, bias and running_mean/var (empty tensors for all other fields)?

dominikgrewe:
There are various other things you might want to save though, e.g. the convolution settings (kernel size, stride, padding etc) for a convolution module. So we could only do without `clearState` if many modules had custom read & write functions. It's worth thinking about though. What other uses for `clearState` are there aside from serialization?

eladhoffer:
@dominikgrewe yes, what I meant is removing the output, gradInput and buffer tensors. Other fields should be kept.

soumith:
@eladhoffer not a good idea. if the intention is to save a model, it's to save the whole model as is. we can introduce an nn.save that calls clearState before hand, and that's okay.

PR #581, opened 2016-01-25
Revert "In-place option for CAddTable"

soumith:
Reverts torch/nn#564


Wakeupbuddy:
Why this one get reverted?

soumith:
it had a bug. Sam is sending out a fixed version.

PR #582, opened 2016-01-25
adding a Bilinear layer

soumith:
Applies a bilinear transformation to the incoming data, i.e. `\forall k: y_k = x_1 A_k x_2 + b`. The `input` tensor given in `forward(input)` is a table containing both inputs `x_1` and `x_2`, which are tensors of size `N x inputDimension1` and `N x inputDimension1`, respectively. The layer can be trained without biases by setting `bias = false`.

Implemented by @lvdmaaten

### DIFF
--- a/Bilinear.lua
+++ b/Bilinear.lua
@@ -1,0 +1,140 @@
+local Bilinear, parent = torch.class('nn.Bilinear', 'nn.Module')
+
+local function isint(x) return type(x) == 'number' and x == math.floor(x) end
+function Bilinear:__assertInput(input)
+   assert(input and type(input) == 'table' and #input == 2,
+      'input should be a table containing two data Tensors')
+   assert(input[1]:nDimension() == 2 and input[2]:nDimension() == 2,
+      'input Tensors should be two-dimensional')
+   assert(input[1]:size(1) == input[2]:size(1),
+      'input Tensors should have the same number of rows (instances)')
+   assert(input[1]:size(2) == self.weight:size(2),
+      'dimensionality of first input is erroneous')
+   assert(input[2]:size(2) == self.weight:size(3),
+      'dimensionality of second input is erroneous')
+end
+function Bilinear:__assertInputGradOutput(input, gradOutput)
+   assert(input[1]:size(1) == gradOutput:size(1),
+      'number of rows in gradOutput does not match input')
+   assert(gradOutput:size(2) == self.weight:size(1),
+      'number of columns in gradOutput does not output size of layer')
+end
+
+function Bilinear:__init(inputSize1, inputSize2, outputSize, bias)
+
+   -- assertions:
+   assert(self and inputSize1 and inputSize2 and outputSize,
+      'should specify inputSize1 and inputSize2 and outputSize')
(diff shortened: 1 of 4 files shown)


paidi:
:+1:

Bilinear.lua
```diff
@@ -0,0 +1,140 @@
...
+
+   -- compute output scores:
+   self.output:resize(input[1]:size(1), self.weight:size(1))
+   for k = 1,self.weight:size(1) do
+      torch.mm(self.buff, input[1], self.weight[k])
+      self.buff:cmul(input[2])
+      torch.sum(self.output:narrow(2, k, 1), self.buff, 2)
+   end
+   self.output:add(
+      self.bias:reshape(1, self.bias:nElement()):expandAs(self.output)
```
fmassa: this should be inside a condition checking the existence of `bias`, right ?

lvdmaaten: You're right, good catch!

soumith: @lvdmaaten are you fixing it, or should I?

lvdmaaten:
It's probably easier if you do it.

Bilinear.lua
lvdmaaten: It's probably easier if you do it.

soumith: fixed in trunk

lvdmaaten: Thanks Soumith!

doc/simple.md
```diff
@@ -119,6 +120,28 @@ x = torch.Tensor({ {1, 0.1}, {2, 0.3}, {10, 0.3}, {31, 0.2} })
...
 
+<a name="nn.Bilinear"></a>
+## Bilinear ##
+
+```lua
+module = nn.Bilinear(inputDimension1, inputDimension2, outputDimension, [bias = true])
+```
+
+Applies a bilinear transformation to the incoming data, i.e. `\forall k: y_k = x_1 A_k x_2 + b`. The `input` tensor given in `forward(input)` is a table containing both inputs `x_1` and `x_2`, which are tensors of size `N x inputDimension1`
+and `N x inputDimension1`, respectively. The layer can be trained without biases by setting `bias = false`.
```
dasguptar: @soumith 
Shouldn't the dimension of `x_2` be `N x inputDimension2`?

lvdmaaten: Yep, you're right!

doc/simple.md
```diff
@@ -119,6 +120,28 @@ x = torch.Tensor({ {1, 0.1}, {2, 0.3}, {10, 0.3}, {31, 0.2} })
 
 The first column contains indices, the second column contains values in a a vector where all other elements are zeros. The indices should not exceed the stated dimensions of the input to the layer (10000 in the example).
 
+<a name="nn.Bilinear"></a>
+## Bilinear ##
+
```
nicholas-leonard: Could you add a reference to a relevant paper or web page?

lvdmaaten: I have no idea who first described bilinear maps... probably Leibniz or Gauss or something?!

https://en.wikipedia.org/wiki/Bilinear_map

PR #583, opened 2016-01-25
add option to nn.Linear to have no bias

soumith:
Implemented by @lvdmaaten

### DIFF
--- a/Linear.lua
+++ b/Linear.lua
@@ -1,13 +1,14 @@
 local Linear, parent = torch.class('nn.Linear', 'nn.Module')
 
-function Linear:__init(inputSize, outputSize)
+function Linear:__init(inputSize, outputSize, bias)
    parent.__init(self)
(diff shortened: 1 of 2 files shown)


Linear.lua
```diff
@@ -22,25 +23,28 @@ function Linear:reset(stdv)
...
-
    return self
 end
 
 function Linear:updateOutput(input)
    if input:dim() == 1 then
-      self.output:resize(self.bias:size(1))
-      self.output:copy(self.bias)
+      self.output:resize(self.weight:size(1))
+      if self.bias then self.output:copy(self.bias) else self.output:fill(0) end
```
fmassa: Just wondering, isn't `zero()` faster than `fill(0)` for contiguous tensors in GPU ?

soumith: you're right. will fix.

PR #585, opened 2016-01-25
added ceil mode to VolumetricMaxPooling

soumith:


fmassa:
Saved models won't have the field `ceil_mode`. Shouldn't we add the back compatibility checks in forward ?

soumith:
You're right, i'll fix this in a PR.

PR #589, opened 2016-01-27
Volumteric Max pooling now supports padding

kmul00:
This allows to provide **padding** in `VolumetricMaxPooling`

modified:   VolumetricMaxPooling.lua
modified:   generic/VolumetricMaxPooling.c
modified:   test.lua

**NOTE:** `VolumetricMaxUnpooling` still doesn't support **padded pooling**, though it will work for *pooling module without padding*. I will submit a *patch* for it soon.


soumith:
Thanks Koustav!

PR #590, opened 2016-01-28
Fixed SelectTable bug and added unit test

sahiliitm:
The bug can be created by running the updated unit test or also by running [rc1.lua here](https://github.com/sahiliitm/read-and-comprehend) (requires the rnn package)


soumith:
Thanks Sahil.

nicholas-leonard:
:+1:

PR #591, opened 2016-01-28
Updated VolumetricMaxUnpooling to support unpooling of padded pooling

kmul00:
Patch for `VolumetricMaxUnpooling` to support *padded pooling* introduced [here](https://github.com/torch/nn/pull/589).
Also, updated **README** to reflect `VolumetricMaxPooling` supports *padding*.

modified:   VolumetricMaxUnpooling.lua
modified:   doc/convolution.md
modified:   generic/VolumetricMaxUnpooling.c
modified:   test.lua


soumith:
good stuff! thanks.

PR #596, opened 2016-01-31
add PowConstant

minhlab:
Add a layer that raise its input to a fixed constant. Used in Stanford neural dependency parsing experiments (Chen & Manning, 2014). Sorry I'm not competent to write unit test for the module.

Chen, D., & Manning, C. (2014). A Fast and Accurate Dependency Parser using Neural Networks. In Proceedings of the 2014 Conference on Empirical Methods in Natural Language Processing (EMNLP) (pp. 740–750). Doha, Qatar: Association for Computational Linguistics.


fmassa:
This module is equivalent to [Power](https://github.com/torch/nn/blob/master/Power.lua), right ?

minhlab:
@fmassa : exactly. I thought it must be next to AddConstant and [MulConstant](https://github.com/torch/nn/blob/master/doc/transfer.md#mulconstant) so I only looked there. It's also a bit strange that Power is not learnable while Mul is learnable.

fmassa:
So, I'm not sure we want inplace computations for `Power`, because:
* for `p` even, the input can't be reversed correctly if it's negative and the gradients won't be computed properly
* computing `a^((p-1)/p)`, for `p` odd leads to `nan` if `a` is negative.

I think there is no `Power` module with learnable parameter because it could probably lead to unstable gradients, but that's just a guess.

minhlab:
Hmmm... I'm not sure either. We can close this PR.

PR #599, opened 2016-02-01
CMul: Added support for non-contiguous input and gradOuput

ivpopov:
Currently, if either the input or gradOuput is non-contiguous, this Module will not work. The change will enable the use of Module in this scenario by performing the necessary tensor copies.

### DIFF
--- a/utils.lua
+++ b/utils.lua
@@ -150,4 +150,16 @@ function nn.utils.addSingletonDimension(t, dim)
   return view
 end
 
+function nn.utils.contiguousView(output, input, ...)
+  output = output or input.new()
+  if input:isContiguous() then
+    output:view(input, ...)
+  else
+    output:resizeAs(input)
+    output:copy(input)
+    output:view(...)
+  end
+  return output
+end
+
 table.unpack = table.unpack or unpack
--- a/CMul.lua
+++ b/CMul.lua
@@ -77,8 +77,8 @@ function CMul:updateGradInput(input, gradOutput)
       self.gradInput:addcmul(1, self.weight, gradOutput)
    else
       local batchSize = input:size(1)
-      self._gradOutput:view(gradOutput, batchSize, -1)
-      self._gradInput:view(self.gradInput, batchSize, -1)
+      nn.utils.contiguousView(self._gradOutput, gradOutput, batchSize, -1)
+      nn.utils.contiguousView(self._gradInput, self.gradInput, batchSize, -1)
       self._weight:view(self.weight, 1, -1)
       self._expand:expandAs(self._weight, self._gradOutput)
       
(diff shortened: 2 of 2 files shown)


soumith:
nice. thanks.!

utils.lua
```diff
@@ -150,4 +150,16 @@ function nn.utils.addSingletonDimension(t, dim)
...
 end
 
+function nn.utils.contiguousView(output, input, ...)
+  output = output or input.new()
+  if input:isContiguous() then
+    output:view(input, ...)
+  else
+    output:resizeAs(input)
+    output:copy(input)
+    output:view(...)
```
fmassa: This doesn't update `output` to have the shape specified by `...`.
Shouldn't it be something like `output:view(input, ...)` ?

fmassa: I meant instead
```lua
output:view(output,...)
```

ivpopov: If you put this on line 160 you will reintroduce the original error. See this fragment from .view()

if istensor(result) and istensor(src) and type(size) == 'number' then
      size = torch.LongStorage{...}
      view = result
      tensor = src
   elseif istensor(result) and istensor(src) and isstorage(size) then
      size = size
      view = result
      tensor = src
   elseif istensor(result) and isstorage(src) and size == nil then
      size = src
      tensor = result
      view = tensor.new()
   elseif istensor(result) and type(src) == 'number' then
      size = {...}
      table.insert(size,1,src)
      size = torch.LongStorage(size)
      tensor = result
      view = tensor.new()

ivpopov: Sorry, just noticed that you want to put output as src in your second suggestion. I think this output:view(output,...) 
should not be necessary and is a tiny bit less readable.

fmassa: Check my second comment.
I believe you are not assigning to `output` the shape indicated from `...` (if only one tensor is provided as argument, the viewed tensor is returned and the original tensor is kept intact)

ivpopov: output:view(...) will enter the following if: 
elseif istensor(result) and type(src) == 'number' then
...
view = tensor.new()
...
view:set(tensor:storage(), tensor:storageOffset(), size)

Why do you think this is not what we need?

fmassa: Because `view` is allocated in [here](https://github.com/torch/torch7/blob/master/Tensor.lua#L446), creating a new tensor, and afterwards it's returned [here](https://github.com/torch/torch7/blob/master/Tensor.lua#L465). The code doesn't change `result`, which is what you pass to the function. What you want instead is to call [this](https://github.com/torch/torch7/blob/master/Tensor.lua#L431), because it will effectively change the passed tensor.

Here is a snippet reproducing the problem
```lua
a = torch.rand(3,3)
a:view(-1);
print(a) -- a is still 3x3
```

ivpopov: You are right. See new PR added.
https://github.com/torch/nn/pull/600

PR #600, opened 2016-02-01
Fix for CMul: Added support for non-contiguous

ivpopov:


fmassa:
Could you add some unit test on `CMul` checking the non-contiguous case ?

soumith:
yea, adding a test would be nice.

ivpopov:
Yes, will add respective tests.

soumith:
Thanks!

PR #602, opened 2016-02-02
Test nn.CMul for non-contiguous input or gradOuput

ivpopov:


fmassa:
Could you merge this PR with https://github.com/torch/nn/pull/600 ? As #600 is not yet merged, travis won't pass this tests.

PR #608, opened 2016-02-04
refactoring for independent API

j-wilson:


torchAPI.lua
```diff
@@ -0,0 +1,215 @@
...
+
+      if template:match('<library>') then
+        template = template:gsub('<library>', self.config.library)
+      end
+
+      if template:match('$([0-9]+)') then
+        template = template:gsub('$([0-9]+)', '%%%1')
+      end
+
+      forward = forward:gsub(self.config.prefix..self.config.pattern, template)
```
apaszke: Will this ever execute more than one time? It's always using the same search pattern and only changes the replacement string. Won't it just replace everything in the first go?

torchAPI.lua
```diff
@@ -0,0 +1,215 @@
...
+  ---- Expand macros
+  local macros = header['macro'] or header['macros']
+  if macros then
+    local macros = self:_interp_macros(macros)
+    local flag   = false
+
+    -- Apply macros with same interpretation everywhere
+    for old, new in pairs(macros) do
+      if type(new) == 'string' then
+        forward = forward:gsub(old, new)
```
apaszke: It's not exactly how C macros work (this will also replace subwords and string contents), but I'm not sure if it's a problem when parsing only the header files.

torchAPI.lua
```diff
@@ -0,0 +1,215 @@
...
+        end
+      end
+    else
+      -- No alternatives case
+      ffi.cdef(forward)
+    end
+  else
+    -- No macros case
+    ffi.cdef(forward)
+  end
```
apaszke: Is there any reason why there can't be a single `ffi.cdef(forward)` at the end (without all the else clauses). You could return from alternative interpretations early if you don't want to execute this.

torchAPI.lua
```diff
@@ -0,0 +1,215 @@
...
+        flag = true -- Are there additional macros to process?
+      end
+    end
+
+    -- Apply macros with alternative interpretations
+    if flag then
+      for id, group in pairs(macros) do
+        if type(group) == 'table' then
+          local forward = forward
+          for old, new in pairs(group) do
```
apaszke: Won't this break if one specifies both integer and string indexed macros (with multiple interpretations)? I think it's worth checking this somewhere.

apaszke:
I think that as soon as it's available as a separate package it's worth adding this. I like that it allows to use a more declarative style when linking C libraries. I'm just wondering if there's any way not to hard code the definitions, but to load them from appropriate files.

Also, for me, the constraint that all functions have to return the same type is quite strong. This holds for THNN, but I think it'll break with many other libraries. Couldn't it be replaced with a non-greedy capturing group (there can be multiple words e.g. `THTensor *`)?

torchAPI.lua
j-wilson: Good point; I'll work out a patch when I push an independent repo.

torchAPI.lua
j-wilson: Upon returning from `_interp_macros()`, macros with multiple interpretations should be sorted into self-contained groups. Key `id` can be either an integer or a string and serves only as an identifier; but, groups should contain string-string type key-value pairs. I'll add an additional check to enforce this for pre-sorted groups.

torchAPI.lua
j-wilson: Good point; the method you've mentioned would be cleaner.

torchAPI.lua
j-wilson: Can't think of an example off the top of my head; but, one could imagine cases where multiple templates are in use. The `match` statement are only there for use of special patterns, which may or may not be present in a given template.

j-wilson:
Can you elaborate on what you mean by "[a] way not to hard code the definitions, but to load them from appropriate files"?

Agree that all functions having the same return type is to restrictive; and, aim to amend this when I push to an independent repo.

torchAPI.lua
apaszke: Oh, ok. So if I specified 2 groups with integer keys and several macros with two interpretations (identified by string keys - like `d` and `f`), then I'd get four cdefs generated there, right?

j-wilson: That's correct. You should get the original two (integer-keyed) groups plus groups 'd' and 'f'.

apaszke:
I mean that torch's header files are quite similar and that most of THNN_h.lua is THNN.h copied, so there's much code duplication in there. I think that THNN.h could be just read from a file, not hardcoded as a string.

j-wilson:
Hmm, I'll look into that. Sounds like a good idea to me.

torchAPI.lua
apaszke: Ok, I think it's worth clarifying this in a comment or in docs, because it was not not obvious for me whether the correct behavior would be to take a cross product of integer and string macro groups, or should they be just applied separately.

j-wilson: Noted. Will include in docs

apaszke: Thanks :+1:

andreaskoepf:
I think in its current form it does not follow the general style and naming-conventions in torch, e.g. the commenting style is very "j-wilson specific" (e.g. I personally do not like 3-line title comments with '------' lines and 'Authored' tags).  Normally classes in nn all start with an upper case letters (not like 'torchAPI'), and if declared inside a lua file normally the local class-variable has the same name as the class (not 'API' vs. 'torchAPI'), and the normal naming convention is lower-camel-case for functions by you crated functios with names underscore like 'extract_handles' and 'c_init'...

PR #609, opened 2016-02-04
Fix loading of old SpatialBatchNormalization modules

colesbury:
Big oops. I had messed up the backwards compatibility code for SpatialBatchNormalization and my test for this was flawed.

The old-style running_std is actually the E[1/sqrt(var + eps)]. I forgot to subtract out the 'eps' when converting to running_var.


soumith:
thanks!

PR #612, opened 2016-02-05
In-place option for CAddTable

colesbury:


CAddTable.lua
```diff
@@ -1,13 +1,17 @@
...
+function CAddTable:__init(ip)
    parent.__init(self)
+   self.inplace = ip
    self.gradInput = {}
 end
 
 function CAddTable:updateOutput(input)
-   self.output:resizeAs(input[1]):copy(input[1])
+   if self.inplace then
+      self.output = input[1]
```
soumith: can you make this self.output:set(input[1])

CAddTable.lua
```diff
@@ -16,9 +20,12 @@ end
 
 function CAddTable:updateGradInput(input, gradOutput)
    for i=1,#input do
-      self.gradInput[i] = self.gradInput[i] or input[1].new()
-      self.gradInput[i]:resizeAs(input[i])
-      self.gradInput[i]:copy(gradOutput)
+      if self.inplace then
+         self.gradInput[i] = gradOutput
```
soumith: same comment as above.

PR #614, opened 2016-02-05
Add gradient backprop for dropout in evaluation mode

benathi:
Initially, the input gradient is not allowed if the dropout layer is not in training mode. However, there shouldn't be anything preventing back propagation. This is useful if we want deterministic behavior with learning model and want to get the derivative with respect to the low input layer.


soumith:
looks good. thanks!

PR #615, opened 2016-02-05
add backprop for evaluate mode - batch normalization

benathi:
Originally, back propagation is only allowed for train mode. This code add back propagation (update gradInput) for evaluate mode. This is useful, for instance, for computing gradient with respect to input at low layers given a trained model.

### DIFF
--- a/BatchNormalization.lua
+++ b/BatchNormalization.lua
@@ -119,27 +119,40 @@ end
 function BN:updateGradInput(input, gradOutput)
    assert(input:dim() == 2, 'only mini-batch supported')
    assert(gradOutput:dim() == 2, 'only mini-batch supported')
-   assert(self.train == true, 'should be in training mode when self.train is true')
    local nBatch = input:size(1)
 
-   self.gradInput:cmul(self.centered, gradOutput)
-   self.buffer:mean(self.gradInput, 1)
-   self.gradInput:repeatTensor(self.buffer, nBatch, 1)
-   self.gradInput:cmul(self.centered):mul(-1)
-   self.buffer:repeatTensor(self.std, nBatch, 1)
-   self.gradInput:cmul(self.buffer):cmul(self.buffer)
+   if self.train then
+      self.gradInput:cmul(self.centered, gradOutput)
+      self.buffer:mean(self.gradInput, 1)
+      self.gradInput:repeatTensor(self.buffer, nBatch, 1)
+      self.gradInput:cmul(self.centered):mul(-1)
+      self.buffer:repeatTensor(self.std, nBatch, 1)
+      self.gradInput:cmul(self.buffer):cmul(self.buffer)
 
-   self.buffer:mean(gradOutput, 1)
-   self.buffer:repeatTensor(self.buffer, nBatch, 1)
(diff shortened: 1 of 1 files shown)


BatchNormalization.lua
```diff
@@ -119,27 +119,40 @@ end
...
       self.gradInput:cmul(self.buffer)
-   end
 
+      if self.affine then
+         self.buffer:repeatTensor(self.weight, nBatch, 1)
+         self.gradInput:cmul(self.buffer)
+      end
+   else
+      if self.affine then
+         self.buffer:repeatTensor(self.bias, nBatch, 1)
```
jhjin: Do we need to consider bias and running_mean during `updateGradInput`? considering that they are constants.

benathi: For self.train = false, the calculations are done in the opposite order compared to updateOutput. This includes offsetting with bias, weight, running_std, and running_mean. They are constant and are used in the forward pass so they need to be used in the backward pass as well.

Let me know if you think I miss something.

BatchNormalization.lua
```diff
@@ -119,27 +119,40 @@ end
...
+         self.gradInput:add(-1,self.buffer)
```
fmassa: Are you sure about your gradient computation ? The bias shouldn't influence the `gradInput`, because it's a constant added to the input (and thus the gradient of it is zero).
Also, here you are adding `-self.buffer` to `self.gradInput`, which is not initialized, giving garbage.
it would be good to have some unit tests for the newly added cases, to check that everything is working as expected.

benathi:
You're right -- I fixed the code and tested it with unit test with both affine = false and affine = true modes.

Let me know if you need the code for my unit tests. I can paste it to this conversation or if there's a better way let me know.

fmassa:
could you add unit tests in `tests.lua` for the newly added case ?

benathi:
Should I add it to nntest.BatchNormalization() function or create a new one?

fmassa:
Just add a new assert in `nntest.BatchNormalization`, checking forward and backward, with jacobian computation for the gradients wrt the input and the parameters

benathi:
Does my unit test have to use Jacobian? I'm not really sure what testJacobianUpdateParameters, testJacobianParameters, or testJacobian do.

Or is it sufficient to copy the same tests and run in 'evaluate' mode?

soumith:
copy the same tests and run in evaluate mode.

soumith:
cc: @adamlerer who was interested in this.

ssakhavi:
Hi, 
I wanted to use this PR . It comes in handy when you want to visualize the trained network. 

Based on what I read in the source code, the only thing changed here is that the `gradInput` and `gradOutput` are the same during `evaluate()`. Is this a correct ?

soumith:
This is an outdated PR. unfortunately, based on a much older codebase.

soumith:
what you can do is send your input as a mini-batch of size 1.

ssakhavi:
Hi Soumith,

I've tried that actually. Because the dimension is one it requires a
batchsize of larger than 1

PR #616, opened 2016-02-07
fixed dropout backprop in evaluation mode

benathi:
My previous code has a bug. I fixed it and tested with unit test. The computed gradient agrees with finite difference gradient.


soumith:
can you add the unit test to test.lua

PR #617, opened 2016-02-08
Lazy initialization of buffers

szagoruyko:
For backend switching mostly, see also the discussion here https://github.com/torch/nn/pull/526

### DIFF
--- a/SpatialConvolution.lua
+++ b/SpatialConvolution.lua
@@ -155,8 +155,8 @@ function SpatialConvolution:accGradParameters(input, gradOutput, scale)
 end
 
 function SpatialConvolution:type(type,tensorCache)
-   self.finput = torch.Tensor()
-   self.fgradInput = torch.Tensor()
+   self.finput = self.finput and torch.Tensor()
+   self.fgradInput = self.fgradInput and torch.Tensor()
    return parent.type(self,type,tensorCache)
 end
 
--- a/SpatialConvolutionLocal.lua
+++ b/SpatialConvolutionLocal.lua
@@ -26,9 +26,6 @@ function SpatialConvolutionLocal:__init(nInputPlane, nOutputPlane, iW, iH ,kW, k
    self.gradWeight = torch.Tensor():resizeAs(self.weight)
    self.gradBias = torch.Tensor():resizeAs(self.bias)
 
-   self.finput = torch.Tensor()
-   self.fgradInput = torch.Tensor()
-
    self:reset()
 end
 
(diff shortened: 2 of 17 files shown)


SpatialConvolution.lua
```diff
@@ -155,8 +155,8 @@ function SpatialConvolution:accGradParameters(input, gradOutput, scale)
 end
 
 function SpatialConvolution:type(type,tensorCache)
-   self.finput = torch.Tensor()
-   self.fgradInput = torch.Tensor()
+   self.finput = self.finput and self.weight.new()
+   self.fgradInput = self.fgradInput and self.weight.new()
```
fmassa: CPU and GPU version of `SpatialConvolutionMM` uses differently `fgradInput` (GPU assumes that `fgradInput` contains all ones). That's why it was being newly allocating this tensor. I think you should at least resize it to zero. See https://github.com/torch/nn/issues/211

SpatialConvolutionLocal.lua
```diff
@@ -148,8 +147,8 @@ function SpatialConvolutionLocal:accGradParameters(input, gradOutput, scale)
 end
 
 function SpatialConvolutionLocal:type(type,tensorCache)
-   self.finput = torch.Tensor()
-   self.fgradInput = torch.Tensor()
+   self.finput = self.finput and self.weight.new()
+   self.fgradInput = self.fgradInput and self.weight.new()
```
fmassa: Same thing here

SpatialConvolutionMM.lua
```diff
@@ -121,8 +120,8 @@ function SpatialConvolutionMM:accGradParameters(input, gradOutput, scale)
 end
 
 function SpatialConvolutionMM:type(type,tensorCache)
-   self.finput = torch.Tensor()
-   self.fgradInput = torch.Tensor()
+   self.finput = self.finput and self.weight.new()
+   self.fgradInput = self.fgradInput and self.weight.new()
```
fmassa: and here

SpatialFullConvolution.lua
```diff
@@ -92,8 +91,8 @@ function SpatialFullConvolution:accGradParameters(input, gradOutput, scale)
 end
 
 function SpatialFullConvolution:type(type, tensorCache)
-  self.finput = torch.Tensor()
-  self.fgradInput = torch.Tensor()
+  self.finput = self.finput and self.weight.new()
+  self.fgradInput = self.fgradInput and self.weight.new()
```
fmassa: Here it's ok, because the CPU version mimics the GPU

SpatialConvolution.lua
fmassa: I actually think you should do it for `finput` as well (as it was before), because converting an already `forwarded` CPU model to GPU would create a huge (and unnecessary) `finput` buffer.

SpatialFullConvolution.lua
szagoruyko: Thanks Francisco, I brought back the old behaviour.

soumith:
Thanks!

PR #626, opened 2016-02-11
BCECriterion THNN implementation

szagoruyko:
Current implementation is very inefficient. Cuda is coming too.


lib/THNN/generic/BCECriterion.c
```diff
@@ -0,0 +1,48 @@
+#ifndef TH_GENERIC_FILE
+#define TH_GENERIC_FILE "generic/BCECriterion.c"
+#else
+
+#define EPS 1e-12
```
fmassa: Could you undef `EPS` in the end ?

soumith:
Thanks Sergey!

PR #627, opened 2016-02-11
THNN SparseLinear conversion

andreaskoepf:
Cross checked with previous impl via [xcheck.lua](https://github.com/andreaskoepf/nn/blob/nn2/xtest.lua#L397).


soumith:
Thanks.

PR #628, opened 2016-02-12
added VolumetricDropout

misko:


soumith:
fantastic, the only thing I ask for is to add documentation in https://github.com/torch/nn/blob/master/doc/simple.md similar to https://github.com/torch/nn/blob/master/doc/simple.md#nn.SpatialDropout

misko:
I made a typo while merging in the new init.lua, i fixed that and squashed the commit. I can build everything fine and it seems to work fine, I will try to check logs to see where it fails. If anyone has any suggestions that would be great.

soumith:
can you squash your commits?

You might want to use: git rebase -i , followed by git push --force

The PR has 3 commits right now.

misko:
Ok, should be good now?

soumith:
Thank you!

PR #634, opened 2016-02-16
Fix problem with nn.SpatialFractionalMaxPooling:clearState()

sergomezcol:
:initSampleBuffer_() was producing an error when called (from :updateOutput()) after calling :clearState(). The randomSamples field was an empty tensor in that case.

Now :clearState() sets indices and randomSamples to nil, so the module is in the same state as after construction.


soumith:
Thank you!

PR #641, opened 2016-02-18
THNN Conversion of Spatial*ConvolutionMap

andreaskoepf:
- [x] unit tests pass
- [x] cross-check with nn2, see [xtest.lua](https://github.com/andreaskoepf/nn/blob/nn2/xtest.lua#L525)


soumith:
Thank you!

soumith:
there's no generic folder anymore :)

PR #642, opened 2016-02-18
added VolumetricBatchNormalization

misko:
Pretty much copy and pasted the SpatialBatchNormalization with minimal changes for VolumetricBatchNormalization. The only changes that needed to be made to the code in SpatialBatchNormalization is to calculate 'n' based on the number of spatial dimensions. Does it make sense to collapse the C code for spatial/volumetric to the more general case and compute 'n' across all spatial dimensions, and then having the LUA wrappers check for the proper dimension of each call?
I don't know if it makes sense to collapse the LUA wrappers to a more general form, because in that case I don't know how to differentiate between a 2D-batched tensor or a 3D tensor because both would be 4 dimensional.


soumith:
replaced by #665

PR #643, opened 2016-02-18
Lazy init when gradOutput is not contiguous

asrata:
It's possible when the input is contiguous but the gradOutput is not (e.g. when doing nn.Parallel and Reshape is located at the end of each branch). So _gradOutput needs another lazy initialization to avoid exceptions.


soumith:
already fixed via https://github.com/torch/nn/pull/645

PR #644, opened 2016-02-19
Padding in the middle

csbrown:
This updates the Padding module so that padding can be inserted at any index along the specified dimension.  It is totally backwards compatible, so you can still call it just like before, and it will work just like before, preserving contiguity insomuch as the previous version did.

The addition is an extra optional parameter called [index] into which you can specify the index at which to pad.  If padding it negative, index counts from the left, if padding is positive, it counts from the right.

I needed to put some zeros in the middle of a Tensor, and this seemed like the best way to do it.


soumith:
update the documentation with this new option

soumith:
closed the PR by mistake, sorry.

csbrown:
docs updated.

Atcold:
@csbrown, I think you should squash the commits, so that @soumith is going to be happy to merge :grimacing:

csbrown:
rebased.

Atcold:
@csbrown, can you change
`module` = `nn.Padding(dim, pad [, nInputDim, value, index])`
to
```lua
module = nn.Padding(dim, pad [, nInputDim, value, index])
```
The first is the old style, the second is the new one.

csbrown:
k

soumith:
Everything looks good on the code side, but please add test cases to test.lua for the index option.
They provide sanity checks testing the backward gradients against finite difference approximations.

soumith:
as soon as you add the test cases, and they pass, this is good for merge.

csbrown:
Added the index option to the Padding test.  The previous test cases were randomized, so I stuck with that paradigm.  Ran the test 1000 times, and there were no errors, so it is highly unlikely that the module is incredibly broken.

soumith:
Thank you!

csbrown:
First Pull Request = Accepted.  woot woot!  Thanks for your patience, @soumith && @Atcold.

PR #645, opened 2016-02-19
Reshape: Fixed wrong place of _gradOutput initialization

mys007:
Last commit broke the case when input is continuous and gradOutput is not. This fixes it.


soumith:
thanks!

PR #649, opened 2016-02-19
Override Module:backwards in ConcatTable.

colesbury:
Overrides backwards to recursively call backwards on child modules. Many other containers, such as `nn.Sequential` and `nn.Concat` already do this.

This makes it easier to share gradInput storages between modules of the same type to save memory.

### DIFF
--- a/ConcatTable.lua
+++ b/ConcatTable.lua
@@ -27,12 +27,12 @@ local function retable(t1, t2, f)
    return t1
 end
 
-function ConcatTable:updateGradInput(input, gradOutput)
+local function backward(self, method, input, gradOutput, scale)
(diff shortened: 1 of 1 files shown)


JeremyMorvan:
This may be a good idea but could the 'local backward' method be a class method ? I made a child class of ConcatTable and I had to rewrite both backward and updateGradInput.

All containers could have the same code format for updateGradInput and backward as they have no parameters to update. I just checked nn.Sequential and nn.Concat, only one function could be implemented instead of quasi-identical updateGradInput and backward (same goes for accGradParameters and accUpdateGradParameters).

soumith:
seems like a good idea, @colesbury

PR #655, opened 2016-02-20
Format THNN function headers and add basic docs

apaszke:
As discussed on gitter, I will add a script that automatically generates api_reference.md out of THNN.h in an upcoming PR.

### DIFF
--- a/lib/THNN/README.md
+++ b/lib/THNN/README.md
@@ -1,72 +1,34 @@
-## API design guidelines
+# THNN
 
-All functions should accept arguments in the following order. Dots represent any module-specific parameters or buffers, disregarding whether they are used for writing or reading. They should follow the order
-```
-[weight], [bias], [any buffers], [additional arguments], [optional arugments]
-```
+THNN is a library that gathers nn's C implementations of neural network modules. It's entirely free of Lua dependency and therefore can be used in any application that has a C FFI. Please note that it only contains quite low level functions, and an object oriented C/C++ wrapper will be created soon as another library.
 
-### Modules
-```
-updateOutput: state, input, output, ...
-updateGradInput: state, input, gradOutput, gradInput, ...
-accGradParameters: state, input, gradOutput, [gradWeight], [gradBias], ...
-```
+There is also a CUDA counterpart of THNN (CUTHNN) in the [cunn repository](https://github.com/torch/cunn/tree/master/lib/THCUNN).
 
-e.g.
-```C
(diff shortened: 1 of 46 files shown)


lib/THNN/README.md
```diff
@@ -1,72 +1,34 @@
...
-```
+THNN is a library that gathers nn's C implementations of neural network modules. It's entirely free of Lua dependency and therefore can be used in any application that has a C FFI. Please note that it only contains quite low level functions, and an object oriented C/C++ wrapper will be created soon as another library.
 
-### Modules
-```
-updateOutput: state, input, output, ...
-updateGradInput: state, input, gradOutput, gradInput, ...
-accGradParameters: state, input, gradOutput, [gradWeight], [gradBias], ...
-```
+There is also a CUDA counterpart of THNN (CUTHNN) in the [cunn repository](https://github.com/torch/cunn/tree/master/lib/THCUNN).
```
andreaskoepf: '(CUTHNN)' should be '(THCUNN)'

lib/THNN/README.md
```diff
@@ -1,72 +1,34 @@
...
-void THNN_(ClassNLLCriterion_updateOutput)(
-          THNNState* state,
-          THTensor *input,
-          THLongTensor *target,
-          THTensor *output,
-          THTensor *weights,
-          THTensor *total_weight,
-          bool sizeAverage)
-```
+Torch's nn module provided many optimized C implementations of modules, but the source files contained Lua specific code and headers so they couldn't be easily compiled and included anywhere else.
```
andreaskoepf: Instead of 'nn module' I would prefer 'Neural Network Package (nn)'.

lib/THNN/README.md
```diff
@@ -1,72 +1,34 @@
...
 
-Use 2 spaces for block indentation.
+* **updateOutput** - applies the module to an input
+* **updateGradInput** - accepts gradient w.r.t. output and previous module input, and computes a gradient w.r.t. that input
+* **accGradParameters** - *(optional, only modules with parameters)* accepts gradient w.r.t. output and previous module input, and computes gradient w.r.t. the parameters
 
+For information on argument types please check the [API reference](doc/api_reference.md).
 
-### Conversion Steps
+This is all THNN library provides. An object oriented implementation similar to nn will be provided in a separate library. This one is just a set of CPU kernels.
```
andreaskoepf: I would remove these three sentences. They sound a little bit informal and I think it is better not to 'promise' a future THNN++ before we really have something.

PR #656, opened 2016-02-20
added support for negative dimensions to JoinTable

CodeRect:


JoinTable.lua
```diff
@@ -8,11 +8,18 @@ function JoinTable:__init(dimension, nInputDims)
    self.nInputDims = nInputDims
 end
 
-function JoinTable:updateOutput(input)
+function JoinTable:_getPositiveDimension(input)
    local dimension = self.dimension
-   if self.nInputDims and input[1]:dim()==(self.nInputDims+1) then
-       dimension = dimension + 1
+   if dimension < 0 then
+      dimension = input:dim() + dimension + 1
```
soumith: if the input is 4-dimensional, and i give dimension to be -1, then I'd expect it to join dim-3, but with this formulation, dimension will be 4. What am I missing here...

CodeRect: the implementation is based on the SplitTable which uses the same logic

soumith: now that I think about it, -1 should be the last dimension, as that's the only way we can choose the last dimension via negative indexing and the rest of the dims will follow.

soumith:
thanks!

PR #659, opened 2016-02-22
Fixed VolumetricFullConvolution.

fbesse:
The module didn't pass gradients check and had limitations (batch mode only, constraints on kernel size and padding). This new implementation removes these limitations, and is now implemented like SpatialFullConvolution. The cuda implementation has been migrated to cunn.


fbesse:
Cunn PR: https://github.com/torch/cunn/pull/222

soumith:
my only question on this PR is, is it backward-compatible?
If I load an old model, do I not have to change the weight layouts to get correct outputs?

soumith:
at least, loading an older model should error out if possible. so that the user doesn't have pleasant surprises. But looking at the history of VolumetricFullConvolution, it's not very old, I doubt that there's a large body of affected users.

fbesse:
Yes there is a backCompatibility function which checks whether the module was loaded from an old version or not, and transposes the weights in this case. Also this module was not passing gradient checks previously so whoever was using it with a backwards pass probably has wrong weights anyway.

soumith:
great. could you squash the merge commits, (works well via "git rebase -i "). This is ready to be pushed.

soumith:
actually, i see two of you worked on this. the merge commits are fine. thanks for the contribution.

PR #660, opened 2016-02-22
Make nn.utils.recursiveType add empty Tensors to the tensorCache.

malcolmreynolds:
This means if we have a module which wraps another module, and does
something like:

function MyWrapper:__init(innerModule)
  self._innerModule = innerModule
  self.gradInput = self._innerModule.gradInput
end

.. Then calling :type before we have forwarded anything (and when we
still have zero-size Tensors) will now correctly preserve this aliasing.


soumith:
Thanks!

PR #661, opened 2016-02-23
Adding SparseLinear with CUDA, requires buffer variable

ebetica:


SparseLinear.lua
```diff
@@ -10,6 +10,7 @@ function SparseLinear:__init(inputSize, outputSize)
    self.gradWeight = torch.Tensor(outputSize, inputSize):zero()
    self.gradBias = torch.Tensor(outputSize):zero()
    self.lastInput = nil
+   self.cudaBuffer = torch.Tensor()
```
fmassa: Could you add the buffer to the `updateOutput` function, like in [here](https://github.com/torch/nn/blob/master/SpatialConvolutionMM.lua#L63) ?
And also clean it in the [`clearState` method](https://github.com/torch/nn/blob/master/SparseLinear.lua#L115-L118)?
This will allow using `clearState` to remove the temporary buffers while saving the network, and allowing to reconstruct saved models

fmassa:
As is, this PR breaks saved models that doesn't have `cudaBuffer`

soumith:
i'll fix this. merged it without thinking about back-compatibility.

PR #665, opened 2016-02-24
Add VolumetricBatchNormalization

colesbury:
All of the BatchNormalization modules now extend `nn.BatchNormalization` and use the same THNN/THCUNN implementation.

### DIFF
--- a/BatchNormalization.lua
+++ b/BatchNormalization.lua
@@ -5,20 +5,20 @@
                    by Sergey Ioffe, Christian Szegedy
 
    This implementation is useful for inputs NOT coming from convolution layers.
-   For Convolution layers, see SpatialBatchNormalization.lua
+   For convolution layers, use nn.SpatialBatchNormalization.
 
    The operation implemented is:
    y =     ( x - mean(x) )
         -------------------- * gamma + beta
-       standard-deviation(x)
+        standard-deviation(x)
    where gamma and beta are learnable parameters.
 
    The learning of gamma and beta is optional.
 
    Usage:
-   with    learnable parameters: nn.BatchNormalization(N [, eps] [,momentum])
+   with    learnable parameters: nn.BatchNormalization(N [,eps] [,momentum])
                                  where N = dimensionality of input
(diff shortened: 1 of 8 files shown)


fmassa:
I really like this PR, very clean and nicely done.
I have one question though, in the case one wants to use `THNN` as a standalone library: I think it's not intuitive that one should **always** feed a 3D tensor to `BatchNormalization` (even if we are in spatial or volumetric domain), and resize the output tensor afterwards so that it has the right size for the next module.
I was wondering if one couldn't change [this part](https://github.com/torch/nn/pull/665/files#diff-c88d7ef17f6b6afacbf4648573e41e25R14) to compute `n` in a for loop over the dimensions of the `input` which are > 2. This solution might not apply as easily to the `cunn` version though, as the input dimensions are [statically determined](https://github.com/torch/cunn/pull/224/files#diff-424d755ea21eebdc3281ce8199f1c2daR215)...

Maybe an easier solution (that could work for both backends) would be to do the resizing of the input/output inside the C/Cuda functions ?

colesbury:
@fmassa I think that's a good idea. The tensor:view() function is implemented in Lua but not in C. I think we can effectively do the same thing in `checktensor` in the CUDA version.

colesbury:
Updated PR with @fmassa's suggestion. (Tests pass VolumetricMaxPooling is flaky)

fmassa:
Very cool, thanks @colesbury !

hughperkins:
Hmmm, this breaks BatchNormalization in clnn :-(  For example, this means that http://torch.ch/blog/2015/07/30/cifar.html will no longer run on clnn

fmassa:
@hughperkins The previous commit (adding C/Cuda implementations of `SpatialBatchNormalization`) also is non-clnn friendly. Can't we (at least temporarily) monkey-patch clnn with the previous implementation ?  (But we should pay attention because the meaning of some variables have changed since then)

hughperkins:
Yes. "we" can :-P  However, it would be nice to get a little @hughperkins in any future pull request that strips out lua implementation, and replaces it with c implementation :-)

PR #674, opened 2016-02-27
Reuse headers from THNN in THNN.lua

fmassa:
Instead of having to declare the function headers both in `lib/THNN/generic/THNN.h` and `THNN.lua`, copy the headers from `lib/THNN/generic/THNN.h` into a lua file which returns a string with the content of `lib/THNN/generic/THNN.h`, and use it in `THNN.lua`. 
In the process, strips all lines starting with a `#`.

Also, add missing function declarations to `lib/THNN/generic/THNN.h` (which were present though in `THNN.lua`

Will do the same thing for 'THCUNN` when we agree on the format.


soumith:
looks good to me!

soumith:
but will wait a couple more days.

soumith:
let's merge this, prep one for THCUNN as well. you need to rebase.

fmassa:
Rebased against master. Will send a PR for THCUNN tomorrow.

Atcold:
LOL, @fmassa, I read "Rebased again, master." Hahaha :laughing:

fmassa:
Thanks @soumith 
I think one of the last bits missing in `THNN` is to move [this part of CMakeLists](https://github.com/torch/nn/blob/master/CMakeLists.txt#L6-L48) to [here](https://github.com/torch/nn/blob/master/lib/THNN/CMakeLists.txt#L11). What do you think ?

soumith:
yep, that makes sense to do!

fmassa:
Ok. Will send a PR

THNN.lua
```diff
@@ -2,994 +2,12 @@ local ffi = require 'ffi'
...
-
-TH_API void THNN_(SpatialReplicationPadding_updateGradInput)(THNNState *state,
-                                                            THTensor *input,
-                                                            THTensor *gradOutput,
-                                                            THTensor *gradInput,
-                                                            int pad_l, int pad_r,
-                                                            int pad_t, int pad_b);
-
-]]
+local generic_THNN_h = require 'nn.THNN_h'
```
Atcold: I'm lost. Where is `THNN_h.lua`?

soumith: can you put some effort and find it. even doing a simple text search on this page will find it for you

Atcold: :unamused: ...
So far I found `lib/THNN/generic/THNN.h` and `THNN.lua`.
There is not any `THNN_h.lua` around.

fmassa: @Atcold It's defined in the [CMakeLists here](https://github.com/torch/nn/pull/674/files#diff-af3b638bc2a3e6c650974192a53c7291R53). It's basically copying `THNN.h` to `THNN_h.lua`

Atcold: Thank you, @fmassa. I was browsing the file tree, and I didn't think it was made on the fly.
I saw also that there are two `THNN.h`... I'm having a look at that now.

PR #680, opened 2016-03-02
Better __tostring()__ and cleans formatting

Atcold:
Better `__tostring()__` for *pooling* modules
Fixing code formatting (Vim took care of that)


Atcold:
Thank you :grimacing:

PR #687, opened 2016-03-04
Smaller dimensions for several tests

GeorgOstrovski:
These tests (VolumetricConvolution, SpatialConvolutionLocal) could previously take very long (up to several minutes on my machine) and are faster with the new dimensions without sacrificing much generality. The speed-up is roughly from ~1min -> ~10sec for each of these tests.


fmassa:
This was indeed much needed !

soumith:
Thanks!

PR #691, opened 2016-03-04
Add _type field to nn.Module

colesbury:
Currently, there is no clear way to get the 'type' of an nn.Module,
although you can set the type via Module:type(<newtype>). This changes
Module:type() to return the current type when called with no arguments.
The module's type is stored in the _type field.

### DIFF
--- a/Module.lua
+++ b/Module.lua
@@ -3,6 +3,7 @@ local Module = torch.class('nn.Module')
 function Module:__init()
    self.gradInput = torch.Tensor()
    self.output = torch.Tensor()
+   self._type = self.output:type()
 end
 
 function Module:parameters()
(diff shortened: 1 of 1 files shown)


soumith:
We've had this for a long time internally, and find it very useful. @dominikgrewe @koraykv @clementfarabet you guys okay with this?

fmassa:
I agree that this is very useful.
I wonder though if the type of the network shouldn't be undefined when we have a `nn.Copy` to cast from one type to another (for example in [cifar.torch](https://github.com/szagoruyko/cifar.torch/blob/master/train.lua#L44-L48)). Should we care about that ?

soumith:
I think in nn.Copy, it will be the type of the output, which seems correct.

clementfarabet:
Yes that's useful, thanks for the patch.

Issue #683, opened 2016-03-03
Better __tostring()__ for containers


Atcold:
I was thinking if something like this would be wanted. For me it helps a lot.

![screenshot 2016-03-02 21 00 19](https://cloud.githubusercontent.com/assets/2119355/13482260/d69ca8f8-e0b9-11e5-9067-4631692526e6.png)
![screenshot 2016-03-02 21 00 39](https://cloud.githubusercontent.com/assets/2119355/13482263/d9074c56-e0b9-11e5-8fca-202e32f49033.png)

How do you like it?

Atcold:
So, it becomes easier to read net diagrams from terminal. For example the *GoogLeNet* here below.

![screenshot 2016-03-02 21 50 19](https://cloud.githubusercontent.com/assets/2119355/13483338/6c307866-e0c1-11e5-97bd-536878954f63.png)

Atcold:
`pretty-nn` is available now as a stand alone package https://github.com/Atcold/torch-pretty-nn.

PR #692, opened 2016-03-04
Add optional pretty print for containers

Atcold:
Sometimes (see #683's *GoogLeNet*) it becomes a little messy playing with `nn.Sequential()`, `nn.Concat()` and `nn.Parallel()`. So, I thought that some colour coding would help everyone.
Pretty print can be disabled (when redirecting the std output to file, for example) by setting `nn.config.prettyPrint` to `false`.

```lua
require 'nn'
c = nn.Parallel(1,2)
for i = 1, 3 do
   local t = nn.Sequential()
   t:add(nn.Linear(4, 2))
   t:add(nn.Reshape(2, 1))
   c:add(t)
end
split = nn.Concat(2)
split:add(c)
block = nn.Sequential()
block:add(nn.View(-1))
block:add(nn.Linear(12, 4))
block:add(nn.View(2, 2))
split:add(block)

print(split)
nn.config.prettyPrint = false
print(split)
pred = split:forward(torch.randn(3, 4))
```

screenshot 2016-03-04 15 26 51

### DIFF
--- a/Concat.lua
+++ b/Concat.lua
@@ -93,22 +93,24 @@ function Concat:accUpdateGradParameters(input, gradOutput, lr)
 end
 
 function Concat:__tostring__()
+   local r = function(s) return s end
+   if nn.config.prettyPrint then r = require('trepl.colorize').red end
    local tab = '  '
    local line = '\n'
-   local next = '  |`-> '
-   local ext = '  |    '
+   local next = r '  |`-> '
+   local ext = r '  |    '
    local extlast = '       '
-   local last = '   ... -> '
-   local str = torch.type(self)
-   str = str .. ' {' .. line .. tab .. 'input'
+   local last = r '   ... -> '
+   local str = r(torch.type(self))
+   str = str .. r ' {' .. line .. tab .. r 'input'
    for i=1,#self.modules do
       if i == #self.modules then
-         str = str .. line .. tab .. next .. '(' .. i .. '): ' .. tostring(self.modules[i]):gsub(line, line .. tab .. extlast)
+         str = str .. line .. tab .. next .. r '(' .. i .. r '): ' .. tostring(self.modules[i]):gsub(line, line .. tab .. extlast)
       else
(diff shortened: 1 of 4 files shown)


soumith:
it's prettier, but:
1. it introduces a dependency on trepl
2. it doesn't auto-detect terminal's color capabilities, it's on the user to avoid the color handling, breaks all existing scripts.
3. It doesn't add a lot of actual value

I am closing it for the reasons above.
I dont always want to reject PRs that youv'e put effort into, so next time you want to work on something, open an issue first with what you propose to do.

Atcold:
@soumith, the issue was here https://github.com/torch/nn/issues/683, I guess next time I'll wait for some feedback.

 1. It's fixable. The colour coding can be hard coded.
 2. Can be disabled with `nn.config.prettyPrint = false`, and I'm not sure what can be broken.
 3. It actually does. If you check https://github.com/torch/nn/issues/683, you'd see how tricky is to understand what's going on.

soumith:
i am not going to accept (2) with some "can be disabled with ...". Either it auto-detects or it is disabled by default.

Atcold:
OK, fair enough. Does `trepl` support auto-detection (which I'm not even sure I understand what it is). A pointer on this would help.
If it does, I think it won't be a big deal to port over, no?

soumith:
trepl supports very flaky auto-detection that have a high chance of failure. I do not want that flaky stuff in nn. I've thought about this and i dont think a default-true is very easy to do. Either disable it by default, or maintain the colorization in your own separate library called nncolorize or something.

Atcold:
So, opting for
```lua
nn.config.prettyPrint = false
```
is all it takes? Then it's likely people won't even notice this feature.

szagoruyko:
separate library like nncolorize or nnswag is better I agree

Atcold:
What about checking whether `package.loaded.trepl` exist or it is `nil`?

PR #693, opened 2016-03-06
Add Unsqueeze Module

pengsun:
This module, nn.Unsqueeze, is the "inverse" operation of nn.Squeeze. It inserts a singleton dim (i.e., dimension 1) at specified position.


Unsqueeze.lua
```diff
@@ -0,0 +1,68 @@
...
+
+local function _checkPos(pos)
+   local pos = pos or error('the position to insert singleton dim not specified')
+   assert(type(pos) == 'number')
+   assert(pos > 0)
+   return pos
+end
+
+local function _assertTensor(t)
+   assert(t or torch.isTensor(t), "Unsqueeze only works on tensor")
```
fmassa: this should be `t and torch.isTensor(t)` right ?

fmassa: actually, only `torch.isTensor(t)` is enough.

szagoruyko:
the name is ambiguous I think

fmassa:
Another point, the forward pass could leverage [`nn.utils.addSingletonDimension`](https://github.com/torch/nn/blob/master/utils.lua#L133). This would avoid code redundancy.

Unsqueeze.lua
pengsun: @fmassa Yeah, you're right, thanks for pointing out this. Will modify the code.

pengsun:
@szagoruyko Initially, I named it as `InsertSingletonDim`. Latter at torch 7 gitter chatting room @soumith mentioned the name `Unsqueeze`, which I think is not bad, since it recalls one its connection to `nn.Squeeze`. So what shall we call this module in the end? Maybe nn.AddSingletonDimension ?

pengsun:
@fmassa Ah... didn't notice this function at all, since it is not mentioned at the doc... Yeah, I would try to wrap it.

soumith:
Thanks @pengsun , nicely done.

pengsun:
Happy to contribute :-)

Atcold:
:+1:

PR #696, opened 2016-03-07
Adding nn.MV

ebetica:
Adding matrix vector multiplication in nn, following the conventions of nn.MM.

### DIFF
--- a/MV.lua
+++ b/MV.lua
@@ -1,0 +1,82 @@
+--[[ Module to perform matrix vector multiplication on two minibatch inputs,
+producing a minibatch.
+]]
+
+local MV, parent = torch.class('nn.MV', 'nn.Module')
+
+-- Backward compatibility
+local unpack = unpack or table.unpack
+
+function MV:__init(trans)
+  parent.__init(self)
+
+  self.trans = trans or false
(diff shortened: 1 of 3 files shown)


MV.lua
```diff
@@ -0,0 +1,78 @@
+--[[ Module to perform matrix vector multiplication on two minibatch inputs,
+producing a minibatch.
+]]
+
+local MV, parent = torch.class('nn.MV', 'nn.Module')
+
+function MV:__init(trans)
+  parent.__init(self)
+
+  self.trans = trans or false
```
soumith: assert that this is boolean.

MV.lua
```diff
@@ -0,0 +1,78 @@
...
+
+  self.gradInput = {torch.Tensor(), torch.Tensor()}
+end
+
```
soumith: local unpack  = unpack or table.unpack, and use that function in the rest of the module. this is for Lua back-compatibility reasons.

soumith:
minor in-line comments, looks good otherwise.

ebetica:
Fixed.

soumith:
Thanks Zeming!

PR #697, opened 2016-03-07
Adding nn.SelectCriterion

lopezpaz:
Analogous to nn.SelectTable, but for criterions.

### DIFF
--- a/SelectCriterion.lua
+++ b/SelectCriterion.lua
@@ -1,0 +1,51 @@
+local SelectCriterion, parent = torch.class('nn.SelectCriterion', 'nn.Module')
+
+function SelectCriterion:__init(index)
+  parent.__init(self)
+  self.index = index
+  self.gradInput = {}
+end
+
+function SelectCriterion:updateOutput(input)
+   local criterions, preds, targets = input[1], input[2], input[3]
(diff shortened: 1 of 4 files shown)


soumith:
The builds fail. it is because you've not added SelectCriterion to init.lua
https://travis-ci.org/torch/nn/builds/114287408

lopezpaz:
Added.

SelectCriterion.lua
```diff
@@ -0,0 +1,51 @@
+local SelectCriterion, parent = torch.class('nn.SelectCriterion', 'nn.Module')
```
soumith: derive from nn.Criterion

ebetica:
FYI the build is failing because of a bug with the tester. This has been fixed so triggering a rebuild will pass it.

soumith:
this PR is fairly wrong, so i'll just close it for now.

nicholas-leonard:
You can use dpnn's [ModuleCriterion](https://github.com/Element-Research/dpnn/blob/master/README.md#nn.ModuleCriterion) to do this kind of thing for any module, not just SelectTable. Also works on either inputs or targets (or both). Would you like a PR?

fmassa:
I think this ModuleCriterion would be a nice addition !

PR #698, opened 2016-03-07
 Adding table input support for batched SparseLinear, implementing gradInput correctly, fixing other bugs

ebetica:
updateGradInput has a dense gradInput, so removing that to match LookupTable. We keep a resize for compatibility (with, for example, an nngraph.Identity() node on an input). Fixed the test not to test for gradInput.

Edit: After some discussion I'm redoing SparseLinear:
1. Now supports table inputs, of the format {batch1, batch2, batch3,...} where each batch is an ninput x 2 sparse vector. This lets us input batches of unequal vector sizes
2. UpdateGradInput is now coded correctly to return a dense vector coded in sparse format.
3. UpdateParameters and zeroGradParameters were wrong if we called accGradParameters more than once, because it did sparse updates only on the last input.

The original batchnum x nnz x 2 format is still supported for now...

### DIFF
--- a/SparseLinear.lua
+++ b/SparseLinear.lua
@@ -1,19 +1,25 @@
 local THNN = require 'nn.THNN'
 local SparseLinear, parent = torch.class('nn.SparseLinear', 'nn.Module')
 
-function SparseLinear:__init(inputSize, outputSize)
+local NO_LAST_INPUT = 0
+local ONE_LAST_INPUT = 1
+local ACC_MULTIPLE_TIMES = 2
+
+function SparseLinear:__init(inputSize, outputSize, doGradInput)
    parent.__init(self)
 
    self.weightDecay = 0
+   self.doGradInput = doGradInput or false
    self.weight = torch.Tensor(outputSize, inputSize):zero()
    self.bias = torch.Tensor(outputSize):zero()
    self.gradWeight = torch.Tensor(outputSize, inputSize):zero()
    self.gradBias = torch.Tensor(outputSize):zero()
-   self.lastInput = nil
 
-   if torch.getnumthreads() > 1 and outputSize >= 128 then
-      self.shardBuffer = torch.Tensor(outputSize, torch.getnumthreads())
-   end
+   assert(type(self.doGradInput) == type(true))
+
+   self.lastInput = nil
+   self.sparseUpdate = NO_LAST_INPUT
+   self.formatted_input = nil
 
    -- state
    self.gradInput:resize(inputSize)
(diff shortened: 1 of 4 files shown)


soumith:
cc: @zhangxiangxiao @myhrev

soumith:
cc: @MichaelMathieu

pengsun:
I think removing `updateGradInput()` would be too much, as we may occasionally need the `gradInput` even if this module is the first one in the network (e.g,. we may want to generate the word-level saliency map or something...).

But I do think we can add a "switch mechanism" for this kind of module to "turn off" the `backward()` (e.g., let `gradInput` be a null tensor).

By the way, I also think we can add `updateGradInput()` as well as a "switch" to `nn.LookupTable`, something like: https://github.com/pengsun/onehot-temp-conv/blob/master/LookupTableExt.lua

soumith:
@pengsun the current updateGradInput is wrong completely

pengsun:
@soumith Oh, sorry, didn't realize that... deleting my previous post...

ebetica:
The current updateGradInput returns something sparse when it should be dense. The stupid way to do this would just be to return something like:

{{1, v1}, {2, v2}, {3, v3},...}

which would be easy to implement and also correct. I'd imagine the reason we don't have it in LookupTable is because of the blowup in gradient size, since it's no longer sparse.

Also, I'm pretty sure the zeroGradParameters and updateParameters functions are also wrong, since they do a sparse zero/update based on the last input. If we do multiple forward/backward passes before calling any of these functions the result will be garbled.

ebetica:
I changed this pull request to basically be fixing SparseLinear, see above edit for all the changes.

SparseLinear.lua
```diff
@@ -116,22 +192,24 @@ function SparseLinear:updateParameters(learningRate)
...
-         self.gradBias:cdata(),
-         self.lastInput:cdata()
+      self.gradWeight:cdata(),
+      self.gradBias:cdata(),
+      self.lastInput:cdata()
       )
    else
       parent.zeroGradParameters(self)
    end
+   self.sparseUpdate = 0
```
soumith: self.sparseUpdate = NO_LAST_INPUT

SparseLinear.lua
```diff
@@ -116,22 +192,24 @@ function SparseLinear:updateParameters(learningRate)
    else
       parent.updateParameters(self, learningRate)
    end
+   self.sparseUpdate = 0
```
soumith: self.sparseUpdate = NO_LAST_INPUT

soumith:
Thanks Zeming!

zhangxiangxiao:
Sorry a bit late to the discussion. The original SparseLinear module does *not* support batch on CPU. 

This module has been used in various implementations of memory networks, ranking models and word vector models. One crucial point was that it can support fast forward and backward propagation on indexed values only via self.lastInput. Parallelization is taken care outside this module itself via higher level optimization techniques such as HogWILD.

The new implementation breaks this by setting self.legacyMode = false when the input is a dimension 2 tensor (i.e., not a batch input). This breaks many projects where using self.lastInput was intended. Can you improve on the back-compatibility of this?

soumith:
@zhangxiangxiao can you give a quick test case so that I can hotfix the module rather than reverting all of it? Now that there's a CUDA version that I merged in yesterday, i'll have to revert a bunch of things.
As soon as you give a repro, i can hotfix it within 24 hours.

zhangxiangxiao:
@soumith This is a performance issue rather than a correctness issue. I can write a piece of code showing where this is necessary.

soumith:
when input is of dimension 2, if we set self.legacyMode to true, that should fix the perf issues then. That's doable..., I dont see why not.

zhangxiangxiao:
@soumith Yeah that's what I thought too. I am not very sure except for updateParamters because I did not look at other functions. Writing the test code...

ebetica:
The only functions that do not do the sparse updates are updateParameters and zeroGradParameters, like you said. I did not realize the performance hit would be so big, but in addition to the hotfix I can also do a patch for doing sparse updates in general on lastInput.

@soumith 's suggestion would disable CUDA for single batch updates.

zhangxiangxiao:
Okay here is a piece of testing code:
```lua
local nn = require('nn')
local sys = require('sys')

local model = nn.SparseLinear(65536, 256)

local input = torch.rand(5, 2)
input:select(2, 1):mul(65536):ceil()

sys.tic()
local output = model:forward(input)
sys.toc(true)

local gradOutput = torch.rand(output:size())

sys.tic()
local gradInput = model:backward(input, gradOutput)
sys.toc(true)

sys.tic()
model:updateParameters(1e-3)
sys.toc(true)
```

And here is the output:
```
$ th test.lua
0.00025105476379395
0.00012683868408203
0.022439002990723
```

It seems only updateParameters were unnecessarily slow. I also changed legacyMode to true for when input:dim() == 2 but there is some bug somewhere.

zhangxiangxiao:
@ebetica Oops yes zeroGradParameters is very important to be fast too (as it is also called once every update)

ebetica:
@zhangxiangxiao Is this level of sparsity a common use case? The CUDA module is not optimized for single batch extremely sparse inputs.

zhangxiangxiao:
It is a common use case for natural language processing, where this module is mostly used. A single core on my machine can go over 5000 samples per second, and using a 10-thread HOGWILD trainer my throughput is at the level of 50,000 samples per second.

At this level of sparsity CPU is much faster than GPU. But I think the two use cases are so distinct that it is good to combine the two.

PR #700, opened 2016-03-08
Update simple.md

n3011:
Improved documentation of Batch Normalization by adding running mean and variance update equation. better understanding and readability.

### DIFF
--- a/doc/simple.md
+++ b/doc/simple.md
@@ -1163,7 +1163,11 @@ where `N` is the dimensionality of input
 `affine` is a boolean. When set to false, the learnable affine transform is disabled. Defaults to true
 
 During training, this layer keeps a running estimate of its computed mean and std.
-The running sum is kept with a default momentum of 0.1 (unless over-ridden)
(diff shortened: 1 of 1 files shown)


doc/simple.md
```diff
@@ -1163,7 +1163,11 @@ where `N` is the dimensionality of input
 `affine` is a boolean. When set to false, the learnable affine transform is disabled. Defaults to true
 
 During training, this layer keeps a running estimate of its computed mean and std.
-The running sum is kept with a default momentum of 0.1 (unless over-ridden)
+The running sum is kept with a default momentum (alpha) of 0.1 (unless over-ridden)
+```lua
+running_mean = (1 - alpha) * running_mean + alpha * batch_mean
+running_inv_std = (1 - alpha) * running_inv_std + alpha * batch_inv_std
```
fmassa: The code was changed. now it's the variance that is updated by moving averages, see https://github.com/torch/nn/blob/master/lib/THNN/generic/BatchNormalization.c#L47

soumith:
as Francisco pointed out, this doc change is no longer valid either. now it is running_var (running variance)

PR #701, opened 2016-03-08
Move compilation flags from nn to THNN CMakeLists

fmassa:
... so that `THNN` can be compiled as a standalone library.


fmassa:
Wait a bit, I'm checking it again

fmassa:
I think this is ok. I had some old build files messing up with my tests.
But I don't trust that much my CMake skills... :)

apaszke:
@fmassa Looks ok to me

PR #703, opened 2016-03-09
clearState preserves structure in tables with Tensors

funous:
When clearState is used on a table with tensors, it now preserves the structure (recursively) and creates a table with empty tensors.

### DIFF
--- a/Container.lua
+++ b/Container.lua
@@ -128,19 +128,21 @@ end
 
 function Container:clearState()
    -- don't call set because it might reset referenced tensors
-   local function clear(f)
-      if self[f] then
-         if torch.isTensor(self[f]) then
-            self[f] = self[f].new()
-         elseif type(self[f]) == 'table' then
-            self[f] = {}
-         else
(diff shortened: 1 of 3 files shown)


Container.lua
```diff
@@ -128,19 +128,21 @@ end
...
-      if self[f] then
-         if torch.isTensor(self[f]) then
-            self[f] = self[f].new()
-         elseif type(self[f]) == 'table' then
-            self[f] = {}
-         else
-            self[f] = nil
+   local function clear(t)
+      if torch.isTensor(t) then
+         return t.new()
```
soumith: shouldn't this be t:set()? t.new() creates a new tensor.

soumith: nevermind, i read the comment.

fmassa: I think that if https://github.com/torch/nn/pull/673 gets merged, we will be able to just call `:set()` on the tensors and keep all the references to the network tensors after `clearState`.

PR #704, opened 2016-03-09
In-place ELU

anibali:
Add an option to do ELU in-place, as can already be done with ReLU.

### DIFF
--- a/ELU.lua
+++ b/ELU.lua
@@ -6,28 +6,36 @@ local ELU, parent = torch.class('nn.ELU', 'nn.Module')
    http://arxiv.org/pdf/1511.07289.pdf
 --]]
 
-function ELU:__init(alpha)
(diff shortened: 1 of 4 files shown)


ELU.lua
```diff
@@ -6,17 +6,20 @@ local ELU, parent = torch.class('nn.ELU', 'nn.Module')
...
+   assert(type(self.inplace) == 'boolean')
 end
 
 function ELU:updateOutput(input)
    input.THNN.ELU_updateOutput(
       input:cdata(),
       self.output:cdata(),
-      self.alpha
+      self.alpha,
+      self.inplace
```
soumith: prob change this to: self.inplace or false (as saved models wont have the .inplace member)

anibali:
@soumith It should work for older saved models now.

soumith:
Thanks man!

PR #705, opened 2016-03-10
Update addBuffer size in accGradParameters

funous:
Calling addGradParameters with an input of different size than the input used for updateOutput resulted in an error previously. This is an issue for some efficient RNN implementations that merge updates together.

This change has no effect if Linear is used in the usual way, it only improves some corner cases.


soumith:
Thanks!

PR #706, opened 2016-03-11
nn/MaskedSelect.lua

rfaulkner:
Add MaskedSelect module which emulates the bahaviour of torch.Tensor():maskedSelect().

e.g.
```
> ms = nn.MaskedSelect()
> mask = torch.randn(2,2):bernoulli():byte()
> input = torch.randn(2,2)
> =mask
 0  0
 1  0
[torch.ByteTensor of size 2x2]

> =input
 0.6046  1.3675
 0.2091 -1.9718
[torch.DoubleTensor of size 2x2]

> out = ms:forward({input, mask})
> =out
 0.2091
[torch.DoubleTensor of size 1]

>  gradIn = ms:backward({input, mask}, out)
> =gradIn
{
  1 : DoubleTensor - size: 2x2
  2 : ByteTensor - size: 2x2
}
> return gradIn[1]
 0.0000  0.0000
 0.2091  0.0000
[torch.DoubleTensor of size 2x2]

> return gradIn[2]
 0  0
 0  0
[torch.ByteTensor of size 2x2]

```


MaskedSelect.lua
```diff
@@ -0,0 +1,56 @@
...
+  parent.type(self, type)
+  if type ~= 'torch.CudaTensor' then
+    self.mask = mask:byte()
+    self.maskIndices = maskIndices:byte()
+  end
+  return self
+end
+
+function MaskedSelect:gc()
+  nn.clearFields(self, 'output', 'gradInput')
```
fmassa: Is this function defined somewhere ?
Also, maybe it would be better to rename it to `clearStates` https://github.com/torch/nn/pull/526 ?

MaskedSelect.lua
```diff
@@ -0,0 +1,56 @@
...
+
+--[[ Reverse maps unmasked gradOutput back to gradInput. ]]
+function MaskedSelect:updateGradInput(input, gradOutput)
+  self.gradInput = torch.Tensor(input:size():prod())
+  local i = 1
+  local f = function(x)
+    self.gradInput[x] = self.gradOutput[i]
+    i = i + 1
+  end
+  self.maskIndices:apply(f)
```
fmassa: is this part equivalent to [`gather`](https://github.com/torch/torch7/blob/master/doc/tensor.md#tensor-gatherdim-index) ? 
The problem with the `:apply` function is that it doesn't work on GPU.

MaskedSelect.lua
rfaulkner: updated this to use nn.utils.clear and renamed.

MaskedSelect.lua
rfaulkner: scatter works perfectly here, thanks!

MaskedSelect.lua
```diff
@@ -0,0 +1,41 @@
...
+--[[ Sets the provided mask value for the module. ]]
+function MaskedSelect:__init()
+  -- mask must be a tensor, convert to byte tensor
+  parent.__init(self)
+end
+
+--[[ Performs maskedSelect operation. ]]
+function MaskedSelect:updateOutput(input)
+  local input, mask = unpack(input)
+  self.output = input:maskedSelect(mask)
```
dominikgrewe: This will allocate a new tensor each time. Re-use `self.output` instead:

`self.output:maskedSelect(input, mask)`

MaskedSelect.lua
```diff
@@ -0,0 +1,41 @@
...
+  return self.output
+end
+
+--[[ Reverse maps unmasked gradOutput back to gradInput. ]]
+function MaskedSelect:updateGradInput(input, gradOutput)
+  local input, mask = unpack(input)
+  local maskIndices =
+    torch.range(1, mask:nElement()):resizeAs(mask:double()):maskedSelect(mask)
```
dominikgrewe: Same here. Keep `maskIndices` around as a private member of the class and call `self._maskIndices:range(...)`.

You should also avoid casting the mask to double, so instead of `resizeAs(mask:double())` do `resize(mask:size())`.

MaskedSelect.lua
```diff
@@ -0,0 +1,41 @@
...
+  local gradInput = torch.Tensor(input:nElement()):zero()
+  gradInput:scatter(1, maskIndices:long(), gradOutput)
+  gradInput:resizeAs(input:double())
+  self.gradInput = {gradInput, torch.Tensor():resizeAs(input):fill(0):byte()}
```
dominikgrewe: There are a few more places here where you create a new tensor each time. Try to avoid that by keeping tensors around as members.

MaskedSelect.lua
```diff
@@ -0,0 +1,41 @@
...
+
+function MaskedSelect:accGradParameters(input, gradOutput, scale)
+end
+
+function MaskedSelect:parameters()
+  return nil
+end
+
+function MaskedSelect:type(type)
+  return self
```
dominikgrewe: This means the module isn't type-castable.

MaskedSelect.lua
rfaulkner: It doesn't look like maskedSelect lets you do this.  It returns the result in a new tensor rather than mutating the calling object.

fmassa: @rfaulkner Are your sure about that ? I just tried it and it worked just fine.

```lua
-- also works for cuda tensors
a = torch.rand(5,5)
b = torch.rand(5,5):gt(0.5)
c = torch.Tensor()
c:maskedSelect(a,b)
```

MaskedSelect.lua
rfaulkner: see new definition

MaskedSelect.lua
rfaulkner: done

MaskedSelect.lua
rfaulkner: done

MaskedSelect.lua
rfaulkner: done

MaskedSelect.lua
```diff
@@ -0,0 +1,69 @@
...
+end
+
+function MaskedSelect:clearStates()
+  nn.utils.clear(self, {'output',
+                        'gradInput',
+                        '_maskIndices',
+                        '_gradBuffer',
+                        '_maskBuffer',
+                        '_gradMask'})
```
fmassa: There is a typo here, it should be `clearState` instead of `clearStates`.
Also, it should return `self` https://github.com/torch/nn/blob/master/Module.lua#L380

MaskedSelect.lua
```diff
@@ -0,0 +1,69 @@
...
+  self._gradBuffer:resize(input:size())
+  self.gradInput = {self._gradBuffer,
+                    self._gradMask:resize(mask:size()):fill(0)}
+  return self.gradInput
+end
+
+function MaskedSelect:accGradParameters(input, gradOutput, scale)
+end
+
+function MaskedSelect:parameters()
```
fmassa: do we need this ?
Same for `accGradParameters`

MaskedSelect.lua
```diff
@@ -0,0 +1,69 @@
...
+  return nil
+end
+
+function MaskedSelect:type(type)
```
fmassa: need to pass `tensorCache` https://github.com/torch/nn/blob/master/Module.lua#L117

MaskedSelect.lua
```diff
@@ -0,0 +1,69 @@
...
+  local gradBuffer = self._gradBuffer
+  local maskIndexBuffer = self._maskIndexBuffer
+  if type ~= 'torch.CudaTensor' then
+    self._maskIndices = nil
+    self._gradBuffer = nil
+    self._maskBuffer = nil
+    self._gradMask = nil
+  end
+
+  parent.type(self, type)
```
fmassa: also need `tensorCache` here

MaskedSelect.lua
```diff
@@ -0,0 +1,69 @@
...
+  return self.output
+end
+
+--[[ Reverse maps unmasked gradOutput back to gradInput. ]]
+function MaskedSelect:updateGradInput(input, gradOutput)
+  local input, mask = unpack(input)
+  self._maskIndexBuffer:range(1, mask:nElement()):resize(mask:size())
+  self._maskIndices:maskedSelect(self._maskIndexBuffer, mask)
+  self._gradBuffer:resize(input:nElement()):zero()
+  self._gradBuffer:scatter(1, self._maskIndices:long(), gradOutput)
```
fmassa: this is not cuda-friendly, as the cuda version of `scatter` requires the indices to be `CudaTensor`.
Also, this line is allocating new memory at every iteration, you can avoid that by having another temporary tensor

MaskedSelect.lua
```diff
@@ -0,0 +1,69 @@
...
+    self._gradBuffer = nil
+    self._maskBuffer = nil
+    self._gradMask = nil
+  end
+
+  parent.type(self, type)
+
+  if type ~= 'torch.CudaTensor' then
+    self._maskIndices = maskIndices:long()
+    self._gradBuffer = gradBuffer:double()
```
fmassa: you should cast to `type` here and the line below (if not it will only work for `DoubleTensor`

MaskedSelect.lua
```diff
@@ -0,0 +1,69 @@
...
+function MaskedSelect:updateOutput(input)
+  local input, mask = unpack(input)
+  self.output:maskedSelect(input, mask)
+  return self.output
+end
+
+--[[ Reverse maps unmasked gradOutput back to gradInput. ]]
+function MaskedSelect:updateGradInput(input, gradOutput)
+  local input, mask = unpack(input)
+  self._maskIndexBuffer:range(1, mask:nElement()):resize(mask:size())
```
fmassa: Currently, there is no `range` function in `cutorch`, meaning that this line won't work as is. A workaround would be to have another buffer in the CPU which computes the `range`, and then copy it to CPU.

MaskedSelect.lua
rfaulkner: done.

MaskedSelect.lua
rfaulkner: removed.

MaskedSelect.lua
rfaulkner: done.

MaskedSelect.lua
rfaulkner: done.

MaskedSelect.lua
rfaulkner: added a CPU buffer to perform this operation.

MaskedSelect.lua
rfaulkner: type casting refactored

MaskedSelect.lua
rfaulkner: added a CPU buffer to perform this operation.

rfaulkner:
@fmassa please let me know if you think this PR needs anymore work.  Thanks for all of the review so far!

rfaulkner:
@dominikgrewe still waiting for feedback on this one, LGTM from your perspective?  @fmassa Please advise if you think this PR is LGTM or needs further work.

MaskedSelect.lua
```diff
@@ -0,0 +1,65 @@
...
+function MaskedSelect:updateOutput(input)
+  local input, mask = table.unpack(input)
+  self.output:maskedSelect(input, mask)
+  return self.output
+end
+
+--[[ Reverse maps unmasked gradOutput back to gradInput. ]]
+function MaskedSelect:updateGradInput(input, gradOutput)
+  local input, mask = table.unpack(input)
+  if type == 'torch.CudaTensor' then
```
fmassa: this should be `input:type()`, right ? or maybe `self._type` ? `type` is not yet defined in this context.

test.lua
```diff
@@ -1306,6 +1306,15 @@ function nntest.MarginRankingCriterion()
 
 end
 
+function nntest.MaskedSelect()
+   local input = torch.randn(4, 5):type('torch.DoubleTensor')
+   local mask = torch.ByteTensor(4, 5):bernoulli()
+   local module = nn.MaskedSelect():type('torch.DoubleTensor')
+   local out = module:forward({input, mask})
+   local err = out:dist(input:maskedSelect(mask))
+   mytester:assertlt(err, 1e-15, torch.typename(module) .. ' - forward err ')
```
fmassa: could you please add some tests for the backward pass ?

fmassa:
@rfaulkner sorry for the delay. Except for the two small comments that I added, it looks good to me, thanks !
One last thing, could you add some documentation for this module ?

MaskedSelect.lua
```diff
@@ -0,0 +1,65 @@
...
+  self._maskIndices = torch.LongTensor()
+  self._maskIndexBuffer = torch.LongTensor()
+  self._maskIndexBufferCPU = torch.FloatTensor()
+  self._gradBuffer = torch.Tensor()
+  self._gradMask = torch.ByteTensor()
+end
+
+--[[ Performs maskedSelect operation. ]]
+function MaskedSelect:updateOutput(input)
+  local input, mask = table.unpack(input)
```
fmassa: one last thing, I'm not sure that `table.unpack` exist in Lua 5.1. what people usually do is to add in the beginning of the file
```lua
local unpack = unpack or table.unpack
```
and then call `unpack` afterwards. Not sure we care that much about Lua 5.1 compatibility though (I'm thinking about https://github.com/torch/torch7/issues/516#issuecomment-176134155 )

rfaulkner: done.

MaskedSelect.lua
rfaulkner: thanks for catching that!  done.

test.lua
rfaulkner: added.

rfaulkner:
@fmassa where is the best place to put docs for this module?  Should these simply be added to the module itself?

dominikgrewe:
Looks good to me. @fmassa any other comments?

fmassa:
@rfaulkner yes, docs in `simple.md` is the best place for this module. Also, sorry for taking long to respond.

This looks good to me as well, could you squash your commits into a single commit so that it can be merged ?

I only have one really not important remark: in the case one has only this module and it calls `module:type()`, it won't return the module type, as was recently introduced by https://github.com/torch/nn/pull/691. As this case will probably never happen, I'd say it's good to go as is. What do you think @dominikgrewe 

Thanks a lot @rfaulkner !

dominikgrewe:
Yes, good point. It should be easy to add so I think we should be consistent here.

Although there are a couple of modules that override Module:type() and also don't behave correctly...

fmassa:
@dominikgrewe all the modules I can think of call `parent.type(self,...)` so I think they should work. I was curious though to see if there are modules that do not follow this behaviour, and [searching for it](https://github.com/torch/nn/search?p=1&q=type%28type%2CtensorCache%29&utf8=%E2%9C%93) showed that `FlattenTable` does [not return the type](https://github.com/torch/nn/blob/ead8af9c14f26116c1fe4fc5d58a6b0de2b2603f/FlattenTable.lua#L97-L101). But that should work for the other modules though.

dominikgrewe:
Most modules call `parent.type(self, ...)` but they don't return its return value if that is not `nil`, e.g. LookupTable, Max, Normalize

fmassa:
@dominikgrewe indeed, I misses this point, thanks !

rfaulkner:
was having a bunch of issues merging with the squash this since a few commits from an old PR got in here from several months back.  A clean PR is here:

https://github.com/torch/nn/pull/752

this one can be closed.

PR #711, opened 2016-03-13
Add a test for Concat

apaszke:


apaszke:
I don't know how it is that I'm always falling into some corner cases in existing tests when I push :smile: 

I included a fix for SpatialReplicationPadding. The problem was that if the input had height of 6, and both bottom and top padding were -3, the output collapsed into a 2D tensor and `updadeGradInput` failed.

soumith:
Thanks Adam.

PR #712, opened 2016-03-14
View reuses the same internal tensors

fmassa:
Currently, every `forward`/`backward` creates a new tensor (which shares the same storage). 
This PR allows to keep track of the module's `output`/`gradInput`, as the tensor pointers won't change at every forward/backward pass.

I think that this behaviour should be enforced for all `nn` modules. If you agree, I can send a PR fixing this behaviour in other modules.

### DIFF
--- a/View.lua
+++ b/View.lua
@@ -26,8 +26,6 @@ end
 function View:__init(...)
    parent.__init(self)
    self:resetSize(...)
-   self.output = nil
-   self.gradInput = nil
    self.numInputDims = nil
 end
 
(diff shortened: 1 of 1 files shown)


soumith:
Totally agree. Have been writing new modules with this in mind.

fmassa:
Great ! :) I'll have a pass over `nn` modules and I'll prepare a PR later today (or when I find some time)

nagadomi:
This commit breaks older saved models. It does not have `self.output` and `self.gradInput`.
```
In 14 module of nn.Sequential:
/home/nagadomi/torch/install/share/lua/5.1/nn/View.lua:80: attempt to index field 'output' (a nil value)
```

fmassa:
@nagadomi sorry for breaking your loading. I haven't thought about the case in which one saves a model before ever forwarding/backwarding it. This should be fixed with #713.

PR #713, opened 2016-03-15
Fix backward-compatibility of nn.View

apaszke:
Fixes issues from #712

### DIFF
--- a/View.lua
+++ b/View.lua
@@ -75,6 +75,7 @@ local function batchsize(input, size, numInputDims, numElements)
 end
 
 function View:updateOutput(input)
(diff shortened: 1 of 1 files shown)


apaszke:
@soumith How about adding a backward-compatibility test? Travis could download an example network from somewhere and try to load and run one forward and backward pass through it.

soumith:
i like that idea. Who is up for creating an insane model that uses like almost all of nn modules, that should be fun.

apaszke:
@soumith I can do that. I first though of a frankenstein network, but then I realised that it would be enough to test each module separately haha.

PR #717, opened 2016-03-17
remove L1Penalty.clearState

clement-masson:
Calling L1Penalty's clearState resulted on an error.
The loss being a scalar, it has no .set method.

```
local loss = m*input:norm(1) 
self.loss = loss  
```
The default Module.clearState should be ok, so I simply removed the L1Penalty's override.


soumith:
Thanks!

clement-masson:
you're welcome :)

PR #718, opened 2016-03-18
Cuda bug

eulerreich:
This fixes a type bug where `nn.LookupTable(5, 6):forward(torch.ones(5))` works fine but `nn.LookupTable(5, 6):cuda():forward(torch.ones(5))` doesn't.


eulerreich:
pull request #719 fixes the errors with CI.

PR #719, opened 2016-03-18
Cuda bug

eulerreich:
This fixes a type bug where `nn.LookupTable(5, 6):forward(torch.ones(5))` works fine but `nn.LookupTable(5, 6):cuda():forward(torch.ones(5))` doesn't.

### DIFF
--- a/LookupTable.lua
+++ b/LookupTable.lua
@@ -15,7 +15,7 @@ end
 
 function LookupTable:backCompatibility()
    self._count = self._count or torch.IntTensor()
-   self._input = self._input or torch.LongTensor()
+   self._input = self._input and self._input:long() or torch.LongTensor()
 
    if not self.shouldScaleGradByFreq then
       self.shouldScaleGradByFreq = false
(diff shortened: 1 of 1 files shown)


apaszke:
@eulerreich Is that really needed? I think it's quite reasonable that you need to ship the input tensor to the GPU before calling forward - just like with any other nn module.
`nn.LookupTable(5, 6):cuda():forward(torch.ones(5):cuda())` works just fine for me.

soumith:
Indeed. As Adam said, this is not a bug at all!

eulerreich:
The bug, at least on my machine, is from LookupTable:makeInputContiguous. If `self._input` is not LongTensor but instead CudaTensor, then with

```
self._input:resize(input:size()):copy(input)
```

in line 42, `self._input` retains the type as CudaTensor regardless of the type of `input`, and as a result, line 55

```
self.output:index(self.weight, 1, input:view(-1))
```

throws an error because input is not a LongTensor.

I made this patch the day I updated my torch distro and all the torch packages to the most current commit. Were your distro updated as well?

apaszke:
Yes, I have just reinstalled everything to make sure I'm up to date and it still works for me (ofc with one more `:cuda()` call). Also, indexing with a CUDA tensor should still work (try `torch.randn(5,5):cuda():index(1, torch.CudaTensor({1, 3}):view(-1))`)

PR #723, opened 2016-03-19
Revert " Adding table input support for batched SparseLinear, implementing gradInput correctly, fixing other bugs"

zhangxiangxiao:
Reverts torch/nn#698

The new SparseLinear module breaks projects in non-batch mode that depends on the fast implementation of updateParameters using self.lastInput. It includes some implementations of memory networks, word embedding and ranking models with HogWILD parallelization. See the comment to the original pull request.


soumith:
see my comments on the other PR

PR #728, opened 2016-03-21
Formatted latex equations

elbamos:
The equations on the criterion page are replaced with images of the equations, properly formatted, using online latex-to-image conversion services.  The equations are also made to use more consistent conventions. 

I think this makes some of the more advanced criteria a lot easier to understand.

### DIFF
--- a/doc/criterion.md
+++ b/doc/criterion.md
@@ -74,10 +74,7 @@ criterion = nn.AbsCriterion()
 ```
 
 Creates a criterion that measures the mean absolute value of the element-wise difference between input `x` and target `y`:
-
-```lua
-loss(x, y)  = 1/n \sum |x_i - y_i|
-```
+![loss(x, y)  = 1/n \sum |x_i - y_i|](https://latex.codecogs.com/gif.latex?loss%28x%2C%20y%29%20%3D%201/n%20%5Csum%20%7Cx_i%20-%20y_i%7C)
 
 If `x` and `y` are `d`-dimensional `Tensor`s with a total of `n` elements, the sum operation still operates over all the elements, and divides by `n`.
 
@@ -151,16 +148,11 @@ This criterion expect a class index (1 to the number of class) as `target` when 
 
 The loss can be described as:
 
-```lua
-loss(x, class) = -log(exp(x[class]) / (\sum_j exp(x[j])))
-               = -x[class] + log(\sum_j exp(x[j]))
-```
+![loss(x, target) = 1/n \sum(target_i * (log(target_i) - x_i))](https://latex.codecogs.com/gif.latex?loss%28x%2C%20class%29%20%3D%20-%5Clog%5Cfrac%7Be%5E%7Bx_%7Bclass%7D%7D%7D%7B%5Csum_j%20e%5E%7Bx_j%7D%7D%5C%5C%20%3D%20-x_%7Bclass%7D%20&plus;%20%5Clog%5Csum_j%20e%5E%7Bx_j%7D)
 
 or in the case of the `weights` argument being specified:
 
-```lua
-loss(x, class) = weights[class] * (-x[class] + log(\sum_j exp(x[j])))
-```
+![loss(o, t) = - 1/n sum_i weights[i] * (t[i] * log(o[i]) + (1 - t[i]) * log(1 - o[i]))](https://latex.codecogs.com/gif.latex?loss%28x%2C%20class%29%20%3D%20weights_%7Bclass%7D%20*%20%28-x_%7Bclass%7D%20&plus;%20%5Clog%28%5Csum_j%20e%5E%7Bx_j%7D%29%29)
 
 The losses are averaged across observations for each minibatch.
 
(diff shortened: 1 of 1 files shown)


soumith:
Before you go further ahead, I just want to make sure you know that this will definitely break the equations in the command-line help that we provide.

For example:

```
> th

  ______             __   |  Torch7
 /_  __/__  ________/ /   |  Scientific computing for Lua.
  / / / _ \/ __/ __/ _ \  |  Type ? for help
 /_/  \___/_/  \__/_//_/  |  https://github.com/torch
                          |  http://torch.ch

th> require 'nn';
                                                                      [0.0469s]
th> ?nn.AbsCriterion
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
AbsCriterion

 criterion = nn.AbsCriterion()

Creates a criterion that measures the mean absolute value of the element-wise
difference between input x and target y :
 loss(x, y)  = 1/n \sum |x_i - y_i|

If x and y are d -dimensional Tensor s with a total of n elements,
the sum operation still operates over all the elements, and divides by
n .

The division by n can be avoided if one sets the internal variable
sizeAverage to false :
 criterion = nn.AbsCriterion()
criterion.sizeAverage = false
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
                                                                      [0.5892s]
th>
```

elbamos:
Thanks for the heads-up. I didn't even realize there was online help.  Can you point me to the specification for what markdown the online help system can handle?

soumith:
it's based on sundown and uses these bindings: https://github.com/torch/sundown-ffi
It uses the sd.renderASCII function: https://github.com/torch/sundown-ffi#ascii-markdown-pretty-print

elbamos:
The alternate text shows properly in sundown:asciiRender, at least cut-and-pasted; it's prefixed by "image" and the color is different than it was, but that seems like a fair tradeoff to make the equations more readable on the web.

soumith:
Cool. This looks good, ready for merge, just squash your commits.

Atcold:
What about using SVG instead? GIFs are so sad...
Moreover there are some problems with the LaTeX: I see everything in italic, whereas function names and some subscripts should be in `\mathremove`.

elbamos:
@soumith Squashed -- should be good to go. 

@atcold This works using a pair of online services that convert latex to jpg on the fly.  I looked, and don't see any that would convert to svg.

soumith:
Sweet! Thanks

Atcold:
@elbamos, the same online service you used provide different output format. You can choose from the dropdown list below the editing text box.

Atcold:
Hmm... I don't think that the docs look anymore professional... There are different font utilised and the math is not properly rendered. I don't like it :/

Atcold:
There are `log` without the functional escaping `\`, so they look like the product of `l * o * g`, then `loss` is in italic everywhere...

elbamos:
@atcold Let me try to organize...
1. The services didn't like \loss.  L seems like a good solution.   
2. Dot products:  Ok
3. Re: "target" and "weight" -- the equations before weren't consistent in using the same terms for x and y.  I intended to go through and try to conform them.  In my view, "target" should be "y", and "weight" should be "w".  Does that work for you?
4. What do you think we should do about class? The services don't like \class.  
5. When I go back in I'll also take a look about using svg, which I agree would be nicer. 

Thanks!

elbamos:
@Atcold Actually, take a look at what I put up on my fork. I went back in and you were right, I was able to do the formatting you described.  I also tried to fix a few other things that had been bothering me about it.

szagoruyko:
did it break command-line? this is what I get 
screen shot 2016-03-26 at 17 45 56

soumith:
hmm ok that's not good. reverting for now, we can figure out the issues

elbamos:
Yah I agree. Soumith-i tested by running the alternative text through the markdown package you suggested. But I think you mentioned the online help may be using a variant-can you point me toward source? Also lmk if you prefer to discuss here or close this and I can submit again when I've tracked it down. Thanks!

Atcold:
@elbamos, let's move this to Gitter (public or private). I want to help.

If *we* decide to go for LaTeX equations, then it should be done *properly* (correct visualisation of math plus terminal compatibility view). Initially, I believe we discussed this together with @nicholas-leonard (I really fancy his style) and someone else, and we agreed that GitHub's Markdown was not the best for visualising math (and it does not support SVGs either). He had plans to move to a different documentation framework, which would allow the use of [*MathJax*](https://www.mathjax.org/) (check the "View samples" button in the [Features](https://www.mathjax.org/#features) section). An alternative could be *MathML*, which can also be outputted by *MathJax*.

So... Let's have a chat before attempting other "improvements" :wink:

elbamos:
@Atcold Ok - I am on gitter.  I agree with everything you've said.  (SVG does seem to be working - no?)

What seems to be happening in the documentation, is that the documentation engine, when it processes the markdown, it pulls the image from the formatting service, and since it is able to do that successfully, does not display the alternate text.  I have been trying to go through the code for the documentation engine to understand how to work around that.

Atcold:
Sweet finding about SVGs @elbamos, I think this is a new feature of GitHub Markdown implementation.

apaszke:
@Atcold I don't think so. Markdown is simply rendered to HTML, and as far as I know uses `` tags to display images, so as long as browsers can handle SVGs there ([and they all do for quite a while already](http://caniuse.com/#feat=svg-img)), it will all be fine.

PR #729, opened 2016-03-23
Reuse same state tensors on forward/backward

fmassa:
Do not create new `output`/`gradInput` tensors at every function call of `forward`/`backward`, as discussed in https://github.com/torch/nn/pull/712.

`nn.Unsqueeze` is missing in this PR, because we need to change the behaviour of `addSingletonDimension` to accept a result tensor (as done by `torch.view`). Will change both in a separate PR.


fmassa:
FYI, Travis is failing due to an error in `VolumetricMaxPooling` (unrelated to this PR).

PR #733, opened 2016-03-25
Unsqueeze reuse same state tensors

fmassa:
Extend `nn.utils.addSingletonDimension` to accept a result tensor as argument.
Add tests for the new case.


soumith:
Thanks Francisco!

PR #735, opened 2016-03-26
Change display order in VolumetricFullConvolution

kmul00:
Ideally while displaying the *kernel* (and likewise for *padding*, etc.), it should be of the format `kT x kW x kH`, as in the [doc](https://github.com/torch/nn/blob/master/doc/convolution.md#nn.VolumetricFullConvolution) or normal [VolumetricConvolution](https://github.com/kmul00/nn/blob/master/VolumetricConvolution.lua#L184-L194).

Here it was of the format `kW x kH x kT`. This caused confusion to me while creating a model and thought it might to others as well.


soumith:
Thanks!

PR #738, opened 2016-03-27
Removed missing link from documentation

shreyshahi:
The documentation file overview.md has a broken link to `complete list of available modules`. Since no such consolidated list appears in the documentation files, I removed the link. The link may be added back only when we have such a consolidated list of modules in the documentation.


soumith:
Thanks Shrey.

PR #739, opened 2016-03-27
[LookupTable] Add Max-norm constraints to LookupTable

xianjiec:
Add the ability to this module to have a fixed maximum norm, i.e., no word embedding (a row in the module weight) can have [normValue]-norm > maxNorm (e.g., 2-norm > 1). The constraint is enforced in the forward pass (rows that are accessed are re-normalized), so that the output always obeys the max-norm constraint.

### DIFF
--- a/LookupTable.lua
+++ b/LookupTable.lua
@@ -3,12 +3,14 @@ local LookupTable, parent = torch.class('nn.LookupTable', 'nn.Module')
 
 LookupTable.__version = 4
 
-function LookupTable:__init(nIndex, nOutput, paddingValue)
+function LookupTable:__init(nIndex, nOutput, paddingValue, maxNorm, normType)
    parent.__init(self)
 
    self.weight = torch.Tensor(nIndex, nOutput)
    self.gradWeight = torch.Tensor(nIndex, nOutput):zero()
    self.paddingValue = paddingValue or 0
+   self.maxNorm = maxNorm or nil
+   self.normType = normType or nil
 
    self:reset()
 end
@@ -28,8 +30,18 @@ function LookupTable:accUpdateOnly()
 end
 
 function LookupTable:setPadding(paddingValue)
-    self.paddingValue = paddingValue
-    return self
+   self.paddingValue = paddingValue
+   return self
+end
+
+function LookupTable:setMaxNorm(maxNorm)
+   self.maxNorm = maxNorm
+   return self
+end
+
+function LookupTable:setNormType(normType)
+   self.normType = normType
+   return self
 end
 
 function LookupTable:scaleGradByFreq()
@@ -55,6 +67,7 @@ end
 
 function LookupTable:updateOutput(input)
    self:backCompatibility()
+   self:renorm(input)
    input = self:makeInputContiguous(input)
    if input:dim() == 1 then
       self.output:index(self.weight, 1, input)
@@ -77,9 +90,9 @@ function LookupTable:accGradParameters(input, gradOutput, scale)
    end
 
    if not gradOutput:isContiguous() then
-       self._gradOutput = self._gradOutput or gradOutput.new()
-       self._gradOutput:resizeAs(gradOutput):copy(gradOutput)
-       gradOutput = self._gradOutput
+      self._gradOutput = self._gradOutput or gradOutput.new()
+      self._gradOutput:resizeAs(gradOutput):copy(gradOutput)
+      gradOutput = self._gradOutput
    end
 
    self.gradWeight.THNN.LookupTable_accGradParameters(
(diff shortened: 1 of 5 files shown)


LookupTable.lua
```diff
@@ -95,6 +108,25 @@ function LookupTable:accGradParameters(input, gradOutput, scale)
...
+      row_idx = row_idx:view(-1)
+   elseif row_idx:dim() ~= 1 then
+      error("input must be a vector or matrix")
+   end
+   -- "row_idx" and "weight" will be modified in the C code
+   self.weight.THNN.LookupTable_renorm(
+      row_idx:cdata(),
+      self.weight:cdata(),
+      self.maxNorm,
+      self.normValue or 2
```
Tuzongyu: looks like maxNorm and normValue args are swapped compared to the C function definition below.

LookupTable.lua
```diff
@@ -95,6 +108,25 @@ function LookupTable:accGradParameters(input, gradOutput, scale)
    )
 end
 
+function LookupTable:renorm(input)
+   if not self.maxNorm then
+      return
+   end
+   local row_idx = input:clone()
```
colesbury: `:clone()` allocates memory every forward pass. This can cause performance problems on the GPU because the corresponding GC/cuda_free's will cause synchronization points.

Instead, re-use a temporary buffer (like in `makeInputContiguous`)

lib/THNN/generic/LookupTable.c
```diff
@@ -116,4 +116,110 @@ void THNN_(LookupTable_accGradParameters)(
...
+      row_idx[ptr++] = row_idx[i];
+  numel = ptr;
+
+#ifdef _OPENMP
+  if (numel > 1000)
+  {
+    // The strategy is to parallelize over the rows that appear in
+    // row_idx, so that thread 1 handles the rows in row_idx[0..numel/nThreads].
+    // This distributes the work evenly to each thread.
+    #pragma omp parallel private(i)
```
colesbury: Can you use `#pragma omp parallel for` here? I think it should do the same thing (partition the 0..numel over threads) with less code.

lib/THNN/generic/LookupTable.c
```diff
@@ -116,4 +116,110 @@ void THNN_(LookupTable_accGradParameters)(
...
+          long k = row_idx[i] - 1;
+          real *row_data = gw + k*stride;
+          THNN_(LookupTable_renormRow)(gw + k*stride, stride, value, maxNorm);
+        }
+      }
+    }
+    return;
+  }
+#endif
+  for (i=0; i<numel; i++)
```
colesbury: Should this be in a `#else`?

LookupTable.lua
xianjiec: I will do this :+1:

test.lua
```diff
@@ -4209,6 +4209,40 @@ function nntest.LookupTable()
    end
    local err = padw_sum - padw:sum()
    mytester:assertlt(err,precision, 'padding update error ')
+   -- test whether the weights changes accordingly when maxNorm is not nil
+   local all_index = torch.randperm(totalIndex):int()
+   -- input can have duplicates
+   local input = torch.repeatTensor(all_index:clone():narrow(1,1,nIndex), 2):int()
```
colesbury: I think this can just be:

`torch.repeatTensor(all_index:narrow(1,1,nIndex), 2)`

test.lua
```diff
@@ -4209,6 +4209,40 @@ function nntest.LookupTable()
...
+   local maxNorm = math.random()
+   local normValue = torch.Tensor{1,2,math.random()}
```
colesbury: This can just be a table instead of a Tensor and then you can use ipairs e.g.:

`for _, normValue in ipairs{1, 2, math.random()} do`

lib/THNN/generic/LookupTable.c
xianjiec: It is because there is a "return" in line 215. Mainly borrow the style of function "LookupTable_accGradParameters" on line 103.

doc/convolution.md
```diff
@@ -263,6 +263,59 @@ Outputs something like:
 [torch.DoubleTensor of dimension 2x4x3]
 ```
 
+LookupTable supports max-norm regularization. One can activate the max-norm constraints
```
colesbury: You should also add the new parameters above to the line:

`module = nn.LookupTable(nIndex, size, [paddingValue])`

lib/THNN/generic/LookupTable.c
```diff
@@ -116,4 +116,98 @@ void THNN_(LookupTable_accGradParameters)(
...
+      row_idx[ptr++] = row_idx[i];
+  numel = ptr;
+
+#ifdef _OPENMP
+  if (numel > 1000)
+  {
+    // The strategy is to parallelize over the rows that appear in
+    // row_idx, so that thread 1 handles the rows in row_idx[0..numel/nThreads].
+    // This distributes the work evenly to each thread.
+    #pragma omp parallel for
```
colesbury: I think you still need a `private (i)` here since `i` is declared above

xianjiec: Hey, I thought "i" is a private by default (see http://openmp.org/mp-documents/OpenMP_Examples_4.0.1.pdf, page 9/228), but I am adding it to make it sure and clearer :).

colesbury: Ah, cool. I didn't know that.

colesbury:
This LGTM, @soumith do you want to look it over?

soumith:
Cool Stuff! Thanks @xianjiec

PR #682, opened 2016-03-02
Improve error handling

apaszke:
When an error occurs in some module, all containers up to the topmost one will be printed. This should make network debugging much easier.

Error messages look like this after these changes:
```
/Users/adampaszke/torch/install/bin/luajit:
In nn.Sequential (2):
In nn.Sequential (6):
/Users/adampaszke/torch/install/share/lua/5.1/nn/Linear.lua:55: size mismatch, m1: [1 x 10], m2: [20 x 10] at /Users/adampaszke/torch/pkg/torch/lib/TH/generic/THTensorMath.c:706
stack traceback:
        [C]: in function 'error'
        .../adampaszke/torch/install/share/lua/5.1/nn/Container.lua:34: in function 'callAndRethrow'
        ...adampaszke/torch/install/share/lua/5.1/nn/Sequential.lua:45: in function 'forward'
        test.lua:21: in main chunk
        [C]: in function 'dofile'
        ...szke/torch/install/lib/luarocks/rocks/trepl/scm-1/bin/th:145: in main chunk
        [C]: at 0x010e36abc0
```

Because of how this works, top of a stack trace will be a bit malformed (it won't point to where the error happened), but actually you can easily reconstruct an execution path from information that has been printed at the top.

PR #741, opened 2016-03-28
Fix error handling

apaszke:
When using luajit strings have no split method. Fixes error from @fmassa's comment at #682.


soumith:
:split is not defined, but afaik string.split exists. Do you want to simplify this PR knowing that, or is this good?

apaszke:
I think it's simpler to go with this one now, if that's ok with you

colesbury:
FYI string.split (and :split) is from xlua (https://github.com/torch/xlua/blob/master/init.lua#L689). It's not part of LuaJIT or PUC Lua.

@apaszke, can you add tests for the error handling?

apaszke:
@colesbury Sure, I'll add them

Atcold:
Thanks @colesbury, that's what I said as well.

PR #748, opened 2016-04-05
Add nil-guard in getParameters, to avoid crashes when there are no parameters

rsfbarreira:


rsfbarreira:
It appears the build errors are not related to the change (connection errors when doing apt-get during the build process).

PR #749, opened 2016-04-05
Remove unnecessary function override in unpooling modules

apaszke:
Doing so breaks deserialization on systems with different architectures (this gives incompatible bytecode error, because the new function has to be serialized with pooling module).

These operations are cheap, so it's not a problem to perform them even when there's no unpooling associated.


apaszke:
@soumith Any idea why was this performed in this way?

soumith:
i cant remember why it was performed this way, maybe slipped through the cracks.

PR #751, opened 2016-04-06
Use torch.tic instead of os.time for seeding

GeorgOstrovski:
For tracking down flaky tests (bugs/timeouts that occur stochastically depending on randomly chosen inputs) I sometimes start tens or hundreds of runs of the tests at once (within a few seconds). Since os.time() has second precision, this usually gives me only very few really different seeds - with torch.tic() many more different seeds can be generated in a short time window.


soumith:
Thanks!

PR #754, opened 2016-04-07
add argument passing to module typing shortcuts

clement-masson:
Hi all.
Until now, to change the type of modules sharing parameters, you had to do:

```
m1 = nn.Linear(5,5)
m2 = m1:clone('weight', 'gradWeight')

t = {}
newType = 'torch.CudaTensor'
m1:type(newType, t)
m2:type(newType, t)
```

The following code did not work : 

```
t = {}
m1:cuda(t)
m2:cuda(t)
```

I propose to change the shortcut functions signature so that the previous code work.

Clément.

### DIFF
--- a/Module.lua
+++ b/Module.lua
@@ -130,16 +130,16 @@ function Module:type(type, tensorCache)
    return self
 end
 
-function Module:float()
-   return self:type('torch.FloatTensor')
+function Module:float(...)
+   return self:type('torch.FloatTensor',...)
 end
 
-function Module:double()
-   return self:type('torch.DoubleTensor')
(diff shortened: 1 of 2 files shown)


soumith:
it looks like a reasonable change without side-effects. Happy to merge, will you modify the documentation to reflect this please?

clement-masson:
done.

Should I prefer rebasing instead of merging when another commit has been merged since the creation of the PR ?  Or it is ok like that ?

soumith:
There's a squash and merge now in github, so no need of manually squashing :)

PR #755, opened 2016-04-07
Add nil-guard in getParameters, to avoid crashes when there are no parameters

rsfbarreira:
Re-created the pull request to see if the flaky builds pass this time.


soumith:
Thank you!

PR #758, opened 2016-04-10
Added Softened version of SoftMax

kmul00:
Instead of traditional `SoftMax()` which computes `f_i(x) = exp(x_i) / sum_j exp(x_j)`, this module `SoftenedSoftMax()` (couldn't come up with a better name, suggestions are welcome) computes :
`f_i(x) = exp(x_i / T) / sum_j exp(x_j / T)`, 

where `T ( > 0 )` is the *Temperature* applied to *soften* the outputs.

If `T = 1`, it is the same as `SoftMax()`.

This helps in generating *soft targets* and can be used for knowledge distillation, as explained by Hinton *et al.* over [here](http://www.ttic.edu/dl/dark14.pdf).

new file:   SoftenedSoftMax.lua
modified:   init.lua
modified:   test.lua


kmul00:
Now that I come to think about it, it seems that this achieves the same thing as having `nn.MulConstant()` followed by a `nn.SoftMax()`.

That is, for *Temperature* `T` :
```lua
model = nn.Sequential():add(nn.SoftenedSoftMax(T))
```
is equivalent to :
```lua
model = nn.Sequential():add(nn.MulConstant(1/T):add(nn.SoftMax())
```
Then you can close this PR, if deemed unnecessary.

soumith:
if you want, i can merge this into nnx https://github.com/clementfarabet/lua---nnx

PR #759, opened 2016-04-11
fix SelectTable variable size gradInput bug

nicholas-leonard:
Hi guys,

This PR fixes a bug raised in https://github.com/Element-Research/rnn/issues/204 where the gradInput of SelectTable becomes corrupted when the size of the input table is increased from one forward to the next.


soumith:
Thanks nick!

PR #760, opened 2016-04-11
Add P(arametric)Sigmoid layer

Kaixhin:
Implements a (steepness) parameterised logistic sigmoid layer. See https://github.com/Element-Research/rnn/issues/192 for motivation.


PSigmoid.lua
```diff
@@ -0,0 +1,18 @@
+local PSigmoid, parent = torch.class('nn.PSigmoid','nn.Module')
+
+function PSigmoid:__init(k)
+   parent.__init(self)
+   self.k = k or 1
+end
+
+function PSigmoid:updateOutput(input)
+   self.output:resizeAs(input):copy(input)
+   self.output:mul(-self.a):exp():add(1):cinv()
```
kmul00: I think this should be `self.k` instead of `self.a`.

fmassa:
This is equivalent to having
```lua
PSigmoid = nn.Sequential():add(nn.MulConstant(k)):add(nn.Sigmoid())
```
right ? (and this format is probably faster due to less kernel calls as well)

PSigmoid.lua
Kaixhin: My bad - fixed.

Kaixhin:
@fmassa good point - and I imagine that would be faster as well. Plus it allows for optimisations like `cudnn.convert` to be used. I'll just double check that @nicholas-leonard is happy with building a parametric sigmoid that way, and then we can close this.

fmassa:
One other point, to avoid using extra memory, one can set the `inplace` flag to `nn.MulConstant`.

PR #763, opened 2016-04-12
fix for nn.Index backprop

syhw:
:backward() was not working when having a nested nn.Index() as there was no gradInput corresponding to the second argument (== the index).


soumith:
fixed via https://github.com/torch/nn/commit/1248424112750af551c5289ac422ae84cf066d61

PR #765, opened 2016-04-13
delete an assertion in batchnormalization

yangky11:
Hi all,

This assertion is not really necessary and it prevents doing back-propagation to input (computing the derivative of loss w.r.t. the input image).


soumith:
Thanks!

PR #767, opened 2016-04-13
Normalize supports arbitrary dimensions

fmassa:
Adds docs + tests.
Also, cleans a bit some unnecessary temporary variables.

**Important remark**: I added the dimension parameter as the second argument to the constructor, making the `eps` the third one. This breaks code that changed the default `eps` during initialization, but I felt that `dim` coming after `eps` felt weird. Let me know if you prefer to let `dim` the last argument.

### DIFF
--- a/Normalize.lua
+++ b/Normalize.lua
@@ -1,25 +1,26 @@
 local Normalize, parent = torch.class('nn.Normalize', 'nn.Module')
 
-function Normalize:__init(p,eps)
+function Normalize:__init(p, dim, eps)
   parent.__init(self)
   assert(p,'p-norm not provided')
   assert(p > 0, p..'-norm not supported')
   self.p = p
+  self.dim = dim or -1
   self.eps = eps or 1e-10
 end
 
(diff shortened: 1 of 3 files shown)


apaszke:
@fmassa I think if you're changing the order of arguments like here, it's worth adding a check that `dim` is an integer. This would at least throw an error if somebody used the `eps` argument.

fmassa:
@apaszke I agree. just pushed a diff checking that `dim` is integer.

fmassa:
I forgot to consider that previously saved models do not have the `dim`field. I'm pushing a patch

fmassa:
Fixed case of old saved models. To make it simple I just added the default value in the forward/backward calls, I didn't overwrite the `read` method.

fmassa:
One note about the negative indexes, if you want I could also add a `numInputDims` parameter, as in [`nn.Max`](https://github.com/torch/nn/blob/master/Max.lua#L11-L19). Let me know what you think.

PR #768, opened 2016-04-14
BatchNormalization: compute save_mean and save_std in evaluattion mode

yangky11:
Hi,

My previously merged PR https://github.com/torch/nn/pull/765 causes an error.
The `backward` of batchnormalization requires variables `save_mean` and `save_std` be computed during the forward pass. 
But, previously those 2 variables are only computed when the layer is in training mode, so that doing back-prop in evaluation mode will cause `nan`

### DIFF
--- a/BatchNormalization.lua
+++ b/BatchNormalization.lua
@@ -125,7 +125,6 @@ end
 local function backward(self, input, gradOutput, scale, gradInput, gradWeight, gradBias)
    self:checkInputDim(input)
    self:checkInputDim(gradOutput)
-   assert(self.train == true, 'should be in training mode when self.train is true')
    assert(self.save_mean and self.save_std, 'must call :updateOutput() first')
 
    input, gradOutput = makeContiguous(self, input, gradOutput)
@@ -142,9 +141,13 @@ local function backward(self, input, gradOutput, scale, gradInput, gradWeight, g
       THNN.optionalTensor(gradWeight),
       THNN.optionalTensor(gradBias),
       THNN.optionalTensor(self.weight),
+      self.running_mean:cdata(),
+      self.running_var:cdata(),
       self.save_mean:cdata(),
       self.save_std:cdata(),
-      scale)
+      self.train,
+      scale,
+      self.eps)
 
    return self.gradInput
 end
--- a/doc/testing.md
+++ b/doc/testing.md
@@ -1,5 +1,69 @@
-## Testing ##
+# Testing #
 For those who want to implement their own modules, we suggest using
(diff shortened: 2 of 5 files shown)


yangky11:
I'm confused why the back-prop of batchnormalization uses `save_mean` and `save_std`.
Shouldn't we use `running_mean` and `running_std`? As they are used to normalize the input in the forward pass.

soumith:
cc: @colesbury 
I knew I missed something in #765 , now I know.

colesbury:
@yangky11, `save_mean` and `save_std` exists so that the backward pass can re-use some results computed in the forward pass. I'm not sure this is correct. The forward algorithm is different in training and evaluate mode, which means that the gradInput computation should probably be different.

yangky11:
@colesbury 
Ahh, thanks, I figured it out.  `save_mean` and `save_std` are mean and std within a mini-batch, whereas `running_mean` and `running_var` are across multiple batches.
In training mode, the input is normalized by `save_mean` and `save_std`.
In evaluation mode, the input is normalized by `running_mean` and `running_var`.
So, when doing back-prop in evaluation mode, there is actually no need to use `save_mean` and `save_std`, which is good since we do not have to pay for the additional overhead of computing them during the forward pass.
I have modified the code. Could you please have a look?

colesbury:
I think this looks correct. Could you add gradient checks in test.lua for evaluate mode? You'll also need to modify the CUNN implementation.

yangky11:
There were still a little more subtlties. But anyway, I finally made it pass `th -lnn -e "nn.test{'BatchNormalization'}"`, which contains new tests for batchnormalization layers in evaluation mode.

soumith:
this needs a rebase, after that it's good to merge.

yangky11:
rebased and squashed

soumith:
it still has conflicts, i'll manually merge it i guess.

soumith:
pushed via https://github.com/torch/nn/commit/26a5a7e504dc6e082cf8610e5c3c581e9649f53f

soumith:
Thanks Kaiyu!

hughperkins:
Hmmmmm :-(   https://github.com/hughperkins/clnn/issues/39

hughperkins:
Actually, I dont think this pr is the root cause of https://github.com/hughperkins/clnn/issues/39 . Though it does look like something that will break OpenCL Torch.  But maybe I already fixed it??? Or maybe my tests for BN are insufficient.  Anyway, I shall post my 'hmmm' in the pr responsible for https://github.com/hughperkins/clnn/issues/39

PR #769, opened 2016-04-14
adding back assertion until a proper backward is added

soumith:


jonathantompson:
That was a quick code review :-)

soumith:
it's a revert of https://github.com/torch/nn/pull/765

Issue #675, opened 2016-02-27
nn.Concat changes the type of output, and breaks when receiving empty gradOutput


iamalbert:
``` lua
model = nn.Sequential()                                                         
  :add( nn.Concat(1)                                                            
        :add(nn.Identity())                                                     
        :add(nn.Identity())                                                     
  )                                                                             
  :add( nn.LookupTable(10,3) )                                                  
                                                                                
                                                                                
input = torch.LongTensor{1,2,3}                                                 
model:forward( input )                  
model:backward(input, torch.rand(6,3) )
```
First, the `:forward` call gives the following error
```
Concat.lua:65: bad argument #1 to 'resizeAs' (torch.DoubleTensor expected, got torch.LongTensor)
```
It's because `nn.Concat` assumes all the outputs from its modules are of the default type, and it crashes when they are not. (Actually there are some other modules, say `nn.Add`, does the same thing)

Second, after we change the `input` to `torch.DoubleTensor{1,2,3}` the `:forward` succeed, but `:backward` breaks at `nn.Concat` again.
```
Concat.lua:70: bad argument #1 to 'narrow' (out of range at /tmp/luarocks_torch-scm-1-8725/torch7/lib/TH/generic/THTensor.c:349)
```
It looks like that `nn.Concat` does not handle the situation when its succeeding module stops the back propagation. Similar 2 problems occurs in `nn.JoinTable`.

https://github.com/torch/nn/blob/master/Concat.lua#L70
https://github.com/torch/nn/blob/master/Concat.lua#L37

fmassa:
For the first problem, you will need to add a `nn.Copy` to the beginning of your module to copy from `long` to `double`.
For the second one, the problem is actually inside `nn.LookupTable`, which happened with several people in the past, see https://github.com/torch/nn/issues/568#issuecomment-172016604 and https://github.com/torch/nn/issues/640 for some examples.

@soumith given that this issue with `nn.LookupTable` is happening so often, we should probably think about a way of addressing it. Adding a note on the documentation ? reshaping and filling a dummy `gradInput` ? making its `updateGradInput` an empty function ?

soumith:
@fmassa returning a dummy zero-filled gradInput sounds like a good idea.

fmassa:
This issue was addressed by #772

PR #772, opened 2016-04-16
Add dummy updateGradInput in LookupTable

fmassa:
See https://github.com/torch/nn/issues/675#issuecomment-210669866 for the motivation for this PR.
Also adds some more buffers to `clearState`

**Notes:**

As there is no constraint on the type of the input tensor to `LookupTable` (its data is copied to `self._input` if the type is not the right one), and as its more natural that `gradInput` and `input` have the same types, I allocate new tensors everytime the user changes the input type (and not the module type).
I'm not sure about this decision. Do you prefer to let `self.gradInput` to have the same type as `self.weight` ?

Let me know if you have remarks about those changes, as I have never used it I might be screwing up something.

### DIFF
--- a/LookupTable.lua
+++ b/LookupTable.lua
@@ -80,6 +80,19 @@ function LookupTable:updateOutput(input)
    return self.output
 end
 
+function LookupTable:updateGradInput(input, gradOutput)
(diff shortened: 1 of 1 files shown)


iamalbert:
This is cool

PR #776, opened 2016-04-18
add noBias for nn.Linear and nn.SpatialConvolution

yangky11:
Hi,

This is to allow `nn.Linear` and `nn.SpatialConvolution` to have no bias, just like [cudnn.torch](https://github.com/soumith/cudnn.torch).

We can revert https://github.com/soumith/cudnn.torch/pull/164 if this is merged.

@fmassa

### DIFF
--- a/Jacobian.lua
+++ b/Jacobian.lua
@@ -187,7 +187,9 @@ function nn.Jacobian.linearModuleDiagHessian(module, input, gradParamName)
       gradOutput1D:zero()
       gradOutput1D[i] = 1
       module.gradWeight:zero()
-      module.gradBias:zero()
+      if module.bias then
+         module.gradBias:zero()
+      end
       module:updateGradInput(input, gradOutput)
       module:accGradParameters(input, gradOutput)
       diagHessian:addcmul(gradParam, gradParam)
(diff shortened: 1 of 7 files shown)


yangky11:
Somehow a test on `VolumetricMaxPooling` failed, but it seems has nothing to do with the new changes.

fmassa:
Current implementation allocates new memory at each forward/backward if the no-bias option is set.
I have started the no-bias option some time ago, but I was wanting to factorize the convolution into smaller blocks as well and I didn't finish it. It's here if you want to have a look https://github.com/fmassa/nn/commit/6a6f5f155edd1e3aa943ddb0f0f91ee79d9d5257

yangky11:
Thanks, I just integrated your code into this PR so that it does not allocate new memories, added some tests, and removed the unnecessary argument `bias` in `updateGradInput`

fmassa:
Cool!
Could you do the same for cunn (as is cunn won't work) ? I also have the changes somewhere in my computer, just need to check where they are.

yangky11:
I'm working on it

soumith:
Thanks Kaiyu!

PR #777, opened 2016-04-18
Add BatchNormalization feature size check

szagoruyko:
Much needed because CPU version will show a cryptic error and GPU will just proceed with undefined behaviour.

### DIFF
--- a/BatchNormalization.lua
+++ b/BatchNormalization.lua
@@ -75,8 +75,11 @@ end
 
 function BN:checkInputDim(input)
    assert(input:dim() == self.nDim, string.format(
-     'only mini-batch supported (%dD tensor), got %dD tensor instead',
-     self.nDim, input:dim()))
+      'only mini-batch supported (%dD tensor), got %dD tensor instead',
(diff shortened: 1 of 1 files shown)


BatchNormalization.lua
```diff
@@ -75,8 +75,11 @@ end
 
 function BN:checkInputDim(input)
    assert(input:dim() == self.nDim, string.format(
-     'only mini-batch supported (%dD tensor), got %dD tensor instead',
-     self.nDim, input:dim()))
+      'only mini-batch supported (%dD tensor), got %dD tensor instead',
+      self.nDim, input:dim()))
+   assert(input:size(2) == self.running_mean:nElement(), string.format(
+      'got %d-feature tensor, expected %d',
+      input:size(2), self.running_mean:nElement(1)))
```
fmassa: `self.running_mean:nElement()`, right ? no need of the `1` in there.

szagoruyko: thanks, fixed

PR #782, opened 2016-04-21
fix MM after :clearState() has been called

### LINKED ISSUE #781: crash in MM.backward after :clearState()
We are experiencing a crash in the backward pass of MM after calling :clearState on the network.

```
/data/vision/torralba/commonsense/torch/distro/install/bin/luajit: ...commonsense/torch/distro/install/share/lua/5.1/nn/MM.lua:51: attempt to index a nil value
stack traceback:
        ...commonsense/torch/distro/install/share/lua/5.1/nn/MM.lua:51: in function 'updateGradInput'
        ...onsense/torch/distro/install/share/lua/5.1/nn/Module.lua:30: in function 'backward'
        debug_clearstate.lua:17: in main chunk
 ```

Minimal code to reproduce:

```lua
require 'nn'
require 'cunn'

mod = nn.MM()
mod:cuda()

a = torch.randn(10,5,5):cuda()
b = torch.randn(10,5,5):cuda()
grad = torch.randn(10,5,5):cuda()

out = mod:forward({a,b})
mod:backward({a,b}, grad)

mod:clearState()

out = mod:forward({a,b})
mod:backward({a,b}, grad)  --- CRASH!
```

If you change the nn.MM to something else, such as nn.CAddTable(), there is no crash.

The code works fine until :clearState() is called. The crash happens with/without cunn/cuda.

Thank you very much!

cvondrick:
fixes #781 

I am not super familiar with internals of Torch, so I hope this change does not cause unintended consequences.

### DIFF
--- a/MM.lua
+++ b/MM.lua
@@ -46,6 +46,9 @@ function MM:updateOutput(input)
 end
 
 function MM:updateGradInput(input, gradOutput)
+  self.gradInput[1] = self.gradInput[1] or input[1].new()
+  self.gradInput[2] = self.gradInput[2] or input[2].new()
+
   assert(#input == 2, 'input must be a pair of tensors')
   local a, b = table.unpack(input)
   self.gradInput[1]:resizeAs(a)


MM.lua
```diff
@@ -87,3 +87,9 @@ function MM:updateGradInput(input, gradOutput)
 
   return self.gradInput
 end
+
+function MM:clearState()
+   local retval = parent.clearState(self)
+   self.gradInput = {torch.Tensor(), torch.Tensor()} -- re-initialize
```
soumith: this should not be torch.Tensor() but should be torch[self._type].new()

cvondrick: thanks! torch[self._type] didn't work because self._type was "torch.FloatTensor" or "torch.CudaTensor", when it needed to be "FloatTensor" or "CudaTensor"

in the end, I went a different way

soumith:
Awesome! Thank you.

PR #784, opened 2016-04-22
Adding hyper-parameter lambda to GradientReversal module

ccsasuke:
Added the hyper-parameter lambda for the Gradient Reversal Layer as described in the original paper.

This allows the user to control how much gradient one would like to propagate back to the preceding layers, and it is also possible to set different values for lambda at various stages during training, which is also adopted in the original paper.


soumith:
This has back-compatibility issues that I'll fix in a second.

PR #789, opened 2016-04-25
Add correctness test on SparseLinear in multithreaded case (and remov…

nkoumchatzky:
…e trailing whitespaces).

Actually, I'm testing that this test works in the OpenMP + BLAS setup. It works with OMP_NUM_THREADS=1 and MKL, fails with OMP_NUM_THREADS > 1 and MKL, just want to test the CI build here.

If you want to convince yourself the test should work, run it with OMP_NUM_THREADS=1 at home!

### DIFF
--- a/test.lua
+++ b/test.lua
@@ -757,7 +757,7 @@ function nntest.Linear()
             local err = jac.testJacobianParameters(module, input, module.bias, module.gradBias)
             mytester:assertlt(err,precision, 'error on bias ')
          end
-         
+
          local err = jac.testJacobianUpdateParameters(module, input, module.weight)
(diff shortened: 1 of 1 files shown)


nkoumchatzky:
Ok, it looks like all the tests fail with OpenMP + OpenBlas. This is likely due to the accGradParameters function using OpenMP at the higher loop level, which allows for concurrent unprotected gradWeight updates if the same indices are found within the input. It seems it was wrong before too.

soumith:
cc: @ebetica

soumith:
there are now correctness tests and fixes via https://github.com/torch/nn/pull/811/files

nkoumchatzky:
Thanks @soumith!

PR #791, opened 2016-04-26
Add SpatialClassNLLCriterion

apaszke:
A CPU counterpart for torch/cunn#250


apaszke:
One thing I'm not 100% sure about is `omp` pragma in `updateGradInput`. Are both loop counters private in the way I wrote it?

lib/THNN/generic/SpatialClassNLLCriterion.c
```diff
@@ -0,0 +1,123 @@
...
+  real *gradInput_data = THTensor_(data)(gradInput);
+
+  long batch_size = THTensor_(size)(input, 0);
+  long n_classes = THTensor_(size)(input, 1);
+  long map_size = THTensor_(size)(input, 2) * THTensor_(size)(input, 3);
+  long sample_size = map_size * n_classes;
+
+  real normalize = sizeAverage ? *total_weight_data : 1.0f;
+
+#pragma omp parallel for
```
soumith: this should be: omp parallel for private(b)

fmassa: I don't know much about omp and stuff, but I thought that the `private (b)` was only needed if `b` was defined outside of the scope of the for loop (a la C89). But I might be completely wrong.

apaszke: Yeah, that's what I understood as well (I.e. every variable declared in parallel block is private to the thread). Initially I had private(b, elem) to be sure, but GCC crashed on undefined references there if I didn't move the declarations above the loops.

soumith: you guys are right. sorry, my bad.

PR #796, opened 2016-04-27
MultiLabelMarginCriterion fixes for CUDA

soumith:
also, __tostring added in SelectTable, fixes for nn.Sum when gradOutput is not contiguous

### DIFF
--- a/Sum.lua
+++ b/Sum.lua
@@ -41,6 +41,11 @@ function Sum:updateGradInput(input, gradOutput)
     -- Instead, do a deepcopy
     local size      = input:size()
     size[dimension] = 1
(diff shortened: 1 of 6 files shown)


Sum.lua
```diff
@@ -49,3 +54,8 @@ function Sum:updateGradInput(input, gradOutput)
     end
     return self.gradInput
 end
+
+function Sum:clearState()
+    self._gradOutput = nil
+    parent.clearState(self)
```
szagoruyko: should be:
```lua
nn.utils.clear(self, '_gradOutput')
return parent.clearState(self)
```

soumith: fixed.

PR #800, opened 2016-04-30
nn.Select support for negative indices

JoostvDoorn:
As nn.SelectTable support negative inputs it seems only natural to also allow nn.Select to support the same.

Just a side note: nn.SelectTable has asserts, should nn.Select have them as well?

I think it's also fine to support this in the Tensor:select function instead.


soumith:
the asserts are not absolutely necessary for this, as the tensor's select function does the assertions.

PR #808, opened 2016-05-06
Adding an optional extra input to SpatialFull- and VolumetricFull- convolution.

fbesse:
This can used to dynamically set the output size, as an alternative to using the adj terms, which has the downside to be fixed at construction time and makes the output size of these modules depend on the size of the input tensor. This is convenient for conv->rev-conv networks, as the input of the conv can be used as extra input to the rev-conv to go back to the same size exactly.

There is no downside to using the extra input, as there is effectively no gradient flowing back from this input, since this commit in nngraph: https://github.com/torch/nngraph/commit/2c66c0d9ceed8fdfd4f0230a4d0d38f6d9fbcee0
which detects that gradOutput is an expanded zero tensor and skips it.


soumith:
Thanks Fred!

PR #809, opened 2016-05-06
Add documentation for nn.Transpose

juesato:
I've found this module useful, and thought it could be helpful to make it easier to find.


soumith:
Thank You!

PR #810, opened 2016-05-07
add docs for nn.Log, nn.AddConstant & nn.MulConstant

yangky11:
Hi all,

These 2 modules simply add/multiply the input tensor by a constant.  

I was trying to use [dcgan](https://github.com/soumith/dcgan.torch) to generate images that are fed into another model in the pipeline. The problem is that the output of dcgan has to be shifted from `[-1,+1]` to `[0,1]` before further processing. Thus I think simple modules to add/multiply constant might be useful.

### DIFF
--- a/Log.lua
+++ b/Log.lua
@@ -1,6 +1,6 @@
 local Log, parent = torch.class('nn.Log', 'nn.Module')
 
-function Log:__init(inputSize)
+function Log:__init()
    parent.__init(self)
 end
 
--- a/doc/simple.md
+++ b/doc/simple.md
@@ -26,11 +26,14 @@ Simple Modules are used for various tasks like adapting Tensor methods and provi
     * [Squeeze](#nn.Squeeze) : [squeezes](https://github.com/torch/torch7/blob/master/doc/tensor.md#tensor-squeezedim) the input;
     * [Unsqueeze](#nn.Unsqueeze) : unsqueeze the input, i.e., insert singleton dimension;  
   * Modules that adapt mathematical Tensor methods :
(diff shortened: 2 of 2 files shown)


fmassa:
This is equivalent to [`nn.AddConstant`](https://github.com/torch/nn/blob/master/AddConstant.lua) and [`nn.MulConstant`](https://github.com/torch/nn/blob/master/MulConstant.lua), right ?

yangky11:
Yes,   I just found them....  
So I add some docs for them

fmassa:
They have docs, but maybe they are not in the best place... https://github.com/torch/nn/blob/master/doc/transfer.md#addconstant

yangky11:
Well,  then let's just add the docs for `nn.Log`

fmassa:
Sure !
And I also wonder if the docs for `AddConstant` and `MulConstant` shouldn't be transferred to *Simple Layers*, what do you think ?

yangky11:
I think it would be good to have some reference of `AddConstant` and `MulConstant` in `simple.md`.
Because when I wanted to find these modules, my first thought was that maybe they are in `simple.md`. 
What about add them in the content of `simple.md` pointing to the real docs in `transfer.md`? something like

* [AddConstant](https://github.com/torch/torch7/blob/master/doc/transfer.md#nn.AddConstant) : add a constant to the input tensor;

fmassa:
I think that's a good idea. Could you send a PR ?

yangky11:
sure!

yangky11:
@soumith  ready to be merged

soumith:
Thank You!

Issue #722, opened 2016-03-18
SparseLinear has a race condition


ebetica:
Check out the parallelism [here](https://github.com/torch/nn/blob/master/lib/THNN/generic/SparseLinear.c#L65).

During W*X, for every non-zero element in X the i-th column of W is added to the j-th row of the output, but this is encased in a omp parallel for, so something like 
`
parallel for (i,j), v in X do
  output[j, {}] += v * W[{}, i]
`
However, we shouldn't be adding two j's at the same time, or we'll run into a race condition. Currently SparseLinear will behave this way, I'm writing a patch for it. My [CUDA implementation](https://github.com/torch/cunn/pull/223) shouldn't run into this issue if it gets merged.

PR #811, opened 2016-05-10
Fixing sparse linear race condition

### LINKED ISSUE #790: SparseLinear:accGradParameters wrong multithreaded behavior?
#789 is a test that passes with OMP_NUM_THREADS=1, but not otherwise with OMP + OpenBlas (or MKL, for that matter). The OMP statement in the SparseLinear.c accGradParameters function makes concurrent updates of gradWeight possible if the same index occurs multiple times in the input. The legacy code should be wrong too. Anything I missed?

EDIT: the behavior we get out of this will make the algo converge eventually as it amounts to doing "hogwild" updates. But it should at least be an option.

ebetica:
Fixes #790 and #722

### DIFF
--- a/SparseLinear.lua
+++ b/SparseLinear.lua
@@ -130,8 +130,13 @@ function SparseLinear:accGradParameters(input, gradOutput, scale)
          gradOutput:resize(1, gradOutput:size(1))
       end
 
+      local rows = self.formatted_input:select(2, 1)
+      local cols = self.formatted_input:select(2, 2)
+      local sortinds = cols * gradOutput:size(1) + rows
+      local _, inds = sortinds:sort(1, false)
+      local newinput = self.formatted_input:index(1, inds)
       input[1].THNN.SparseLinear_accGradParameters(
-         self.formatted_input:cdata(),
+         newinput:cdata(),
          gradOutput:cdata(),
          self.gradWeight:cdata(),
          self.gradBias:cdata(),
--- a/lib/THNN/generic/SparseLinear.c
+++ b/lib/THNN/generic/SparseLinear.c
@@ -48,7 +48,7 @@ void THNN_(SparseLinear_updateOutput)(
           THTensor *weight,
           THTensor *bias)
 {
-  long h, i;
+  long h, i, j, hp0, hp1;
   long outDim = THTensor_(size)(weight, 0);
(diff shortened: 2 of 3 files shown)


soumith:
Thanks Zeming!

lib/THNN/generic/SparseLinear.c
```diff
@@ -59,25 +59,43 @@ void THNN_(SparseLinear_updateOutput)(
 
   long nnz = THTensor_(size)(input, 0);
 
+  THLongTensor * csr = THLongTensor_newWithSize1d(batchSize+1);
```
fmassa: Isn't this tensor leaking ?

SparseLinear.lua
```diff
@@ -130,8 +130,13 @@ function SparseLinear:accGradParameters(input, gradOutput, scale)
          gradOutput:resize(1, gradOutput:size(1))
       end
 
+      local rows = self.formatted_input:select(2, 1)
+      local cols = self.formatted_input:select(2, 2)
+      local sortinds = cols * gradOutput:size(1) + rows
```
fmassa: This line and the following 2 allocates memory at each call

lib/THNN/generic/SparseLinear.c
```diff
@@ -165,26 +183,42 @@ void THNN_(SparseLinear_accGradParameters)(
              "gradOutput must be contiguous");
 
   long nnz = THTensor_(size)(input, 0);
-  // THTensor_(resize2d)(gradOutput, batchSize, outDim);
 
-  // gradWeight += gradOutput * input
-#pragma omp parallel for private(h, i) schedule(static) if (\
-  nnz * outDim > 10000)
+  THLongTensor* csc = THLongTensor_newWithSize1d(inDim+1);
```
fmassa: This tensor needs to be freed as well, right?

fmassa:
Hi,
Sorry for my lack of reactivity those past few weeks, I've been quite busy lately.
I made some inline comments.
Thanks for the fix !

lib/THNN/generic/SparseLinear.c
soumith: yea both are n00b leaks. I'm fixing them. cc: @ebetica

Issue #814, opened 2016-05-11
clearState destroy input


iamalbert:
```lua
 require 'nn'                  

                               
 local a = torch.rand(2, 5)    
                               
 local m = nn.SelectTable(1)   
 local out = m:forward({a})                                  
 print(a)                 

 m:clearState()                
 print(a)                 
```
outputs
```
 0.6709  0.2496  0.9766  0.6969  0.3582
 0.6664  0.3344  0.4896  0.4498  0.9036
[torch.DoubleTensor of size 2x5]

[torch.DoubleTensor with no dimension]
```

this bug is killing me ..

kmul00:
Hi,

This is not a *bug* per-say, I think. 

If you look at [this](https://github.com/torch/nn/blob/master/SelectTable.lua#L14) line, in the `forward` pass of `SelectTable`, you will see that it performs *memory-copy* of the *input* to the *output*. So when you perform `clearState`, it clears the *output*, which is nothing but the pointer to the *input*. This is done keeping in mind the efficient memory management of Torch.

To get around it, you can do something like 
```lua
out = m:forward({a:clone()})
```

iamalbert:
Thanks for you suggestion. However, the clone of input not only requires extra memory , but also impossible to do if `nn.SelectTable` is inside a large network.

I am now using
```lua
nn.SelectTable.clearState = function(self) self.gradInput = {} end
```
as a workaround

fmassa:
I think this deserves a fix for SelectTable.

iamalbert:
I just see that NarrowTable has the same problem too

fmassa:
Hum, wait, this module is like Identity, can handle tensors or tables of tables of tables of tensors. No qo easy then

RuiShu:
How about manually resetting self.output to {} (or torch.Tensor) when calling clearState for those particular modules?

iamalbert:
I agree with @RuiShu

fmassa:
I think the simplest for now is to simply copy the clearState from Identity and add them to SelectTable and NarrowTable. Do some of you want to send a PR ?

soumith:
fixed via #816

PR #816, opened 2016-05-11
prevent SelectTable and NarrowTable clears input

iamalbert:
#814 

copy clear state from Identity to NarrowTable and SelectTable


soumith:
Thanks!

PR #818, opened 2016-05-12
add SpatialWeightedClassNLLCriterion and WeightedSmoothL1Criterion and their cunn correspondence

lim0606:
Hi, 

I am currently working on re-implementing faste-rcnn (http://arxiv.org/abs/1506.01497) and instance-aware semantic segmentation (http://arxiv.org/abs/1512.04412) via torch.

As a part of the work, I made SpatialWeightedClassNLLCriterion (as well as WeightedClassNLLCriterion) and WeightedSmoothL1Criterion (Their cunn correspondences are also updated in https://github.com/torch/cunn/pull/270). The brief explanation of these modules are explained in the updated doc for `criterion`. 

I'm not sure these modules are properly designed for this repository, but I just want to share them with anyone who might need these. 

Best regards, 

Jaehyun


soumith:
cc: @szagoruyko 

SoftMaxAtTest -- is that module really necessary?
Sergey, wdyt about this PR?

lim0606:
@soumith 

Hi, 

Once I thought it is necessary since the (softmax) normalized values are necessary only in rpn/fast-rcnn test. However, it was solved by modifying the pipeline for testing. 

I couldn't clean up the specific commit for the layer from all the history.

Do you think it is better to PR again after cleaning up? 

Best regards, 

Jaehyun

szagoruyko:
@lim0606 I don't think we need this in `nn`, eg. implementation of Fast-RCNN needs 30 lines of extra code on top of `nn`. Regarding Faster-RCNN, you can do a PR to `imagine-nn`, we will happily review it there.

lim0606:
Okay :)

soumith:
closing based on Sergey's comments.

PR #819, opened 2016-05-12
Add gradInput nil check for SplitTable

asrata:
There are cases when self.gradInput is nil in SplitTable, e.g., when it is the very bottom of the network. A check is necessary to avoid exceptions.


soumith:
Thanks!

PR #824, opened 2016-05-16
replace the StochasticGradient example with optim example

hughperkins:
Address https://github.com/torch/torch7/issues/684

### DIFF
--- a/doc/training.md
+++ b/doc/training.md
@@ -4,102 +4,101 @@
 Training a neural network is easy with a [simple `for` loop](#nn.DoItYourself).
 While doing your own loop provides great flexibility, you might
 want sometimes a quick way of training neural
-networks. [StochasticGradient](#nn.StochasticGradient), a simple class
-which does the job for you is provided as standard.
+networks. [optim](https://github.com/torch/optim) is the standard way of training Torch7 neural networks.
 
-<a name="nn.StochasticGradient.dok"></a>
-## StochasticGradient ##
+`optim` is a quite general optimizer, for minimizing any function that outputs a loss.  In our case, our
+function will be the loss of our network, given an input, and a set of weights.  The goal of training 
+a neural net is to
+optimize the weights to give the lowest loss over our training set of input data.  So, we are going to use optim
+to minimize the loss with respect to the weights, over our training set.  We will feed the data to 
+`optim` in minibatches.  For this particular example, we will use just one minibatch, but in your own training
+you will almost certainly want to break your training set into minibatches, and feed each minibatch to `optim`,
+one by one.
 
-`StochasticGradient` is a high-level class for training [neural networks](#nn.Module), using a stochastic gradient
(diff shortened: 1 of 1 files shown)


soumith:
This is really nicely done! Thanks Hugh!

hughperkins:
Thanks!  Seems there are a few typos and stuff, so I will fix those bit by bit...

apaszke:
It might be a good place to mention that `:getParameters()` shares the memory with network, but it's actually rearranging everything under the hood it, so all old references are invalidated (and it's more like the network's parameters are views onto the parameter vector afterwards). It's often a source of confusion.

hughperkins:
Updated, see: https://github.com/hughperkins/nn/blob/example-tweaks/doc/training.md#nn.traningneuralnet.dok

Atcold:
Now the `optim` doc looks like this:
> # Optimization package
> This package contains several optimization routines and a logger for Torch:
> - Overview;
> - Optimization algorithms;
> - Logger.
You can add a 4th point
 - Example.

Atcold:
@hughperkins, let me know if you don't have time. I can fix this overnight.
I need it for my class...

soumith:
it seems better to teach users how to use optim, rather than nn.StochasticGradient.

Atcold:
Yes, I'm covering that in my next class, as advanced material. Nevertheless, I believe we should restore the legacy documentation and, optionally, move the optim way in the `optim` documentation. We can even specify which is the suggested alternative, but removing doc is unacceptable, in my opinion.

Moreover,  with `StochasticGradient` you can train a network in 2 lines of code. `optim` requires much more code.

hughperkins:
So....
- I made the change, because I was getting lots of issues on clnn where people were training a network, without batching, and were surprised when it didnt work/ran slowly, and they were basically doing stuff in a way quite differently from how eg I see Sergey and so on using it
- you do however make some good points, ie, you say:
  - optim doc should be in 'optim' repo, rather than 'nn' repo
  - probably not ideal to remove old doc
- I guess we can put the optim example into ... optim ... for now, and we can think about how to stretch something over the top somehow?  ie at a higher level than nn/optim, or perhaps just link to optim perhaps?

For your class, presumably you can link temporarily to the old location, ie:

https://github.com/torch/nn/blob/9f64ffa3e20c73ada3ab8c1564eb821e719f0155/doc/training.md#nn.traningneuralnet.dok

... so that we dont need to rush to change things, and your class works.

hughperkins:
What I'll do for now is add the Training using StochasticGradient into the middle.  Is it good?  Well, I'll make a PR, and we can think about it?

hughperkins:
https://github.com/torch/nn/pull/970

hughperkins:
For migrating the optim example from nn to optim:
- optim-side pr: https://github.com/torch/optim/pull/134
- nn-side pr: https://github.com/torch/nn/pull/971

PR #826, opened 2016-05-17
small tweaks to training example

hughperkins:
a bunch of cleaning and tweaks to the training example


hughperkins:
Note: I've split into two commits, second commit moves manual training exapmle to the start, and changes the emphasis slightly: manual training is presented to show low-level concepts of what happens in training, but for normal usage we would use `optim`.

apaszke:
Why do you want to move the manual example to the start? I'd say that for beginners optim + `:forward()` + `backward()` is all they probably need.

apaszke:
But the section on flattening parameters is awesome! 😄

hughperkins:
I was tempted to ditch manual training section altogether https://github.com/torch/nn/issues/827

But actually I think it makes a nice introduction, so they can see how forwards/backwards works.

Then, on to optim

Issue #828, opened 2016-05-20
Small Problem in the Lookup Table Documentation


RishabhMalviya:
I think the matrix multiplication analogy is wrong in the documentation for the nn.LookupTable documentation (https://github.com/torch/nn/blob/master/doc/convolution.md#nn.LookupTable). 

It isn't a major problem, just that the actual output of the forward operation is the transpose of that shown in the documentation.   
`(M P)'`  

It's not a major problem, but if you're nit-picky about mathematical details like me, it gets to you.

nicholas-leonard:
@RishabhMalviya Could you send a PR?

RishabhMalviya:
[Done](https://github.com/torch/nn/pull/831)

BhargavaRamM:
I just want to mention one very small documentation mistake in nn.Bilinear. According to the documentation, the input tensor given in forward(input) is a table containing both inputs x_1 and x_2, which are tensors of size N x inputDimension1 and N x inputDimension1, respectively. It is supposed to be  N x inputDimension1 and N x inputDimension2. 

[https://github.com/torch/nn/blob/master/doc/simple.md#nn.Bilinear](url)

soumith:
@BhargavaRamM thanks, fixed via https://github.com/torch/nn/commit/a81143f095c2aeb897f3f043699d10d0c4a96375

PR #831, opened 2016-05-21
Small error in equation in LookupTable part of `doc/convolution.md`

RishabhMalviya:
With reference to [this](https://github.com/torch/nn/issues/828) issue.


soumith:
Thank you!

Issue #832, opened 2016-05-24
LogSoftMax's backward returns an wrong gradInput while the gradOutput is not contiguous. 


qiuxin2012:
Test code:
```lua
require "nn"
layer = nn.LogSoftMax()
layer:zeroGradParameters()
input = torch.randn(4, 10)
data = torch.randn(4, 20)
gradOutput = data:narrow(2, 1, 10):fill(0)

output=layer:forward(input)
gradInput=layer:backward(input, gradOutput)

th> gradInput
 0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000
-0.7167 -0.0574  0.6574 -0.2126  0.6495 -0.3899  0.1446 -1.5247  0.4429  1.0070
 0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000
 0.8361  0.4220  0.9231 -0.7864 -0.9474  1.9022 -0.4281 -0.5947 -0.7279 -0.5989
[torch.DoubleTensor of size 4x10]

```

In this case, gradInput should be a tensor filled with zero. Am I right?

To fix this, we need to add makeContiguous for input and gradOutput in LogSoftMax.

nicholas-leonard:
I can confirm the bug. Not sure makeContiguous will work though. @qiuxin2012 Can you send a Pull Request?

qiuxin2012:
@nicholas-leonard Please check the Pull Request. Thanks.

nicholas-leonard:
@qiuxin2012 Sorry about the PR. Didn't know @soumith had a fix in mind :)

qiuxin2012:
@nicholas-leonard He did much better. :)

PR #833, opened 2016-05-24
add makeContiguous to LogSoftMax

qiuxin2012:
As [LogSoftMax's backward returns an wrong gradInput while the gradOutput is not contiguous.](https://github.com/torch/nn/issues/832)
I think we need makeContiguous for input and gradOutput to fix this.

### DIFF
--- a/LogSoftMax.lua
+++ b/LogSoftMax.lua
@@ -1,6 +1,24 @@
 local LogSoftMax = torch.class('nn.LogSoftMax', 'nn.Module')
 
+local function makeContiguous(self, input, gradOutput)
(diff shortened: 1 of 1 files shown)


fmassa:
It would be good to [add entries on `clearState`](https://github.com/torch/nn/blob/master/Normalize.lua#L144-L155) about `_input` and `_gradInput` as well.

soumith:
actually, this is not needed for the GPU case, as it directly handles the non-contiguous case (in the code) with striding. I will fix it in an hour for CPU case directly in THNN. Thanks for the effort.

PR #835, opened 2016-05-27
nn.GPU

nicholas-leonard:
This PR adds `nn.GPU` which can be used to distribute modules across multiple devices. 

OMG why is it in nn and not in cunn? As discussed with @soumith, putting in nn means that it can be used in CPU-only environments, which is common for production.

The unit tests are located in cunn to avoid a `pcall(function() require cunn end)` in the `nn` unit tests (see PR https://github.com/torch/cunn/pull/282).

### DIFF
--- a/GPU.lua
+++ b/GPU.lua
@@ -240,7 +240,13 @@ function GPU:read(file)
    local header = file:readObject()
    local object
    if header[1] == 'torch.CudaTensor' then
-      object = cutorch.withDevice(header[2], function() return file:readObject() end)
+      local device = header[2] 
+      if device > cutorch.getDeviceCount() then
+         print"Warning : model was saved with more devices than available on current host."
+         print"Attempting to load module onto device 1"
+         device = 1
+      end
+      object = cutorch.withDevice(device, function() return file:readObject() end)
    else
       object = file:readObject()
    end
@@ -259,9 +265,9 @@ function GPU:__tostring__()
 end
 
 function GPU:accUpdateGradParameters(input, gradOutput, lr)
-   error"Not Implemented"
+   error("Not Implemented for "..torch.type(self))
 end
 
 function GPU:sharedAccUpdateGradParameters(input, gradOutput, lr)
-   error"Not Implemented"
+   error("Not Implemented for "..torch.type(self))
 end
--- a/GPU.lua
+++ b/GPU.lua
@@ -1,0 +1,274 @@
+------------------------------------------------------------------------
+--[[ GPU ]]--
+-- Decorates a module such that its parameters are
+-- hosted on a specified GPU device.
+-- The operations are also executed on that device.
+-- Arguments input and gradOutput are converted to the specified device 
+-- before being fed to the decorated module. 
+-- Returned output is on the specified outdevice (defaults to device). 
+-- Returned gradInput is allocated on the same device as the input.
+-- The unit test is located in cunn.
+------------------------------------------------------------------------
+local GPU, parent = torch.class("nn.GPU", "nn.Container")
+
+function GPU:__init(module, device, outdevice)
+   parent.__init(self)
+   assert(torch.type(device) == 'number')
+   self.device = device
+   self.outdevice = outdevice or device
+   
+   assert(torch.isTypeOf(module, 'nn.Module'))
+   self.modules[1] = module
+   
+   if module:type() == 'torch.CudaTensor' then
+      self:cuda()
+   end
+end
+
+function GPU.recursiveModuleDevice(obj, device)
+   if type(obj) == 'table' and not torch.isTypeOf(obj, 'nn.GPU') then
+      for k,v in pairs(obj) do
+         obj[k] = GPU.recursiveModuleDevice(v, device)
+      end
+   elseif torch.type(obj):match('torch.Cuda.*Tensor') then
+      if obj:getDevice() ~= device then
+         obj = obj:clone() -- this will reallocate it to device
+         local newdevice = obj:getDevice()
+         -- when nElement() == 0 newdevice is 0
+         assert(newdevice == device or newdevice == 0)
+      end
+   end
+   assert(obj ~= nil)
+   return obj
+end
+
+-- set the device of the decorated module
+function GPU:setDevice(device)
+   self.device = device or self.device
+   
+   assert(self.modules[1])
+   self.modules[1] = cutorch.withDevice(self.device, function() 
+      return self.recursiveModuleDevice(self.modules[1], self.device) 
+   end)
+   return self
+end
+
+-- returns a dst that has device device for each element in src
+function GPU.recursiveSetDevice(dst, src, device)
+   if torch.type(src) == 'table' then
+      dst = torch.type(dst) == 'table' and dst or {}
+      for k,v in ipairs(src) do
+         dst[k] = GPU.recursiveSetDevice(dst[k], v, device)
+      end
+      for k=#src+1,#dst do
+         dst[k] = nil
+      end
+   elseif torch.type(src):match('torch.Cuda.*Tensor') and src:getDevice() ~= device and src:getDevice() ~= 0 then
+      if not (torch.type(dst):match('torch.Cuda.*Tensor') and dst:getDevice() == device) then
+         dst = src.new()
+      end
+      dst:resizeAs(src):copy(src)
+   else
+      dst = src
+   end
+   return dst 
+end
+
+-- makes sure dst is a identical to src except but on the same device as proto
+function GPU.recursiveSetDeviceAs(dst, src, proto)
+   local device
+   if torch.isTensor(proto) then
+      device = proto:getDevice()
+   elseif torch.type(proto) == 'number' then
+      device = proto
+   end
+   if torch.type(src) == 'table' then
+      dst = torch.type(dst) == 'table' and dst or {}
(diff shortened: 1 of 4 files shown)


doc/simple.md
```diff
@@ -1357,3 +1358,50 @@ One can also call:
...
+for sequential models where a single GPU doesn't have enough memory. 
+
+Example using 4 GPUs:
+
+```lua
+mlp = nn.Sequential()
+   :add(nn.GPU(nn.Linear(10000,10000), 1))
+   :add(nn.GPU(nn.Linear(10000,10000), 2))
+   :add(nn.GPU(nn.Linear(10000,10000), 3))
+   :add(nn.GPU(nn.Linear(10000,10000), 4, cutorch.getDevice()))
```
iamalbert: I am wondering if this line can run in CPU-only environments

nicholas-leonard: the cutorch.getDevice will not.

nicholas-leonard:
I have tested the nn.GPU implementation on a language model task using 4 GPUs and it works.

iamalbert:
This is cool but the name and the order of parameters is .. a little confusing

```lua
:add(nn.OnGPU(1, nn.Linear(10000,10000))
:add(nn.OnGPU(2, nn.Linear(10000,10000))
:add(nn.OnGPU(3, nn.Linear(10000,10000))
:add(nn.OnGPU(4, cutorch.getDevice(), nn.Linear(10000,10000)))
```
may be clearer than
```lua
:add(nn.GPU(nn.Linear(10000,10000), 1))
:add(nn.GPU(nn.Linear(10000,10000), 2))
:add(nn.GPU(nn.Linear(10000,10000), 3))
:add(nn.GPU(nn.Linear(10000,10000), 4, cutorch.getDevice()))
```
how do you think?

nicholas-leonard:
@iamalbert I prefer nn.GPU, but I like your order. Except for the optional outdevice argument which I think should still be last (as it is optional):

```lua
:add(nn.GPU(1, nn.Linear(10000,10000))
:add(nn.GPU(2, nn.Linear(10000,10000))
:add(nn.GPU(3, nn.Linear(10000,10000))
:add(nn.GPU(4, nn.Linear(10000,10000)), cutorch.getDevice())
```

szagoruyko:
The order is `module, device` in ModelParallel already, it would be confusing if we introduce another order

nicholas-leonard:
@szagoruyko Good point. The argument of backwards compatibility always wins (plus I am lazy). I will leave it as it is.

szagoruyko:
@nicholas-leonard there is a couple of places like `self.output = output` that break tensor sharing and optnet, can we avoid it?

nicholas-leonard:
@szagoruyko how does it break tensor sharing (I don't know anyone that shares outputs). As for optnet, what does it do that cannot support `self.output = output`? because I am pretty sure I do this in many other places.

fmassa:
Hi,

Sorry for the delay in replying.

The current underlying requirement in `optnet` is that the tensor/storage objects corresponding to the `output` and `gradInput` of each module doesn't change across runs of `forward`/`backward`.
There are two cases actually:
* when one wants to create graph visualizations, the tensor shouldn't change.
Thus doing something like the following will currently not work (although there is [a pending PR which tries to fix it](https://github.com/fmassa/optimize-net/pull/10), but I'm not 100% sure it doesn't have bad side effects):
```lua
function MySelectModule:updateOutput(input)
  self.output = input[1] -- creates a new tensor which shares the storage
  return self.output
end
```
* when we care only about optimizing for memory, `optnet` only looks for the storages. So a module which allocates new memory for the `output`/`gradInput` during each `forward`/`backward` won't be able to reuse buffers, but I think that it shouldn't cause any problems wrt the correctness of the entire network `forward`/`backward`.

About this PR, as it's a generic module that supports both `tensors` and `tables`, using `set` is not an option. But from a quick look I have the impression that this module could potentially work as is with `optnet`, but I haven't tried it to check.

nicholas-leonard:
So then I guess this PR is ready to merge then :)

GPU.lua
```diff
@@ -0,0 +1,274 @@
...
+   parent.__init(self)
+   assert(torch.type(device) == 'number')
+   self.device = device
+   self.outdevice = outdevice or device
+   
+   assert(torch.isTypeOf(module, 'nn.Module'))
+   self.modules[1] = module
+   
+   if module:type() == 'torch.CudaTensor' then
+      self:cuda()
```
soumith: this :cuda() is no executing in the context of "device". needs a fix

GPU.lua
```diff
@@ -0,0 +1,274 @@
...
+      for k=#src+1,#dst do
+         dst[k] = nil
+      end
+   elseif torch.type(src):match('torch.Cuda.*Tensor') and src:getDevice() ~= device and src:getDevice() ~= 0 then
+      if not (torch.type(dst):match('torch.Cuda.*Tensor') and dst:getDevice() == device) then
+         dst = src.new()
+      end
+      dst:resizeAs(src):copy(src)
+   else
+      dst = src
```
soumith: if src device = dst device, they are sharing the tensor. if src device != dst device, they are not sharing the tensor. this is not obvious, and probably will lead to bugs

GPU.lua
```diff
@@ -0,0 +1,274 @@
...
+   return dst 
+end
+
+-- makes sure dst is a identical to src except but on the same device as proto
+function GPU.recursiveSetDeviceAs(dst, src, proto)
+   local device
+   if torch.isTensor(proto) then
+      device = proto:getDevice()
+   elseif torch.type(proto) == 'number' then
+      device = proto
```
soumith: else... error

GPU.lua
```diff
@@ -0,0 +1,274 @@
...
+   end
+   if torch.type(src) == 'table' then
+      dst = torch.type(dst) == 'table' and dst or {}
+      for k,v in ipairs(src) do
+         dst[k] = GPU.recursiveSetDeviceAs(dst[k], v, proto[k])
+      end
+      for k=#src+1,#dst do
+         dst[k] = nil
+      end
+   elseif torch.type(src):match('torch.Cuda.*Tensor') and src:getDevice() ~= device and src:getDevice() ~= 0 then
```
soumith: this whole function can just call GPU.recursiveSetDevice above after figuring out the device id in line 79-84. Lots of duplication

GPU.lua
nicholas-leonard: It is : https://github.com/nicholas-leonard/nn/blob/db7b97209e3a4b540d51447be6ebfaf3b8fc28a9/GPU.lua#L158-L162

GPU.lua
```diff
@@ -0,0 +1,274 @@
...
+end
+
+function GPU:updateOutput(input)
+   if self._type == 'torch.CudaTensor' then
+      local output = cutorch.withDevice(self.device, function()
+         self._input = self.recursiveSetDevice(self._input, input, self.device)
+         return self.modules[1]:updateOutput(self._input)
+      end)
+      
+      if self.device ~= self.outdevice then
```
soumith: the code never reaches this whole block. it has a return statement in line 108 (two lines above)

soumith: oh nevermind about this comment. my bad.

GPU.lua
```diff
@@ -0,0 +1,274 @@
...
+      self.gradInput = nil
+      self._input = nil
+      self._gradOutput = nil
+      parent.type(self, type, typecache)
+   end
+   return self
+end
+
+function GPU:clearState()
+   self.output = nil
```
soumith: dont call = nil

call nn.utils.clear('output', 'gradInput')

GPU.lua
```diff
@@ -0,0 +1,274 @@
...
+   end
+   local header = {self._type, self.device}
+   file:writeObject(header)
+   file:writeObject(object)
+end
+
+function GPU:read(file)
+   local header = file:readObject()
+   local object
+   if header[1] == 'torch.CudaTensor' then
```
soumith: it would be really useful / worth it here to check if total number of devices <= header[2].
If there are less devices than header[2], then print a WARNING and fallback to device[0].
That way, the model can be deserialized on a 1-GPU machine, though it was created on a 4-GPU machine, for example.

GPU.lua
```diff
@@ -0,0 +1,274 @@
...
+function GPU:__tostring__()
+   if self.modules[1].__tostring__ then
+      return torch.type(self) .. '(' .. self.device ..') @ ' .. self.modules[1]:__tostring__()
+   else
+      return torch.type(self) .. '(' .. self.device ..') @ ' .. torch.type(self.modules[1])
+   end
+end
+
+function GPU:accUpdateGradParameters(input, gradOutput, lr)
+   error"Not Implemented"
```
soumith: useful to say: error("Not implemented in " .. torch.type(self))

GPU.lua
nicholas-leonard: You are right. They used to be more different. Refactored in latest commit.

GPU.lua
nicholas-leonard: Done in latest commit.

soumith:
Thanks for your patience Nicholas!

szagoruyko:
just realized, we need to add support for other cuda datatypes than float

soumith:
he seems to be checking for Cuda*Tensor, is that not sufficient?

PR #838, opened 2016-05-30
Channel-wise Randomized ReLU

xternalz:
Enabling all elements of the same channel to share the same `a`. This lowers the aggressive randomness of RReLU and may help to improve its performances (according to the author of RReLU paper [https://gist.github.com/antinucleon/5de43ed2ff11b944f1bd](https://gist.github.com/antinucleon/5de43ed2ff11b944f1bd)).


soumith:
this looks good to go as soon as you have a cunn PR

PR #839, opened 2016-06-01
Changes to compile on Windows

elikosan:
With these fixes i can install torch7 with luarocks on Windows.
Compiler: Visual Studio 2015
OS: Windows 7 Pro


lib/THNN/generic/BatchNormalization.c
```diff
@@ -10,10 +10,10 @@ void THNN_(BatchNormalization_updateOutput)(
   bool train, double momentum, double eps)
 {
   long nInput = THTensor_(size)(input, 1);
-  long n = THTensor_(nElement)(input) / nInput;
+  long f,n = THTensor_(nElement)(input) / nInput;
 
   #pragma omp parallel for
-  for (long f = 0; f < nInput; ++f) {
+  for (f = 0; f < nInput; ++f) {
```
soumith: if you make this change, then you'll have to change the pragma to: #pragma omp parallel for private(f)

elikosan: Hi Soumith,

The loop iteration variables are private within their loops. So i believe it is not necessary to add private() here.

Eric.

soumith:
Thanks!

elikosan:
Thanks for the quick merge!
Best,
Eric.

PR #822, opened 2016-05-16
added support for negative lenghts to Narrow

CodeRect:
Imagine you would like to chop off ten pixels off of two sides of an image, whose size is unknown. To make this feasible, I changed the Narrow module such that it accepts negative lengths and behaves as follows.

```
th> a = torch.rand(1, 5, 6)

th> a
(1,.,.) = 
  0.4982  0.6585  0.4621  0.9703  0.7528  0.6298
  0.5106  0.2586  0.6974  0.5139  0.2741  0.0780
  0.5161  0.2279  0.6495  0.0238  0.0226  0.7178
  0.1969  0.4823  0.2021  0.7990  0.2389  0.1396
  0.6029  0.0335  0.0230  0.5429  0.8701  0.8451
[torch.DoubleTensor of size 1x5x6]

th> nn.Narrow(2, 2, -1):forward(a)
(1,.,.) = 
  0.5106  0.2586  0.6974  0.5139  0.2741  0.0780
  0.5161  0.2279  0.6495  0.0238  0.0226  0.7178
  0.1969  0.4823  0.2021  0.7990  0.2389  0.1396
  0.6029  0.0335  0.0230  0.5429  0.8701  0.8451
[torch.DoubleTensor of size 1x4x6]

th> nn.Narrow(2, 2, -2):forward(a)
(1,.,.) = 
  0.5106  0.2586  0.6974  0.5139  0.2741  0.0780
  0.5161  0.2279  0.6495  0.0238  0.0226  0.7178
  0.1969  0.4823  0.2021  0.7990  0.2389  0.1396
[torch.DoubleTensor of size 1x3x6]

th> nn.Narrow(3, 2, -2):forward(a)
(1,.,.) = 
  0.6585  0.4621  0.9703  0.7528
  0.2586  0.6974  0.5139  0.2741
  0.2279  0.6495  0.0238  0.0226
  0.4823  0.2021  0.7990  0.2389
  0.0335  0.0230  0.5429  0.8701
[torch.DoubleTensor of size 1x5x4]
```

Thank you!

PR #843, opened 2016-06-04
extended documentation of / added a test case for Narrow

CodeRect:
Extended the documentation of Narrow and added a test case for it. This has been requested in #822.


soumith:
Thanks!

PR #849, opened 2016-06-14
Added ReLU6 layer, test and doc.

jonathantompson:
The module is like the one found here:

http://tflearn.org/activations/#relu6


soumith:
just FYI, I added this 2 years ago into nnx: https://github.com/clementfarabet/lua---nnx/blob/master/SaturatedLU.lua

This unit really helps when training networks that eventually have to be mapped to limited precision, like 8-bit or 16-bit integer computes. Atleast that's what I used it for at MuseAmi.

But since you wrote so much docs and stuff, can accept it.
However, can I ask you to make the "6" to be configurable? There are certain batchnorm-less networks where you might have to go higher than 6, when mapping to 16-bit for example.

szagoruyko:
should we then have ClippedReLU with both thresholds configurable?

jonathantompson:
Ahh, I didn't see your nnx layer. Lots of people (myself included) tend to avoid nnx in the interest of using only stable code.  But I should have checked.  Sorry mate.

In any case, this is a more optimized implementation (it looks like yours is pure lua) and I think this should be in the default repo in the interest of feature parity.

"This unit really helps when training networks that eventually have to be mapped to limited precision, like 8-bit or 16-bit integer computes. At least that's what I used it for at MuseAmi."

Right, this is why I'm adding it :-)  Actually, I wrote this in the doc but maybe you're repeating it here for other torch people.

So I'm not entirely sold on having a variable threshold for a few reasons:

1. The default ReLU layer has 0 value and threshold and doesn't even take it in as a parameter to the constructor. I want to be consistent and do the same for ReLU6.
2. The clip value is arbitrary, since the previous layer's bias and weights can learn whatever scale you ask.  Mathematically there's no reason for a FP32 network to ever want to change the threshold, and so adding that control will only confuse people.
3. I agree that a value of "6" is not idea for 16-bit integer networks, however there are always many other modifications one needs to make to transform a fp32 network to a working 16bit network which will cause the bias and weights to change ANYWAY.  The only thing that matters is that there is a saturation value.
4. Most importantly: I want to be consistent with TF nomenclature so that the MANY TF users that use a layer called "ReLU6" aren't confused. Or is the torch community still pretending TF doesn't exist? :-)

soumith:
I'm not asking you to remove ReLU6. Refactor it into a separate class such that the upper bound is configurable, and then subclass ReLU6 from that base class, with 6 as the hardcoded value.

For example, what I am saying is:
```
local ReLU6, parent = torch.class('nn.ReLU6', 'nn.SaturatedLU')
function ReLU6:__init()
    parent.init(self, 0, 0, 6, 6)
end
```
After you create a class that makes the upper bound configurable.

soumith:
if you are unwilling to do that, i can merge this, and then do the refactor myself.

jonathantompson:
SaturatedLU is pure lua and does not support in-place operations (as opposed to optimized C/CUDA with in-place support).  Why would I want to derive it?

It's OK then. I'll just take down the PR.  It's not worth everyone's time.

fmassa:
Hi,
Thanks for the PR @jonathantompson, I have never thought about the reason why TF had this `relu6` :-)

I have one comment though: isn't this layer almost equivalent to [`nn.HardTanh(0,6)`](https://github.com/torch/nn/blob/master/lib/THNN/generic/HardTanh.c#L16-L22) (doc [here](https://github.com/torch/nn/blob/master/doc/transfer.md#hardtanh))?
Sure, `nn.HardTanh` doesn't allow inplace operations, so this PR is more memory efficient, but still, IMHO they are a bit too similar to have different implementations, right? 

I'd vote for adding inplace operations to `nn.HardTanh`, and deriving `nn.ReLU6 = nn.HardTanh(0,6)`, as @soumith mentioned.

jonathantompson:
Sure.  Sounds good to me.

PR #851, opened 2016-06-15
inplace HardTanh, subclass ReLU6

szagoruyko:
as discussed in https://github.com/torch/nn/pull/849
* remove relu6 from THNN
* add inplace to HardTanh
* subclass ReLU6 from HardTanh


ReLU6.lua
```diff
@@ -1,7 +1,9 @@
-local ReLU6, parent = torch.class('nn.ReLU6', 'nn.Module')
+local ReLU6, parent = torch.class('nn.ReLU6', 'nn.HardTanh')
```
soumith: while this is the most effective approach (and i think so as well), after discussing with @jonathantompson , he was concerned that having a self.max_val in ReLU6 will make people believe that it's a tunable parameter.
Can you instead not subclass it but call the THNN.HardTanh.updateOutput functions with the value 6 directly?

szagoruyko: @soumith done

HardTanh.lua
```diff
@@ -14,7 +18,8 @@ function HardTanh:updateOutput(input)
       input:cdata(),
       self.output:cdata(),
       self.min_val,
-      self.max_val
+      self.max_val,
```
soumith: backcompat? self.max_val or 1
self.inplace or false

szagoruyko: added inplace, why self.max_val? it was there

soumith:
Thanks Sergey!

PR #853, opened 2016-06-17
Added VolumetricReplicationPadding.

jonathantompson:
This is a simple clone of SpatialReplicationPadding to 3D.

NOTE: we could make SpatialReplicationPadding call this Volumetric implementation (with ``depth = 1`` and ``pfront = pback = 0``), however I was worried about the performance penalty (if anything) of having an unnecessary for loop for the Spatial (2D) version.  The tradeoff is this way we have more code duplication... Thoughts?  I'm happy to remove the SpatialReplicationPadding C code and instead point it to this (more general) version.

CUDA implementation incoming.


soumith:
Thanks!

jonathantompson:
No thank you :-)

PR #857, opened 2016-06-22
Correction in output size for dilated convolution

codeAC29:
Calculation of output size for dilated convolution was incorrect.


soumith:
Thanks!

PR #858, opened 2016-06-22
add noBias for nn.SpatialFullConvolution

PraveerSINGH:
Adds noBias for nn.SpatialFullConvolution


fmassa:
Could you add an entry in the documentation as well mentioning the `noBias` option ?

soumith:
Thanks Praveer!

Issue #774, opened 2016-04-16
[feature request] relaxing the restriction of nn.Linear


iamalbert:
For now `nn.Linear` only accept 1D or 2D tensors as input. It would be better if its batch-mode can deal with 3D and 4D tensors. That is, a relaxed `nn.Linear(n,m)` can transform a tensor of dimension `p x q x n` into another tensor of `p x q x m`, and likewise `p x q x r x n` into `p x q  x r x m`.

This is useful when processing sequences. For example, suppose I have a tensor of `N x T x D` (representing a mini-batch of `N` sequences, each having `T` vectors of `D`-dimension), and I want an extra abstraction of all the vectors. If `N` and `T` are variable among instances, this cannot be done by adding `nn.View` before/after the linear module, but the relaxed version of `nn.Linear(D,D2)` can make it.

What do you think about this suggestion? If I want to send a PR, should I change the implementation of `nn.Linear` or add a new module like `nn.GeneralizedLinear` or something?

JoostvDoorn:
I think this use case exists with many other modules as well. In @nicholas-leonard's RNN package you have `nn.Sequencer()` that can be used on tables, it is more advanced as it also allows for use with RNNs. It might make sense to have a Container that can do the same just on tensors.

iamalbert:
yes, but using `nn.Sequencer` introduces too much overhead

szagoruyko:
I think it would be useful to have this, but should be a separate from nn.Linear module that wraps it. Can't think of a good name now.

nicholas-leonard:
So like @iamalbert mentioned, you can do this by encapsulating the input module with two `nn.View` instances : 
```lua
seqlinear = nn.Sequential()
   :add(nn.View(p*q*r,n))
   :add(nn.Linear(n,m))
   :add(nn.View(p,q,r,m))
```
But this will not support varying values of p, q, r, etc... Such that I agree with @iamalbert that this would be a useful module. 

It could be done by simply by inheriting a `nn.Linear` module where the above view changes are applied manually before and after calling the `nn.Linear.updateOutput` : 

```lua
local SeqLinear, parent = torch.class("nn.SeqLinear", "nn.Linear")
function SeqLinear:updateOutput(input)
   input = input:view(...) -- squash the first dims
   parent.updateOutput(input)
   self.output = self.output:view(...) -- unsquash
   return self.output
end
```

But then we would end up needing to do this for other modules like `SpatialConvolution`, etc... 

[...]

PR #860, opened 2016-06-25
Bottle container

JoostvDoorn:
This is a fix for #774, using the 3rd suggested method from @nicholas-leonard [(see)](https://github.com/torch/nn/issues/774#issuecomment-217342870) to implement a Bottle container.


soumith:
the code was unnecessarily more verbose than needed, and the variable names were not very informative.
Instead of writing feedback and going back and forth, I fixed the issues in https://github.com/torch/nn/pull/860

Your attribution has been fully preserved and you will be shown as the contributor of the other PR in the git logs.

PR #861, opened 2016-06-25
Adds negative dim arguments to Select and Narrow

JoostvDoorn:
Adds negative dimension arguments to Select and Narrow.


nicholas-leonard:
This was long overdue :)

soumith:
Thanks! looks good.

PR #867, opened 2016-07-11
Adding TemporalDynamicKMaxPooling

Multihuntr:
I wanted to use a Dynamic K Max Pooling layer as described in: 
A Convolutional Neural Network for Modelling Sentences
 by Nal Kalchbrenner, Edward Grefenstette, Phil Blunsom
[https://arxiv.org/abs/1404.2188]

It is essentially a layer that will dynamically reduce the length of a sequence by some factor, down to a minimum size. If you do not include a scaling factor, then it will simply pick the maximums to reduce the sequence to the specified minimum length.

So I made it; thinking it may be useful to others.
This layer is intended for use with input sequences of differing lengths, but works for minibatches of sequences with the same length (in a tensor).


soumith:
Thanks a lot, and sorry for the delay in reviewing.

PR #869, opened 2016-07-12
Fix SpatialClassNLLCriterion when using OMP

colesbury:
Although the `b` variable is automatically private, the `elem` loop variable is not private.

Test: change test.lua to run with more than one thread


apaszke:
As far as I remember I originally placed both `b`, and `elem` declarations inside the loops, but it created some problems with MSVC and someone moved them outside (see [this PR](https://github.com/torch/nn/pull/839/files)). It's probably better to leave them there, but declare them as private in the pragma.

colesbury:
This for-loop is not a `pragma omp parallel for` loop

colesbury:
I moved the `elem` declaration outside of the loop

soumith:
Thanks!

PR #879, opened 2016-07-25
NLL Criteria: weight bound checking

mys007:
Checks that there are as many weights as classes (before, undefined values could be used).


soumith:
Thanks!

PR #882, opened 2016-07-27
Adding SpatialUpSamplingBilinear

paulineluc14:


lib/THNN/generic/SpatialUpSamplingBilinear.c
```diff
@@ -0,0 +1,127 @@
...
+  gradOutput = THTensor_(newContiguous)(gradOutput);
+  THTensor_(zero)(gradInput);
+  real *data1 = THTensor_(data)(gradInput);
+  real *data2 = THTensor_(data)(gradOutput);
+  int channels = THTensor_(size)(gradInput, 0) * THTensor_(size)(gradInput, 1);
+  int height1 = THTensor_(size)(gradInput, 2);
+  int width1 = THTensor_(size)(gradInput, 3);
+  int height2 = THTensor_(size)(gradOutput, 2);
+  int width2 = THTensor_(size)(gradOutput, 3);
+  assert(height1 > 0 && width1 > 0 && height2 > 0 && width2 > 0);
```
soumith: Replace this with THAssert. Looks like osstorch has a problem with this

lib/THNN/generic/SpatialUpSamplingBilinear.c
```diff
@@ -0,0 +1,127 @@
...
+  output = THTensor_(newContiguous)(output);
+  THTensor_(zero)(output);
+  real *idata = THTensor_(data)(input);
+  real *odata = THTensor_(data)(output);
+  int channels = THTensor_(size)(input, 0) * THTensor_(size)(input, 1);
+  int height1 = THTensor_(size)(input, 2);
+  int width1 = THTensor_(size)(input, 3);
+  int height2 = THTensor_(size)(output, 2);
+  int width2 = THTensor_(size)(output, 3);
+  assert(height1 > 0 && width1 > 0 && height2 > 0 && width2 > 0);
```
soumith: Replace this with THAssert. Looks like osstorch has a problem with this

soumith:
Thanks Pauline!

paulineluc14:
Thank you for your help !!

Le 27 juil. 2016 ? 20:26, Soumith Chintala <notifications@github.com<mailto:notifications@github.com>> a ?crit :


Thanks Pauline!

-

joeyhng:
The output sizes of SpatialUpSamplingNearest and SpatialUpSamplingBilinear are not the same. Would it be better to make them consistent? 

I haven't thought about the implementation issues, but I personally think the output size of SpatialUpSamplingNearest is more intuitive and easier to use.

paulineluc14:
Thanks joeyhng for your feedback !

Let me explain why I think this API is more transparent for SpatialUpSamplingBilinear.
For HxW pixels - where H = (h-1)x(w-1) + h (and likewise for w and W) - computation of the bilinear interpolation is according to the definition. On the remaining pixels of the borders, several choices can be made, which depend on the application. 
- simply copy of the nearest pixel
- extrapolation using linear interpolation
- symmetric padding
- filling in with NaN values
I guess most of these choices could be implemented using other torch layers (at least for the copy of the pixels and the symmetric padding). On the other hand, one can also design their network so as not to have to make any of these arbitrary choices.

Computation of SpatialUpsamplingBilinear, on the contrary, doesn't lead to such considerations on the borders.

Perhaps to make things clearer that both layers don't have the exact same API, we can think about changing the layer's name, for instance to SpatialBilinearInterpolation.

fmassa:
Does it make sense to support a second input argument (as in SpatialFullConvolution) that, if provided, gets the size of the input nulber 2 and use it as the size of the output of SpatialIpSamplingBilinear ? This would allows to exactly reproduce input sizes, when given that sometimes sope pixels are lost in the boundaries due to strides

paulineluc14:
Hi fmassa, thanks for your comment ! Could you specify your question ? Why would you want the output size to be the same as the input size ? Anyway it already achieves that, if scale = 1 then H = h and W = w so nothing happens.
PS : I had made a typo above, I just replaced w with scale, is that what was troubling you ?

fmassa:
Hi Pauline,
sorry, I wasn't clear on my first comment.

what I mean by passing a second input argument is that the scale (and additional padding offsets that may vary from input to input) are determined by the size of the second input, but the first input is the one which will be upsampled (and has different size than input 2).

why I'm saying this ?
Say we want the output of a deep network to have the same size as its input (for semantic segmentation for example). For a generic input size, after a few convolutions/max poolings with stride different than 1, we are usually not able to upsample with a fixed scale the output (say with a scale of 32 for most networks) so that we consistently get the output of the network to match the size of the input of the network.

paulineluc14:
Hi Francisco, yes I understand how that would be useful, thanks for the clarification. 

I just wonder whether it wouldn't be simpler to be doing that outside of a network, with already existing functions (like in the torch/image package). If you are doing this inside a network, I suppose it is because you would like to be able to learn the network's parameters taking into account the fact that the predictions are going to be bilinearly upsampled. If so, it would be better that the bilinear upsampling method is consistent from one sample to the other, whatever the input size. Again, I think that this should be taken into account during the network's design, but I may be wrong.

I guess I will also see in the future if certain applications lead me to implementing such an option, and of course share it here if I do.

arunmallya:
Hi, I needed to use the bilinear upsampling in a case where the output size was not an integer scale of the input size. It was rather easy to create a work around for the bilinear upsampling case and I have submitted a pull request - https://github.com/torch/nn/pull/991. 

Making this work for the nearest upsampling will however be a bit more work.

PR #885, opened 2016-07-28
add CMaxTable class

juesato:
This is analogous to the other table layers, but only with a max, rather than add or multiply operation. I couldn't think of a good way to implement this to work with arbitrary sized tensors using existing layers, so I thought this might be helpful.

If you think it makes sense to include this feature, I'll add documentation and tests, I just wanted to check whether this feature makes sense before going ahead with that.

Sample usage:
```
th> a = torch.range(1,6):double():resize(3,2)
                                                                      [0.0003s]	
th> b = torch.Tensor(3,2):fill(3.5):double()
                                                                      [0.0002s]	
th> a
 1  2
 3  4
 5  6
[torch.DoubleTensor of size 3x2]

                                                                      [0.0003s]	
th> b
 3.5000  3.5000
 3.5000  3.5000
 3.5000  3.5000
[torch.DoubleTensor of size 3x2]

                                                                      [0.0003s]	
th> nn.CMaxTable():forward({a,b})
 3.5000  3.5000
 3.5000  4.0000
 5.0000  6.0000
[torch.DoubleTensor of size 3x2]
```

### DIFF
--- a/CMaxTable.lua
+++ b/CMaxTable.lua
@@ -1,0 +1,33 @@
+local CMaxTable, parent = torch.class('nn.CMaxTable', 'nn.Module')
+
+function CMaxTable:__init()
+   parent.__init(self)
+   self.gradInput = {}
+   self.maxIdx = torch.Tensor()
+end
+
+function CMaxTable:updateOutput(input)
+   self.output:resizeAs(input[1]):copy(input[1])
+   self.maxIdx:resizeAs(input[1]):fill(1)
+   for i=2,#input do
+      local mask = torch.gt(input[i], self.output)
+      self.maxIdx:maskedFill(mask, i)
+      self.output:maskedCopy(mask, input[i][mask])
+   end
+   return self.output
+end
+
+function CMaxTable:updateGradInput(input, gradOutput)
+   for i=1,#input do
+      self.gradInput[i] = torch.Tensor()
+      self.gradInput[i]:resizeAs(input[i]):fill(0.0)
+      local mask = torch.eq(self.maxIdx, i)
+      self.gradInput[i]:maskedCopy(mask, gradOutput[mask])
+   end
+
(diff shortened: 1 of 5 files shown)


soumith:
this makes sense. please go ahead and add tests / doc.
Thanks!

juesato:
@soumith I've added tests and docs - can you take a look?

juesato:
@soumith ping?

nhynes:
Just out of curiosity, is there a reason to prefer duplicating code for both operations instead of making, for example, CExtremumTable that either takes min/max as a parameter or provides a base class for (Min|Max)Table?

juesato:
Not really, I think it feels a little heavy for two very small classes, since there would still need to be 3 classes I think since the `CMin/MaxTable` classes need to override the `updateOutput`. I can't think of another situation in which the `CExtremumTable` would be useful, so it seems like extra overhead.

Another way of implementing this I considered would be some sort of `CToTensorTable` layer, which would take a table of Tensors, add a singleton dimension in front and then join along that. Then we could use all the standard layers e.g. Max, Min, Add, followed by a SplitTable. But perhaps this incurs extra computational overhead (I'm not sure).

nhynes:
> two very small classes [...] still need to be 3 classes

I agree that three classes is definitely worse than two for this size. Going the route of  `CExtremumTable` with a min/max argument might be confusing for new users, too. I just get a bit antsy when I see copy+paste code :)

> table of Tensors, add a singleton dimension in front and then join along that [...] overhead

Yeah,  that would require O(n) unnecessary copies.

Thanks for addressing my concerns!

juesato:
Okay, thanks for the feedback! That makes sense to me.

soumith:
Thanks for the PR and sorry for the delay in review.

PR #890, opened 2016-07-31
Dice coefficients in torch

lakehanne:
This implements the [Sorensen's DICE Criterion](https://en.wikipedia.org/wiki/S%C3%B8rensen%E2%80%93Dice_coefficient#Difference_from_Jaccard). 

I have made comments in the `criterion.md` doc file and provided a unit test example as well. 

Hopefully it helps someone in the near future.

Example usage:

```lua

require 'torch'
require 'nn'

local dice = nn.DICECriterion

inputs = torch.FloatTensor(1, 5)
preds  = torch.FloatTensor(1, 5)

inputs = torch.range(1, 5)
preds  = inputs:clone()

loss = dice:forward(preds, inputs)
df_do = dice:backward(preds, inputs)

print('loss', loss)
```

Prints 

```bash
	loss	-0.9999999999999	
```

Example 2:
```lua
require 'cutorch'
input = torch.FloatTensor(3, 64, 64)
pred  = torch.FloatTensor(3, 64, 64)

input:fill(111)
pred:fill(111)

loss_img = dice:forward(pred, input)
df_do 	 = dice:backward(pred, input)

print('loss_img', loss_img)
```

Prints 

```bash
	loss_img	-1	
```

### DIFF
--- a/DICECriterion.lua
+++ b/DICECriterion.lua
@@ -1,0 +1,96 @@
+--[[
+	Author: Olalekan Ogunmolu, July 2016
+			patlekano@gmail.com
+]]
+
+local DICECriterion, parent = torch.class('nn.DICECriterion', 'nn.Criterion')
+
+local eps = 1e-12
+
+function DICECriterion:_init(weights)
+	parent._init(self)
+
+	if weights then
+	   assert(weights:dim() == 1, "weights input should be 1-D Tensor")
+	   self.weights = weights
+	end
+
+end
+
+function DICECriterion:updateOutput(input, target)
+
+	assert(input:nElement() == target:nElement(), "input and target size mismatch")
+
+	local weights = self.weights
+
+	local numerator, denom, common, output
+
+	if weights ~= nil and target:dim() ~= 1 then
+	      weights = self.weights:view(1, target:size(2)):expandAs(target)
+	end
+
+	-- compute numerator: 2 * |X n Y|   ; eps for numeric stability
+	common = torch.eq(input, target)  --find logical equivalence between both
+	common:mul(2)
+	numerator = torch.sum(common)
+
+	-- compute denominator: |X| + |Y|
+	denom = input:nElement() + target:nElement() + eps
+
+	output = numerator/denom
+
+	self.output = -output
+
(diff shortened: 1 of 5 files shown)


DICECriterion.lua
```diff
@@ -0,0 +1,95 @@
...
+
+	local buffer = self.buffer
+	local weights = self.weights
+	local gradInput = self.gradInput --or input.new()
+
+	if weights ~= nil and target:dim() ~= 1 then
+	    weights = self.weights:view(1, target:size(2)):expandAs(target)
+	end
+
+	buffer:resizeAs(input)
```
soumith: buffer here is uninitialized memory. you need to zero it first.

DICECriterion.lua
```diff
@@ -0,0 +1,95 @@
...
+	-- compute |X| + |Y| + eps 
+	buffer:add(input:nElement()):add(target:nElement()):add(eps)
```
soumith: |X| is what? norm of x? or the dimesionality of X? if it is norm of X, this line is wrong.

DICECriterion.lua
```diff
@@ -0,0 +1,95 @@
...
+	-- compute (|X| + |Y| + eps)^2 + eps
+	buffer:cmul(buffer):add(eps)
+	-- compute |X|(|X| + |Y| + eps)^2 + eps
+	buffer:mul(input:nElement())
+
+	gradInput:resizeAs(input)
```
soumith: same here. uninitialized memory, dont do a :add two lines below without zeroing it.

soumith:
made inline comments. also add test like here: https://github.com/torch/nn/blob/master/test.lua#L1314-L1319

DICECriterion.lua
lakehanne: Pardon the confused representation. I picked it up from the wiki page. It is the sum of the elements in sample X. It actually is sum_i(X) where X is a sample dataset.

lakehanne:
Thanks for your comments. Please see commit **1431ee33e55265c5073b2f0c13b074d65041bcee**

Issue #910, opened 2016-08-05
[proposal] CMul dimensions


nhynes:
What do people think about requiring CMul's `input:dim()` to match `weight:dim()`? `weight` would be expanded along singleton dimensions to provide for per-axis multiplication.

Motivating this change is Caffe's `Scale` layer, which operates on a particular axis; having this layer would make it easier to construct resnets using loadcaffe.

The unfortunate part is that this would also be a backwards incompatible change since one would no longer be able to do something like the following (which is, indeed strange and may not be the most intuitive behavior).
```lua
local t = torch.rand(3,4)
local c = nn.CMul(3, 4)
c:forward(t)
```

PR #911, opened 2016-08-05
Expand CMul weights

nhynes:
This will allow CMul'ing along a particular axis (and without a single `setNumInputDims`!)

```lua
local c = nn.CMul(1, 2, 3, 1)
local t = torch.rand(3, 2, 3, 4)
c:forward(t)
c:backward(t, t)
```

This PR is a possible implementation of the proposal in #910  that maintains backward compatibility, as measured by the tests passing. I'll add tests for this new feature, too, if it's deemed useful!

### DIFF
--- a/CMul.lua
+++ b/CMul.lua
@@ -47,10 +47,15 @@ function CMul:updateOutput(input)
       
       self._output:cmul(self._weight)
    else
-      local batchSize = input:size(1)
-      self._output:view(self.output, batchSize, -1)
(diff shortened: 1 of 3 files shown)


soumith:
this is good to merge. can you append unit tests and doc to this PR.
Thanks.!

nhynes:
Great, thanks! I've added some tests and updated the docs; I hope that they're sufficiently incisive.

soumith:
Thanks!

PR #913, opened 2016-08-07
initialized buffer/gradInputs memory and added tests for dice

lakehanne:
This includes the memory initialization for buffer and gradInput. It also clarifies the definition of |X|.

### DIFF
--- a/DICECriterion.lua
+++ b/DICECriterion.lua
@@ -1,0 +1,108 @@
+--[[
+		Computes the Sorensen-dice coefficient of similarity given two samples. 
+		The quotient of similarity is defined as:
+
+   		 	    Q =     2 * (X n Y)
+         		     -------------------
+        		      sum_i(X) + sum_i(Y)
+		where X and Y are the two samples; 
+		(X n Y) denote the intersection where the elements of X and Y are equal.
+
+	Author: Olalekan Ogunmolu, July 2016
+			patlekano@gmail.com
+]]
+
+local DICECriterion, parent = torch.class('nn.DICECriterion', 'nn.Criterion')
+
+local eps = 1
+
+function DICECriterion:_init(weights)
+	parent._init(self)
+
+	if weights then
+	   assert(weights:dim() == 1, "weights input should be 1-D Tensor")
+	   self.weights = weights
+	end
+
+end
+
+function DICECriterion:updateOutput(input, target)
+
+	assert(input:nElement() == target:nElement(), "input and target size mismatch")
+
+	local weights = self.weights
+
+	local numerator, denom, common, output
+
+	if weights ~= nil and target:dim() ~= 1 then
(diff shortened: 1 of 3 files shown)


DICECriterion.lua
```diff
@@ -0,0 +1,108 @@
...
+	end
+
+	if self.sizeAverage then
+	    gradInput:div(target:nElement())
+	end
+
+	return gradInput
+end
+
+function DICECriterion:accGradParameters(input, gradOutput)
```
soumith: this is not needed, as this is a Criterion. Criterions dont have accGradParameters

soumith:
you need to add the criterion to init.lua , otherwise it wont be loaded.
Make sure the continuous build passes.

lakehanne:
Is there any other thing I need to do apart from adding `require('nn.DICECriterion')` to init? I found that running `th -lnn -e "nn.test{'DICECriterion'}" returned the ff error: 

```bash
/home/lex/torch/install/share/lua/5.1/torch/Tester.lua:508: Couldn't find test 'DICECriterion'
```

soumith:
now the tests fail because your backward gradients dont match gradients from finite difference. there's likely a bug in your gradient calculation.

lakehanne:
That's because I added a normalizing constant to the gradient formula's denominator to avoid division errors.

Maybe I should remove it?

soumith:
can you make the normalizing constant be optional in the constructor, and disable it when running the tests...

lakehanne:
Sure

Sincerely,
Lekan

lakehanne:
```bash
torch/install/share/lua/5.1/torch/Tester.lua:508: Couldn't find test 'DICECriterion'	
```

Was I supposed to do some tricks in Tester.lua as well?

It seems to be taking longer than usual to make this PR work with Travis and your unit tests. Feel free to close/delete the PR. I will leave my implementation on my forked copy for at least a month from now if you or anyone can manage to integrate it into the unit tests.

Cheers.

tommy-qichang:
Hi Lekan,
I just wanna know whether this DICECriterion will merge or not? I need it desperately.
And will this criterion support volumetric inputs?

lakehanne:
It should merge. I believe I implemented it on a volumetric input back in the summer if I remember correctly.

tommy-qichang:
Hi
I think your implementation has some problems especially for back propagation part.

> http://campar.in.tum.de/pub/milletari2016Vnet/milletari2016Vnet.pdf

In this paper, it showed the gradient formula which I think is correct.

lakehanne:
Thanks for spotting this. I think I plugged the wrong equations into wolfram alpha to generate the gradients early on. Could you send a PR?

tommy-qichang:
I'm pretty busy right now. But I can send a PR maybe next January.

PR #914, opened 2016-08-07
add 3D deconvolution description into VolumetricFullConvolution 

barrykui:
add "3D deconvolution" and "3D upsampling" keywords into VolumetricFullConvolution, so that  VolumetricFullConvolution can be easily searched.


soumith:
Thank you!

PR #918, opened 2016-08-09
Fix for SpatialFullConvolution for table input after clearState

fmassa:
When the input is a `table`, `clearState` will empty `gradInput` by replacing it with an empty table.
This causes the backward pass to fail because `self.gradInput[1]` will be `nil`.

Should I add tests for this particular case in this PR? For the moment there are practically no tests for `clearState`.


soumith:
Thanks!

PR #921, opened 2016-08-10
Add Spatial Dilated Max Pooling

kmul00:
I was in two minds regarding whether to make changes to the existing `SpatialMaxPooling` itself or have a separate front-end `lua` file and make changes in the back-end `C` code.

Finally went with the latter to reduce confusion.

new file:   SpatialDilatedMaxPooling.lua
modified:   SpatialMaxPooling.lua
modified:   init.lua
modified:   lib/THNN/generic/SpatialMaxPooling.c
modified:   test.lua


kmul00:
Checks are failing because I forgot to update the function signature of `SpatialMaxPooling` in `THNN_H.lua`. 

But now I cannot resolve it since it's added in the `.gitignore`.

soumith:
you need to add it to THNN.h afaik. THNN_H.lua is autogenerated right...

kmul00:
Yeah, my bad !
Fixed. :)

soumith:
you planning a cunn version too?

SpatialDilatedMaxPooling.lua
```diff
@@ -0,0 +1,72 @@
...
+local SpatialDilatedMaxPooling, parent = torch.class('nn.SpatialDilatedMaxPooling', 'nn.SpatialMaxPooling')
+
+function SpatialDilatedMaxPooling:__init(kW, kH, dW, dH, padW, padH, dilationW, dilationH)
+   parent.__init(self, kW, kH, dW, dH, padW, padH)
+
+   self.dilationW = dilationW or 1
+   self.dilationH = dilationH or 1
+end
+
+function SpatialDilatedMaxPooling:ceil()
```
soumith: i dont think you need these right? they are already in SpatialMaxPooling (:ceil and :floor)

kmul00: I am using the initialization of MaxPooling. By default it sets to floor mode. From what I understood, you need to explicitly call :floor() or :ceil(). 

I thought similar needs to be done for DilatedPooling as well. Since those were separate function calls.

soumith: yes, but since you are deriving nn.SpatialDilatedMaxPooling from nn.SpatialMaxPooling, the methods are defined already.

kmul00: Oh I see. Yeah then those are not needed. I'll fix it.

kmul00:
`cunn` version I am planning to have a look at. 

These are easy extensions, so I guess it won't be much of a problem. I cannot say for sure that I will be able to do it, since I am not that adept in CUDA coding. I will have a look and see if I can do it, else I will inform.

SpatialDilatedMaxPooling.lua
```diff
@@ -0,0 +1,62 @@
...
+function SpatialDilatedMaxPooling:updateGradInput(input, gradOutput)
+   input.THNN.SpatialMaxPooling_updateGradInput(
+      input:cdata(),
+      gradOutput:cdata(),
+      self.gradInput:cdata(),
+      self.indices:cdata(),
+      self.kW, self.kH,
+      self.dW, self.dH,
+      self.padW, self.padH,
+      self.dilationW, self.dilationH,
```
kmul00: Also, though I am sending dilations as parameters, it was just to maintain consistency. Those are actually not needed while calculating the gradients since I already have the indices stored. So you think I should remove the dilation parameters, then the existing MaxPooling updateGradInput can be used as it is. No need of this new function.

soumith: send them in. keep it consistent.

kmul00: Yeah, cool.

soumith:
ok i'll merge this once the cunn version is in. Because at the moment, cunn will fail, as the function signature changed.

kmul00:
`CUDA` version [here](https://github.com/torch/cunn/pull/317).

soumith:
rebased and merged via https://github.com/torch/nn/commit/bc5bc382b9ce49a7f36a894f76f9c34d6468d579

Please send me a PR for documentation.

kmul00:
[Here](https://github.com/torch/nn/pull/925).
Apologies for missing it earlier.

szagoruyko:
this change in SpatialMaxPooling API annoyingly breaks any code that depends on THNN or THCUNN.

kmul00:
Hi Sergey,

Sorry about that. Just to be sure, since I made changes to an existing function signature in `THNN` (*SpatialMaxPooling*), it's breaking existing codes dependent on it ?

Any suggestions regarding how to achieve this without changing the existing signature of *SpatialMaxPooling* in `THNN` ? 

Having a separate API just for Dilated Pooling seemed unreasonable to me since 90% of the code is reused from Max Pooling.

szagoruyko:
we should either have local functions that implement dilated maxpooling and instantiate both SpatialMaxPooling and StatialDilatedMaxPooling functions from them, or just rename the current SpatialMaxPooling functions to dilated and instantiate max pooling from them to keep the API consistent

kmul00:
Yeah, sounds cool.
I would wait for @soumith to chip in regarding which can be the preferred way to go about and make the necessary changes after that.

soumith:
I thought this was okay because that's kinda how the caffe folks did it. They added dilation parameters just like stride and size. Let's go with whatever sergey suggested.

kmul00:
Okay, so I am going to go with the second option that @szagoruyko  suggested. Rename the current `SpatialMaxPooling` to `DilatedSpatialMaxPooling` and instantiate `MaxPooling` from it so the existing signatures remain consistent.

PR #927, opened 2016-08-13
cmake: add soversion for libTHNN.so

CDLuminate:
This change is imported from Debian package.


fmassa:
This change *as is* won't work with `ffi`.
We discussed about it in the `gitter` channel https://gitter.im/apaszke/THNN on August 5, but we have not yet reached an agreement on this.

CDLuminate:
@fmassa I took a look on that gitter history, and do you mean that this change might bring unexpected result to OS X users although it works for linux users?

fmassa:
@CDLuminate This change broke the build for half of the systems in Travis (which are running in Linux), check the history.

soumith:
because package.cpath is usually ?.so and not ?.dylib etc.
You cannot change it from MODULE to SHARED

CDLuminate:
I'll look into it later.

PR #932, opened 2016-08-16
Don't overwrite self.size in Concat

### LINKED ISSUE #931: nn.Concat overrides size()
`nn.Concat` stores the output size in `self.size` and thereby overrides the `size()` method of the container interface defined in `nn.Container`. I guess this can be safely fixed by storing the output size in `outputSize` instead of `size`?

jgehring:
size() is a method of the Container base class.
fixes #931


soumith:
.outputSize wont exist for models loaded from disk. Can you initialize it in :forward if it is nil

soumith:
Thanks Jonas!

PR #933, opened 2016-08-16
fixing Max and Min for new cutorch types

soumith:
fixes https://github.com/torch/cutorch/issues/468

cc: @szagoruyko @shuzi @jayleicn


soumith:
will merge once sergey reviews it.

szagoruyko:
looks good

PR #936, opened 2016-08-22
Stochastic Inference for Dropout

xternalz:
Applying dropout stochastically during inference, as an alternative to the 'expectation' approach.


Dropout.lua
```diff
@@ -20,13 +21,14 @@ function Dropout:updateOutput(input)
       self.output:resizeAs(input):copy(input)
    end
    if self.p > 0 then
-      if self.train then
-         self.noise:resizeAs(input)
-         self.noise:bernoulli(1-self.p)
+      local si = self.si and not self.train
+      if self.train or si and not self.v2 then
+         local noise = si and input:clone() or self.noise:resizeAs(input)
```
fmassa: Do you need to clone the input at every iteration? Also, the previous `self.noise:resizeAs(input)` is not adapted in this situation?

xternalz: This helps to save memory during testing when self.noise is cleared/uninitialized.

Dropout.lua
```diff
@@ -20,13 +21,14 @@ function Dropout:updateOutput(input)
       self.output:resizeAs(input):copy(input)
    end
    if self.p > 0 then
-      if self.train then
-         self.noise:resizeAs(input)
-         self.noise:bernoulli(1-self.p)
+      local si = self.si and not self.train
+      if self.train or si and not self.v2 then
```
fmassa: Are you sure about this condition?
Just a few lines downstream in the same `if` block, there is the case where `v2` is used considered, but it will never reach it with the current condition.

xternalz: It will reach the block for `v2` when `self.train == true`.

Dropout.lua
fmassa: If you want to save memory during inference, it is usually better to just share the `noise` field among other dropout layers or other internal buffers from other modules.
Currently, this line is actually allocating new memory at every iteration, and freeing it just afterwards, which is a blocking operation on GPU. We usually try to avoid that as much as possible in `nn`.

xternalz: Thanks for pointing this out. I will remove the cloning and use just `self.noise`. Has the feature of sharing `noise` field among layers been implemented in your [optnet](https://github.com/fmassa/optimize-net) library?

Dropout.lua
```diff
@@ -1,10 +1,11 @@
 local Dropout, Parent = torch.class('nn.Dropout', 'nn.Module')
 
-function Dropout:__init(p,v1,inplace)
+function Dropout:__init(p,v1,inplace,stochasticInference)
    Parent.__init(self)
    self.p = p or 0.5
    self.train = true
    self.inplace = inplace
+   self.si = stochasticInference or false
```
apaszke: `si` is a very cryptic name. It would be better to keep it more verbose.

soumith:
please document this option

soumith:
hi @xternalz . This is good to go, except that it looks like it needs a rebase on top of master.

joeyhng:
Would it be clearer if removing ```and not self.train``` which seems not needed?

Also, can this be done by simply setting the module in evaluate mode?

xternalz:
Please check again @soumith.

Dropout.lua
```diff
@@ -20,7 +21,7 @@ function Dropout:updateOutput(input)
       self.output:resizeAs(input):copy(input)
    end
    if self.p > 0 then
-      if self.train then
+      if self.train or self.stochastic_inference and not self.train and not self.v2 then
```
soumith: hi, just looked at the logic again. isn't this wrong?
if it is evaluate mode, this will never come here because self.v2 will almost always be true.

shouldn't it just be:

if self.train or self.stochastic_inference

SpatialDropout.lua
```diff
@@ -1,15 +1,16 @@
...
    self.p = p or 0.5
    self.train = true
+   self.stochastic_inference = stochasticInference or false
    self.noise = torch.Tensor()
 end
 
 function SpatialDropout:updateOutput(input)
    self.output:resizeAs(input):copy(input)
-   if self.train then
+   if self.train or self.stochastic_inference and not self.train then
```
soumith: this can also simply be:

if self.train or self.stochastic_inference

soumith:
added final comments. sorry for not catching these earlier.

Dropout.lua
xternalz: Yea, you are right.

soumith:
Thanks @xternalz !

tastyminerals:
Guys, shouldn't `Dropout.lua` in **rnn** package be the same `Dropout.lua` in **nn**? **Element-Research** have updated it to include `mono` option. So, `LSTM.lua`, `FastLSTM.lua`, `GRU.lua` all have `mono` option which they expect for `Dropout` to accept:

**nn** `Dropout:__init(p,v1,inplace,stochasticInference)`
**rnn** `Dropout:__init(p,v1,inplace,lazy,mono)`

PR #937, opened 2016-08-22
A container for applying a single module to all table elements

jgehring:
This module can be useful when dealing with multi-dimensional mini-batches, for
example when training recurrent neural networks (time X batch_size X elements).
It simply clones the contained module as necessary and applies each clone to the
respective input element.


soumith:
Thanks Jonas!

jgehring:
Thanks for the quick merge!

Issue #920, opened 2016-08-10
nn.JoinTable negative dimension support errors


deet:
The code
```
nn.JoinTable(-1):forward{torch.randn(5, 1), torch.randn(5, 1)}
```
produces an error at 

`nn/JoinTable.lua:14: attempt to call method 'dim' (a nil value)`

(https://github.com/torch/nn/blob/ead8af9c14f26116c1fe4fc5d58a6b0de2b2603f/JoinTable.lua#L14)

while
```
nn.JoinTable(1):forward{torch.randn(5, 1), torch.randn(5, 1)}
```
works fine.

This code appears copied from from SplitTable, but the input to JoinTable is table, not a tensor, so there is no `dim` method.

soumith:
thanks for the report. fixed now in master.

PR #940, opened 2016-08-24
fixing JoinTable bug for negative indices

soumith:
fixes https://github.com/torch/nn/issues/920


CodeRect:
The same logic is being used in other modules as well, just to name two:
https://github.com/torch/nn/blob/master/SplitTable.lua#L12
https://github.com/torch/nn/blob/master/Sum.lua#L14

Are these potentially affected and need a fix as well?

Issue #734, opened 2016-03-25
Adding dataset assertions for StochasticGradient


gforge:
I recently found that the metatable wasn't copied between threads and ran into an error that while obvious (once you're familiar with Lua) perhaps isn't the most informative:

> /root/torch/install/share/lua/5.1/nn/StochasticGradient.lua:32: attempt to index local 'example' (a nil value)

I suggest adding a basic assertion to check if the dataset has the expected metatable function:

```Lua
assert(type(getmetatable(dataset)["__index"]) == "function", 
       "The dataset has no __index function set in the metatable")
```

I think this would solve many of the issues us new to language have when trying to construct our first network. Additional check that the function also returns the expected output would be nice but that is probably less critical.

PR #946, opened 2016-08-31
HardTanh does not use inclusive bounds in inplace mode

gchanan:
HardTanh does not use inclusive bounds in inplace mode, which makes it
inconsistent with the cuda version, as well as Threshold, ReLU, etc.

Note this doesn't fix torch7 issue #734 (https://github.com/torch/torch7/issues/734), but the inconsistency was causing issues on trivial with tests between the cuda and non-cuda versions of HardTanh.


soumith:
Thanks Greg!

Issue #949, opened 2016-09-03
Unstable nn.MultiLabelSoftMarginCriterion


arunmallya:
Shouldn't the Sigmoid and Binary Cross Entropy functions be combined inside nn.MultiLabelSoftMarginCriterion for numerical stability? 

The current implementation calls Sigmoid followed by BCE on the inputs, and the reverse for gradient computation. However, when activations are large, the results go all wrong.

```
x = torch.Tensor{-50}
t = torch.Tensor{1}

mls = nn.MultiLabelSoftMarginCriterion()
mls_loss = mls:forward(x, t)
> 27.63

mls_grad = mls:backward(x, t)
>
1e-10 *
-1.9287
[torch.DoubleTensor of size 1]

```

However, the numerically stable loss/gradient equation as described in https://www.ics.uci.edu/~pjsadows/notes.pdf (Eq 1-10) calculates the gradient 
to be **-1** and not **1e-10 *-1.9287**.
The dirty denominator which causes problems gets cancelled out in Eqs. 6 and 7. The stable equations are also utilized in caffe: https://github.com/BVLC/caffe/blob/master/src/caffe/layers/sigmoid_cross_entropy_loss_layer.cpp#L42

soumith:
hi @arunmallya . This looks like it is indeed unstable.
nn.MultiLabelSoftMarginCriterion can be short-cut to be much more stable. 
Are you willing to send a PR for this? if not i will get to it in due time.

arunmallya:
Yeah, will do.

arunmallya:
Submitted pull request: https://github.com/torch/nn/pull/953

PR #953, opened 2016-09-05
Use stable loss and gradient for MultiLabelSoftMarginCriterion

arunmallya:
This replaces the earlier MultiLabelSoftMarginCriterion consisting of a Sigmoid, followed by a BCE for loss and gradient computation. The earlier structure led to instability due to numerical precision. 

The new code combines these two steps and uses a more stable form of the loss and gradient.

This addresses issue **https://github.com/torch/nn/issues/949**.

### DIFF
--- a/MultiLabelSoftMarginCriterion.lua
+++ b/MultiLabelSoftMarginCriterion.lua
@@ -8,7 +8,7 @@
 -- and with weights:
 -- l(x,y) = - sum_i weights[i] (y[i] * log(p[i]) + (1 - y[i]) * log (1 - p[i]))
 --
---
+-- This uses the stable form of the loss and gradients.
 --]]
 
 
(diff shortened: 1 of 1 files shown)


soumith:
hi arun, i see that you completely ignored the "weights" parameter in the new implementation :)
hope that's not an oversight...

arunmallya:
Oh, my bad. Will add that in.

soumith:
@arunmallya i want to get this PR to completion. Would you have time to finish it? If not I can take it over...

arunmallya:
Would greatly appreciate if you could take over. I am a bit tied up with CVPR at the moment. Thanks!

ayumiymk:
the source doesn't update?

nicholas-leonard:
@ayumiymk I added some unit tests for weights and made it more efficient by reusing buffers. Thanks for the PR!

PR #965, opened 2016-09-13
Accept 5D weights in VolumetricConvolutionMM

apaszke:


lib/THNN/generic/VolumetricConvolutionMM.c
```diff
@@ -2,6 +2,20 @@
...
 
+static int THNN_(view_weight)(THTensor **_weight)
+{
+  THTensor *weight = *_weight;
+  THArgCheck(weight->nDimension == 2 || weight->nDimension == 5, 4,
+          "weight tensor should be 2D or 5D - got %dD", weight->nDimension);
+  if (weight->nDimension == 5) {
+    long s1 = weight->size[0];
+    long s2 = weight->size[1] * weight->size[2] * weight->size[3] * weight->size[4];
+    *_weight = THTensor_(newWithStorage2d)(weight->storage, 0, s1, -1, s2, -1);
```
fmassa: Need to add `storageOffset` here, as was included in
https://github.com/torch/nn/blob/master/lib/THNN/generic/SpatialConvolutionMM.c#L104

soumith: @apaszke same mistake again :P

soumith: done via https://github.com/torch/nn/commit/8f2b88a5a4f2b541b9efc2fb72a47572811efd17

apaszke: Damn, I must have copied a snipped from the old commit 😕 Sorry for that.

PR #966, opened 2016-09-14
[test.lua] Make i/o forward/backward tests approximate

tkoeppe:
I found these tests to be failing regularly, where the actual values are very small but non-zero.

If you can confirm that this is happening, then this patch makes the tests approximate so they pass most of the time.


soumith:
Thank you!

PR #970, opened 2016-09-20
add StochasticGradient doc back

hughperkins:


hughperkins:
(I havent checked how well this 'tells a story', but it at least gets StochasticGradient doc back, and then we can tweak it later sometime)

Atcold:
Thank you, @hughperkins! This is really nice of you.

PR #973, opened 2016-09-20
Don't serialize shared parameters in clone()

jgehring:
There's no need to serialize parameters that will be shared immediately
afterwards. In particular, the current implementation of clone and share is
inefficient for single modules in a larger network with flattened parameters
since the storage holding all parameters will be serialized.

Here, the write method of nn.Module is temporarily overwritten with a function
that writes empty tensors for all shared parameters.

### DIFF
--- a/Module.lua
+++ b/Module.lua
@@ -102,12 +102,38 @@ function Module:share(mlp, ...)
    return self
 end
 
+local function sharedWrite(...)
+   local arg = {...}
+   local shared = {}
+   for i,v in ipairs(arg) do
+       shared[v] = true
+   end
+   return function(self, file)
+      local object = {}
+      for k, v in pairs(self) do
+         if shared[k] then
+            assert(torch.isTensor(v), 'Shared parameters have to be Tensors')
+            object[k] = v.new()
+         else
+            object[k] = v
+         end
+      end
+      file:writeObject(object)
+   end
+end
+
 function Module:clone(...)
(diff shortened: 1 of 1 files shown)


soumith:
Thanks Jonas!

albanD:
@jgehring @soumith won't this break all modules that overwrite the `write` function because it assumes `local oldWrite = nn.Module.write` and thus after calling the `clone` function, the `write` function will always be the `nn.Module`'s one and not the one defined for this module anymore?
Also it assumes that the `write` function is simply write all the elements in the table and ignore any custom write such as the one from [`nn.GPU`](https://github.com/torch/nn/blob/ad4a4a1ed7ed8fddc9014c8af8474686c31cab5b/GPU.lua#L228-L237) ?

jgehring:
@albanD not sure I understand your concerns. After cloning, only `nn.Module.write` will be restored. i.e. not `self.write`. In other words, `MyCustomModule.write` won't be touched. Subclasses of `nn.Module` are free to implement their own `write` methods, they will be called as usual from within `torch.save`. All that happens is that the base implementation of `write` is being swapped out for cloning.

There's a potential issue if you're overwriting `write` in `nn.Module`'s meta-table. However, if you went down that road you might just as well re-implement `clone`.

albanD:
Hoo my bad, worst case then will be that if you reimplemented a `write` method, it will not benefit from this. Sorry for the silly question!

PR #982, opened 2016-09-27
improving error messages in THNN

soumith:
Error messages across board have been largely improved in the C library of THNN.
Now, instead of reporting that a tensor has mismatched shape, or that they have wrong dimension, the error will also report what the expected dimension was, and what the tensor shape is...

For example:
Old message:
```
bad argument #2 to 'v' (3D or 4D (batch mode) tensor expected for input
```
New message:
```
bad argument #2 to 'v' (3D or 4D (batch mode) tensor expected for input, but got: [100 x 100]
```

Some places where error messages weren't there, argchecks have been added. Most critically in SoftMax and LogSoftMax

### DIFF
--- a/lib/THNN/generic/Abs.c
+++ b/lib/THNN/generic/Abs.c
@@ -17,6 +17,7 @@ void THNN_(Abs_updateGradInput)(
           THTensor *gradOutput,
           THTensor *gradInput)
 {
+  THNN_CHECK_NELEMENT(input, gradOutput);
   THTensor_(resizeAs)(gradInput, input);
   TH_TENSOR_APPLY3(real, gradInput, real, gradOutput, real, input,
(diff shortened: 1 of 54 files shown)


andreaskoepf:
Those error message improvements are definitely very helpful. @soumith I saw you changed the indentation at a couple of places. Do you think you found some good rules that could make it into some kind of style guideline for torch-development, e.g. for placement of arguments, how to deal with long lines etc.?

soumith:
@andreaskoepf while we dont have a strict style guide, i just thought keeping it to fit within 80-90 lines seemed nicer for small screens.
The first thing we'd have to do if we had a strict style guide is to fix torch to follow it....

Atcold:
This is great for debugging. Then you @soumith!

PR #990, opened 2016-10-02
ClassNLLCriterion broken for cunn1.0-0

mys007:
The changes introduced in commit #979 have broken cutorch/cunn 1.0-0 because there is no `cudaLong` there.  This fix makes the conversion conditional.
An alternative would be to remove the ugly type castings completely, they are not used e.g. in `MultiMarginCriterion` either.

### DIFF
--- a/ClassNLLCriterion.lua
+++ b/ClassNLLCriterion.lua
@@ -29,13 +29,13 @@ end
 function ClassNLLCriterion:updateOutput(input, target)
    if type(target) == 'number' then
       if input:type() == 'torch.CudaTensor' then
-          self.target = self.target:cudaLong()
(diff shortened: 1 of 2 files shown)


soumith:
Thanks Martin. I should've versioned nn 1.0-0 as well, this was an oversight.

mys007:
For a similar reason, https://github.com/torch/cunn/pull/323 should be merged into cunn/1.0-0. Actually, are new updates going to be periodically merged into the 1.0 branch or is it considered frozen and all people should switch to the new Cuda tensor types?

soumith:
@mys007 it's considered fromzen, and folks should move to new tensor types.
Also your PR is buggy. there is never a torch.cudaLong variable that exists (even in 1.0-0).
I'm fixing the PR to check for the existence of torch.CudaLongTensor

PR #991, opened 2016-10-02
Specify output height and width instead of scale

arunmallya:
Optionally pass in a table containing output height and width instead of scale.
This is useful when there is no exact integer scale that can produce the required output size. 

```
module = nn.SpatialUpSamplingBilinear(scale)
or
module = nn.SpatialUpSamplingBilinear({oheight=H, owidth=W})
```

### DIFF
--- a/SpatialUpSamplingBilinear.lua
+++ b/SpatialUpSamplingBilinear.lua
@@ -8,21 +8,28 @@ input planes.
 
 The Y and X dimensions are assumed to be the last 2 tensor dimensions.  For
 instance, if the tensor is 4D, then dim 3 is the y dimension and dim 4 is the x.
-scale_factor is assumed to be a positive integer.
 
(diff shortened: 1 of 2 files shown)


fmassa:
Hi,

I still think that allowing a second input argument to the forward function that specifies the size of the output (as in `SpatialFullConvolution`) would be more general, as it would allow for varying input sizes to work correctly. 
This was discussed in https://github.com/torch/nn/pull/882#issuecomment-235901664

arunmallya:
A quick question: I was using the above layer in an nngraph. Doesn't passing in a table in the forward prevent the layer from being used in this manner in an nngraph? Is there any way of using the table-input forward in an nngraph forward?

soumith:
i'm merging this, as it's beneficial as-is as well.

PR #992, opened 2016-10-02
More broken compatibility with cutorch 1.0

mys007:
This extends https://github.com/torch/nn/pull/979


soumith:
this is broken. torch.cudaLong does not really exist. Check for torch.CudaLongTensor instead

soumith:
See https://github.com/torch/nn/commit/0a0c589dfe7ca696316e1deccfe483ed45ab930e

mys007:
I'm really sorry for the introduced bug. I've fixed this PR.

soumith:
Thanks!

PR #996, opened 2016-10-05
Add weight exception to doc

fchouteau:
Update documentation: Weights work for ClassNLLCriterion and CrossEntropyCriterion only if sizeAverage is set to false.


soumith:
Thanks!

PR #997, opened 2016-10-05
Fix Bilinear crash when backward is called after clearState.

adityaramesh:
[This MWE](https://gist.github.com/adityaramesh/e31f9bafee2a85243f1d4714fd3cb1ff) shows that Bilinear module crashes when `backward` is called after `clearState`. I modified `updateGradInput` in `Bilinear.lua` to repopulate `self.gradInput` after it is emptied by `clearState`. This fixes the crash.

### DIFF
--- a/Bilinear.lua
+++ b/Bilinear.lua
@@ -86,7 +86,8 @@ function Bilinear:updateGradInput(input, gradOutput)
       self:__assertInputGradOutput(input, gradOutput)
 
       if #self.gradInput == 0 then
-          for i = 1, 2 do self.gradInput[i] = input[1].new() end                                                                                                                                               end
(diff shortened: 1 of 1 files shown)


Bilinear.lua
```diff
@@ -84,6 +84,10 @@ function Bilinear:updateGradInput(input, gradOutput)
    assert(self)
    if self.gradInput then
       self:__assertInputGradOutput(input, gradOutput)
+
+      if #self.gradInput == 0 then
```
fmassa: This `if` need to be closed.

adityaramesh: Sorry about that, I missed the last line when I was copying the fix from somewhere else. I just fixed this.

soumith:
Thank you!

PR #998, opened 2016-10-08
CAdd module for component-wise addition

alykhantejani:
Added a CAdd module for component-wise addition analogous to the CMul module. 

Applies a component-wise addition to the incoming data, i.e. `y_i = x_i + b_i` across any or multiple dimensions. If the size for a particular dimension is 1, the multiplication will be expanded along the entire axis.


soumith:
Thanks Alykhan! Looks pretty good.

PR #1004, opened 2016-10-16
Ported Linear module to THNN

lantiga:
This PR implements a port of `Linear.lua` to `THNN`.
I have tried to keep the implementation as close as possible to the original code and to adopt a coding style that fits `THNN`.
All tests in `test.lua` pass.

### DIFF
--- a/Linear.lua
+++ b/Linear.lua
@@ -9,6 +9,7 @@ function Linear:__init(inputSize, outputSize, bias)
       self.bias = torch.Tensor(outputSize)
       self.gradBias = torch.Tensor(outputSize)
    end
+   self.addBuffer = torch.Tensor(outputSize)
    self:reset()
 end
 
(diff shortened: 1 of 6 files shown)


soumith:
two quick comments.
1. I think Linear.lua should still be lua based, as it's a simple module and is nicely hackable and changeable from Lua.

2. THNN should (and now with this PR) has a Linear layer.

So, revert the lua changes, but keep the THNN changes?

What do you think, does that sound reasonable?

lantiga:
Sure, this make sense. I'm happy as long as a THNN implementation is available :-)
I'll revert the Lua part, not sure how to still test the THNN part through test.lua though. Should I add a separate LinearTHNN.lua? Suggestions welcome.

soumith:
i think you can keep the current version in the PR in test/testTHNN.lua or something.

lantiga:
Good, I'll look into it.
One last thing, in view of future PR's. Should I proceed this way or should we make a list of modules that will have a dual implementation?

soumith:
i think depending on the module, we'd move it completely into THNN, or keep the lua-bits and THNN bits.
If you make a list of modules you intend to THNN-ify, we can quickly discuss them and you can proceed.

lantiga:
Sounds good, I'll do that.

I also have to see why SparseLinear (which I didn't touch) just failed on TravisCI on one of the six configurations https://travis-ci.org/torch/nn/jobs/168067751 ...

soumith:
the tests are a bit brittle. dont worry about the failure.

soumith:
hmm, actually that SparseLinear failure looks not just a one-off. Looking into it.

lantiga:
appreciated!

lantiga:
Ran all tests multiple times locally, I can't seem to reproduce (macOS).

soumith:
It seems to be a one off, don't worry about it

lantiga:
👍🏻

lantiga:
Reverted changes to Linear.lua

lantiga:
I added a `testTHNN` function to `test.lua` that will require THNN-only modules in the `test` directory and use them to replace their "official" Lua counterparts.
To use it, just do e.g.
```
th -lnn -e "nn.testTHNN{'Linear'}"
```

soumith:
Thanks Luca!

lantiga:
Thanks for merging!

PR #1005, opened 2016-10-16
PixelShuffle module for sub-pixel convolution

alykhantejani:
Added the PixelShuffle layer which rearranges elements in a tensor of shape `[C*r, H, W]` to a tensor of shape `[C, H*r, W*r]`. 

This is useful for implementing efficient sub-pixel convolution with a stride of `1/r`, which can be used for tasks such as super-resolution (see ["Real-Time Single Image and Video Super-Resolution Using an Efficient Sub-Pixel Convolutional Neural Network" - Shi et al.](https://arxiv.org/abs/1609.05158)).


soumith:
Thank you! :)

PR #1009, opened 2016-10-19
Support generic type Spatial modules

gchanan:
This involves two main changes:
1) Convert indices in C code to Index types so they don't use half types, which aren't precise enough for indices.
2) Convert index tensors to cudaLongs when run in cuda in lua code.

NOTE: this requires https://github.com/torch/cunn/pull/346 at the same time or cunn will be broken.


gchanan:
This includes all changes necessary to get C-code cunn modules to run with generic floating point types.  It is possible that there are other required changes to get all modules to run correctly with generic floating point types, e.g. in lua only code, as I only ran the C-code tests to verify.

NOTE: as above this requires merging torch/cunn#346 at the same time or cunn will be broken.

PR #1010, opened 2016-10-20
added support for tables as inputs to Concat

lukasc-ch:
1) added support for tables as inputs to Concat, 
2) removed some code redundancy (similar to ConcatTable)

### DIFF
--- a/Concat.lua
+++ b/Concat.lua
@@ -1,4 +1,4 @@
-local Concat, parent = torch.class('nn.Concat', 'nn.Container')
+local Concat, parent = torch.class('iis.Concat', 'nn.Container')
 
 function Concat:__init(dimension)
    parent.__init(self)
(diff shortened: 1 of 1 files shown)


soumith:
hey Lucas, I wonder why you added table logic to nn.Concat, instead of using ConcatTable?

lukasc-ch:
Hi @soumith 
ConcatTable and Concat have different functionality: ConcatTable adds the results of its children modules' outputs to a table, where Concat concatenates its children modules' output into a tensor along some dimension (i.e. different behavior on how to create the output). 
Before this PR, some functions of Concat start with self.gradInput:resizeAs(input). This is a problem in case of e.g. (not doing something useful, but minimal working example)
```
m = nn.Sequential()
    :add(nn.ConcatTable()
        :add(nn.Identity())
        :add(nn.Identity()))
    :add(nn.Concat(2)
        :add(nn.SelectTable(1))
        :add(nn.SelectTable(2)))
```
which would then crash during the backward call.

fmassa:
I usually used `nn.ConcatTable` followed by `nn.JoinTable` in those situations. Is the merged `nn.Concat` in this PR more memory efficient than the two modules separately?

lukasc-ch:
Hi Francisco
It should use the same amount of memory, but be more versatile. The above case could indeed be handled by ConcatTable and JoinTable, but consider
```
m = nn.Sequential()
    :add(nn.ConcatTable()
        :add(nn.Identity())
        :add(nn.Identity()))
    :add(nn.Concat(2)
        :add(nn.SelectTable(1))
        :add(nn.SelectTable(2))
        :add(nn.SelectTable(2)))
```
or more complex stuff with hierarchical tables
```
m = nn.Sequential()
    :add(nn.ConcatTable()
        :add(nn.Identity())
        :add(nn.ConcatTable()
            :add(nn.Identity())
            :add(nn.Identity()))
    :add(nn.ConcatTable()
        :add(nn.Concat(2)
            :add(nn.SelectTable(1))
            :add(nn.Sequential()
                :add(nn.SelectTable(2))
                :add(nn.SelectTable(1))))
         :add(nn.Sequential()
            :add(nn.SelectTable(2))
            :add(nn.SelectTable(2))))
```

lukasc-ch:
Is there anything preventing this from getting merged? Is it still unclear what functionality is add by this PR / should I explain more?

soumith:
thanks @lukasc-ch .

Issue #1012, opened 2016-10-24
A bug in CAdd.lua, when running in batch model.


qiuxin2012:
When running this test code
```
input = torch.rand(4, 5)
gradOutput = torch.rand(4, 5)
cadd = nn.CAdd(5)
output = cadd:forward(input)
gradInput = cadd:backward(input, gradOutput)
```
I get below error
```
torch/install/share/lua/5.1/torch/Tensor.lua:413: The number of covered elements is not a multiple of all elements.
stack traceback:
	[C]: in function 'assert'
	/home/xin/torch/install/share/lua/5.1/torch/Tensor.lua:413: in function 'specifyFully'
	/home/xin/torch/install/share/lua/5.1/torch/Tensor.lua:455: in function 'view'
	/home/xin/torch/install/share/lua/5.1/nn/CAdd.lua:91: in function 'accGradParameters'
	/home/xin/torch/install/share/lua/5.1/nn/Module.lua:32: in function 'backward'
	[string "gradInput = cadd:backward(input, gradOutput)"]:1: in main chunk
	[C]: in function 'xpcall'
	/home/xin/torch/install/share/lua/5.1/trepl/init.lua:670: in function 'repl'
	.../xin/torch/install/lib/luarocks/rocks/trepl/scm-1/bin/th:199: in main chunk
```

In the line 91 of CAdd.lua, ` self._gradBias:view(self.gradBias, batchSize, -1)` should be `self._gradBias:view(self.gradBias, 1, -1)`.

qiuxin2012:
@alykhantejani  Could you help me check this issue?

fmassa:
I wonder if you shouldn't define the module as `cadd = nn.CAdd(1,5)`. Adding the 1 makes it work for me

qiuxin2012:
@fmassa Thanks for your reply.
Your solution is a good work around. Set `cadd = nn.CAdd(1,5)` will go to line 88, not line 91. 

But I still think we had better fix line 91.

alykhantejani:
@qiuxin2012 I think you're right, it should be `self._gradBias:view(self.gradBias, 1, -1)` will submit a fix today

qiuxin2012:
@alykhantejani Thanks very much.

PR #1013, opened 2016-10-24
hotfix for bug in CAdd

alykhantejani:
Hotfix for bug reported [here] (https://github.com/torch/nn/issues/1012#issuecomment-255679397).

Sorry for the large diff, I also converted 4 spaces -> 3 spaces (sorry I missed this in the first PR). The non-whitespace changes are in the `accGradParameters` function, where it now correctly handles the scenario when the input tensor is larger than the `bias` inside of the `CAdd` module. 

Have added a test for this case too.


soumith:
Thanks!

PR #1019, opened 2016-11-03
Fix output formula in SpatialDilatedConvolution

ankitvgupta:
Fixed a small typo in the output width and height formulas in SpatialDilatedConvolution.

 "+" should be a "-"

### DIFF
--- a/doc/convolution.md
+++ b/doc/convolution.md
@@ -478,8 +478,8 @@ The parameters are the following:
 If the input image is a 3D tensor `nInputPlane x height x width`, the output image size
 will be `nOutputPlane x oheight x owidth` where
 ```lua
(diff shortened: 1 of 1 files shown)


fmassa:
I'm not sure this fix is right.
The formula in the documentation seems to be consistent with what is computed in the C code, as seen [here](https://github.com/torch/nn/blob/master/lib/THNN/generic/SpatialDilatedConvolution.c#L46-L47).

ankitvgupta:
Hey @fmassa 

I believe there are an extra set of parentheses in the C version.

ankitvgupta:
No problem! Let me know if there is anything else I need to do to get this merged (first time contributing).

soumith:
Thanks Ankit

PR #1024, opened 2016-11-08
Don't jongjmp out of omp loops in unpooling modules

apaszke:
Even though these functions have bounds checks they segfault sometimes for invalid indices. It's probably because of calling `longjmp` in `THError`, while still being inside an OpenMP loop - according to the specs the parallel region has to execute completely (even `break` is illegal).


soumith:
Thanks!

PR #1028, opened 2016-11-11
make CMaxTable go right with cuda

anoidgit:
use .new() rather than torch.Tensor() which will return a Tensor with default Tensor type.


soumith:
thanks!

anoidgit:
@soumith It is my pleasure.

PR #1030, opened 2016-11-11
Fix checking and spacing of dilation parameters

gchanan:
in SpatialDilatedConvolution and SpatialDilatedMaxPooling.


soumith:
whoops! thanks!

PR #1031, opened 2016-11-14
PairwiseDistance: Default using Euclidean distance.

sagarwaghmare69:
By default using Euclidean/l2 distance/norm.


soumith:
Thanks!

PR #1042, opened 2016-11-21
Update containers.md

trypag:
Improve the global readability of this documentation by fixing a few typos and indentation.
Also I added so more comments in the Parallel container section, which I think is more difficult to understand than others.


trypag:
I don't think this build error is related to my PR, considering I have only modified .md files.

soumith:
Thanks!

trypag:
Thank you Soumith !

PR #1044, opened 2016-11-21
DistanceRatioCriterion

sagarwaghmare69:
Distance ratio criterion for training siamese model with triplets. With unit tested, README updated with sample example. Commits rebased.


soumith:
rebased at https://github.com/torch/nn/pull/1044
your commit has been preserved

PR #1049, opened 2016-11-26
Fix shapeCheck in Spatial Pooling modules

fmassa:
There was a problem with `shapeCheck` sometimes when `ceil_mode=True`, as [this part was missing from the size estimates](https://github.com/torch/nn/blob/master/lib/THNN/generic/SpatialDilatedMaxPooling.c#L201-L208).
The following snippet didn't work prior to this change.
```lua
m = nn.SpatialMaxPooling(2,2,2,2,1,1):ceil()
m:forward(torch.rand(1,1,5,5))
m:backward(torch.rand(1,1,5,5),torch.rand(1,1,3,3))
```
with error message
```
Need indices of dimension 4 and indices.size[3] == 4 but got indices to be of shape: [1 x 1 x 3 x 3]
```

The same change should be made in `THCUNN`. I'll send a PR later.


soumith:
Thanks Francisco, please send a pr to thcunn soon

PR #1057, opened 2016-12-01
Improve fix/shape checks for Temporal and Spatial modules

gchanan:
1) Adds shape checking for Temporal Convolution (this already exists in cunn, but was skipped in nn).
2) Adds gradOutput shape checks for temporal modules
3) Removes unnecessary (too restrictive and/or redundant) shape checks in Spatial Pooling modules.


soumith:
thanks!

PR #1063, opened 2016-12-08
Update convolution.md

willfrey:
The TemporalSubSampling equation had the wrong indexing.

The equation previously read:
output[i][t] = bias[i] + weight[i] * sum_{k=1}^kW input[i][dW*(t-1)+k]

but should read:
output[t][i] = bias[i] + weight[i] * sum_{k=1}^kW input[dW*(t-1)+k][i]


soumith:
thanks!

PR #1064, opened 2016-12-08
Add KSparse layer for KSparse autoencoders

MrMojoR:
The code implements this article's idea of k-sparse layer for autoencoders: https://arxiv.org/abs/1312.5663


soumith:
you need to add this to init.lua and add tests.
After that, I will reopen the PR.
Thanks.

PR #1073, opened 2016-12-19
Masking feature for nn.Linear and nn.SpatialConvolution

evcu:
I am implementing network pruning/compression from the `Learning both Weights and Connections for Efficient Neural Networks` paper as my final project. I would like to add pruning feature to those layers by setting a mask for each of those layers. 

There was couple of ideas in my mind. I first try those by adding a hook on training and saw that it works. Then there was two easy options;

- Implement it as a separate module `nn.PrunedLinear`. 
- Modify individual layers by adding the necessary lines.

I chose the second one to enable training first with all parameters, then iterating between pruning and retraining straightforward.

If the mask is same size as the weight tensor, then it is applied. Gradients are correctly calculated with pruned(zero-set) weights. Then right after accumulating the grad for all weights, the masked applied to the gradients to prevent grad update on pruned connections. I hope in future this can be implemented with sparse-tensors and

### DIFF
--- a/Linear.lua
+++ b/Linear.lua
@@ -1,5 +1,4 @@
 local Linear, parent = torch.class('nn.Linear', 'nn.Module')
-
 function Linear:__init(inputSize, outputSize, bias)
    parent.__init(self)
    local bias = ((bias == nil) and true) or bias
@@ -16,6 +15,19 @@ function Linear:noBias()
    self.bias = nil
    self.gradBias = nil
    return self
+end
+
+function Linear:setMask(mask)
+   if self.weight:isSameSizeAs(mask) then
+      self.mask = mask
+      self.weight:cmul(self.mask)
+   else
+      error('Size of Mask must be same as the weight tensor')
+   end
+end
+
+function Linear:removeMask(mask)
+      self.mask = nil
 end
 
 function Linear:reset(stdv)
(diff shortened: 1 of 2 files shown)


evcu:
How should I submit the tests? Should I add them to change the test.lua file? I have the two Jacobian Tests and two demo as a itorch notebook.

apaszke:
Isn't `nn.PartialLinear` the same thing as `nn.PrunedLinear`?

evcu:
I am not sure in which context the `PartialLinear` is used, but it is not same. Because Partial Linear prunes the **all** connections going to an output neuron. An example would be:
```lua
require 'nn'
m = nn.PartialLinear(3,5) 
 input = torch.randn(10,3) 
print(m:forward(input)) --[torch.DoubleTensor of size 10x5]
m:setPartition(torch.Tensor({2,4}))
print(m:forward(input)) --[torch.DoubleTensor of size 10x2]
```

I think implementing a separate module for this simple masking would be overkill. Therefore I've committed changes on `nn.Linear` and `nn.SpatialConvolution`. They are simple and easy to use.

ClementPinard:
I also happen to work on the implementation of Song Han's paper, but i only used a big mask for all the parameters. You then set the learningRates directly on optimState.
```
parameters, gradParameters = model : getParameters()
mask = torch.ne(parameters,0)
optimState.learningRates=mask:double()
```

soumith:
i dont consider this an essential into the core. Subclassing makes sense here. Closing.

PR #1075, opened 2016-12-21
added CReLU transfer function + tests

alykhantejani:
Added the CReLU transfer function as described in Understanding and Improving Convolutional Neural Networks via Concatenated Rectified Linear Units (https://arxiv.org/abs/1603.05201).


alykhantejani:
Fixing merge conflict and will update this PR

alykhantejani:
Merged and fixed global function declaration in `test.lua`

CReLU.lua
```diff
@@ -0,0 +1,51 @@
...
+function CReLU:updateOutput(input)
+   self.batched = true
+   if input:nDimension() == 3 then
+      self.batched = false
+   end
+
+   if self.batched then
+      parent.updateOutput(self, input)
+   else
+      parent.updateOutput(self, input:view(1, input:size(1), input:size(2), input:size(3)))
```
soumith: you can fix CReLU to not depend on 3D and 4D inputs.
Considering that it is CReLU and not SpatialCReLU, please do the following:

- remove the self.batched checks...
Do:
```
function CReLU:updateOutput(input)
   parent.updateOutput(self, input:view(1, -1))
   self.output:resizeAs(input)
   return self.output
end
```
make similar changes for updateGradInput

alykhantejani: @soumith  Thanks for the suggestion, I think that's a cool way to do it. However, I think this works fine for the `updateGradInput`, but in the `updateOutput` the `output` size has twice as many feature maps (channels), so we would still need to know if it's batched or not to do the `resize` if we went down this route?

I've pushed the changes to `updateGradInput`, let me know if you can think of a simpler way to do the `updateOutput`

soumith: ok, in that case, there's a lot of precedence for this. The constructor to nn.CReLU needs to take an argument nInputDims: https://github.com/torch/nn/search?utf8=%E2%9C%93&q=nInputDims

Then, updateOutput will become:
```lua
function CReLU:updateOutput(input)
   local input_
   local batched = input:dim() == self.nInputDims
   if not batched then
      input_ = input:view(1, -1)
  else
      input_ = input:view(input:size(1), -1)
  end
   parent.updateOutput(self, input)
   local osize = input:size()
   if not batched then
       osize[1] = osize[1] * 2
   else
      osize[2] = osize[2] * 2
   end
    self.output:resize(osize)
   return self.output
end
```

What do you think about this?

alykhantejani: looks good to me. Didn't know that the convention was to have `nInputDims` as an input, good to know - thanks!

Will make the changes and update the PR

alykhantejani: @soumith I had a look at the use cases of `nInputDims`, I'm a bit confused about how it should be used with `CReLU`. For example, if I want to perform `CReLU` over a 4-dim tensor (batched) I would initialize the module with `nn.CReLU(4)`. But if I want to do it over a 3-dim tensor (non-batched), would I use `nn.CReLU(3)`, if so in the implementation above for an input of size `[3,4,5]` the output would be of size `[3,8,5]` not the expected `[6,4,5]`

I think I'm just misunderstanding how `nInputDims` should be interpreted

soumith: nInputDims = number of non-batch input dims.
So, nInputDims=3 in your case.
When it gets batched input (i.e. 4D), let's say [2,3,4,5], then it gives out [2,6,4,5], as `osize[2] = osize[2] * 2`.
When it gets non-batched input [3,4,5], it gives out [6,4,5], as `osize[1] = osize[1] * 2`

alykhantejani: @soumith  thanks for the help so far. I've made the changes to `updateOutput` to use `nInputDims` and also made the following changes:

- Changed the test to no longer check that the first `x/2` feature maps were `ReLU(input)` and the second `x/2` were `ReLU(-input)` as now the order is not guaranteed, instead I check that the output feature maps contains all of the ones from `ReLU(input)` and `ReLU(-input)`

- I added `CReLU:backward` as well as `CReLU:updateGradInput` as `CReLU` is of type `Sequential` and `Sequential.backward` doesn't call `updateGradInput`, but `Jacobian.testJacobian` does, so both need to be defined (if I'm not mistaken, please let me know if you agree)

- Updated docs to reference the `nInputDims` parameter.

CReLU.lua
```diff
@@ -0,0 +1,66 @@
...
+      osize[1] = osize[1] * 2
+   else
+      osize[2] = osize[2] * 2
+   end
+   self.output:resize(osize)
+   return self.output
+end
+
+function CReLU:backward(input, gradOutput)
+   local batched = input:dim() == (self.nInputDims + 1)
```
soumith: delete the contents of `function CReLU:backward(input, gradOutput)`
replace them with:
```lua
function CReLU:backward(input, gradOutput)
   return self:updateGradInput(input, gradOutput)
end
```

soumith:
okay made a last minor comment, after that this is good to go.

alykhantejani:
Cool, have made the change. Thanks for your help :)

soumith:
thanks!

PR #1076, opened 2016-12-22
Add SpatialAutoCropMSECriterion

alykhantejani:
Added the `SpatialAutoCropMSECriterion` which performs MSECriterion even if the input has a smaller spatial resolution than the target. This is achieved by cropping the target to the same size as the input tensor.

This is useful for image regression tasks, such as super resolution in which non-padded convolutions reduce the output resolution from your network. This layer allows you to still perform MSE with the center of the target image

### DIFF
--- a/SpatialAutoCropMSECriterion.lua
+++ b/SpatialAutoCropMSECriterion.lua
@@ -1,0 +1,74 @@
+--[[
+   SpatialAutoCropMSECriterion.
+   Implements the MSECriterion when the spatial resolution of the input is less than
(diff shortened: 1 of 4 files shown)


soumith:
had to fix a bunch of global var leaks on your PR, but fixed rebased and merged into master now.
See: https://github.com/torch/nn/commit/ee2d79daa1985d5d34d43558b67371c981f5d8ff

Generally, it's easy to check for such leaks by running:
th -g -lnn -e "nn.test()"

alykhantejani:
Thanks, and thanks for the pointer!

PR #1092, opened 2017-01-03
Avoid function serialization when sharing parameters

guillaumekln:
Overwriting the `accUpdateGradParameters` field exposed the function to serialization and made the saved object incompatible with other Lua versions.

This could be avoided with a simple conditional branching to call the shared version of this function.


soumith:
thanks Guillaume!

PR #1100, opened 2017-01-10
Added TemporalRowConvolutionMM layer, tests, and documentation

willfrey:
Added row convolutions (a.k.a. lookahead convolutions) as described in [this paper](http://www.cs.cmu.edu/~dyogatam/papers/wang+etal.iclrworkshop2016.pdf).


willfrey:
Any advice for making this work for luajit2.1 and gcc4.6.3?

My first hunch is that memory isn't getting properly zeroed out when allocated but I can't see where that might be happening.

Going to try to start up a VM and reproduce this error.

willfrey:
This is ready for you to review.

lib/THNN/generic/TemporalRowConvolutionMM.c
```diff
@@ -0,0 +1,488 @@
...
+
+	THNN_CHECK_DIM_SIZE(input, ndim, dimF, inputFrameSize);
+
+	if (gradOutput != NULL) {
+		THNN_CHECK_DIM_SIZE(gradOutput, ndim, dimF, inputFrameSize);
+		THNN_CHECK_DIM_SIZE(gradOutput, ndim, dimS, nOutputFrame);
+	}
+}
+
+static int THNN_(view_weight_rowconv)(THTensor **_weight) {
```
soumith: why is this function really needed?
You are defining this module afresh, you can control the exact shape of the weight. Just define the weight shape as you think is appropriate and let's stick to it.

lib/THNN/generic/TemporalRowConvolutionMM.c
```diff
@@ -0,0 +1,488 @@
...
+	int ndim = input->nDimension;
+	int dimS = 0;
+	int dimF = 1;
+
+	if (ndim == 3) {
+		++dimS;
+		++dimF;
+	}
+
+	THTensor *tinput = THTensor_(newTranspose)(input, dimS, dimF);
```
soumith: the way you wrote this function with all the transposes, it'd be a shame to not also have the function support feature-dimension be 2nd dim optionally (the way it's in pytorch, i.e. Batch x Channels x Width), because when you want to use this in to pytorch, Conv1d's output can be sent directly into this.
Consider adding a boolean flag to the function signature that controls this transposition

TemporalRowConvolution.lua
```diff
@@ -0,0 +1,129 @@
...
+  return self
+end
+
+function TemporalRowConvolution:reset(stdv)
+  if stdv then
+    stdv = stdv * math.sqrt(3)
+  else
+    stdv = 1 / math.sqrt(self.kW * self.inputFrameSize)
+  end
+  if nn.oldseed then
```
soumith: you can remove the oldseed stuff

willfrey: Done.

lib/THNN/generic/TemporalRowConvolutionMM.c
willfrey: Removed the function and fixed the weight to a static shape in the Lua interface.

lib/THNN/generic/TemporalRowConvolutionMM.c
willfrey: It now supports `nBatchFrame x inputFrameSize x nInputFrame` by supplying an optional fourth argument `featFirst = true` or by setting `module.featFirst = true`.

soumith:
thanks! :)

PR #1103, opened 2017-01-13
Add SpatialAdaptiveAveragePooling.

ruotianluo:
Most of the code is based on SpatialAdaptiveMaxPooling, only changing max pooling to average pooling.


lib/THNN/generic/SpatialAdaptiveAveragePooling.c
```diff
@@ -0,0 +1,249 @@
...
+#pragma omp parallel for private(k)
+  for (k = 0; k < nslices; k++)
+  {
+    /* loop over output */
+    long i, j;
+    for(i = 0; i < oheight; i++)
+    {
+      int y_start = (int)floor((float)i / oheight * iheight);
+      int y_end   = (int)ceil((float)(i + 1) / oheight * iheight);
+      int kH = y_end-y_start;
```
ruotianluo: Is it better to use
int y_start = (int)floor(float(i * iheight) / oheight);
int y_end   = (int)ceil((float)((i + 1) * iheight) / oheight);
or 
int y_start = i * iheight / oheight;
int y_end = i * iheight / oheight + (i * iheight % oheight)>0?1:0;

The first one is faster and more accurate than the original implementation (but still involves float number calculation). The second is uglier but accurate.

fmassa:
This looked good. Any special reason why you closed it?
Also, do you plan to add a cuda version?

ruotianluo:
I made some little change.  The cuda version is almost ready, I will make the pull requests tonight.

ruotianluo:
#1105 torch/cunn#418

PR #1106, opened 2017-01-16
noBias function for VolumetricConvolutions and VolumetricFullConvolutions

shubhamjain0594:
Added noBias function for VolumetricConvolutions and VolumetricFullConvolutions. Also added tests for them and modified the library .c files for it to handle null bias.


shubhamjain0594:
The commits and branch name not aligned with CONTRIBUTING.md. Closing.

PR #1107, opened 2017-01-16
Added noBias module for VolumetricConvolution and VolumtericFullConvo…

shubhamjain0594:
Added noBias function for VolumetricConvolutions and VolumetricFullConvolutions. Also added tests for them and modified the library .c files for it to handle null bias.

### DIFF
--- a/VolumetricConvolution.lua
+++ b/VolumetricConvolution.lua
@@ -1,3 +1,4 @@
+local THNN = require 'nn.THNN'
 local VolumetricConvolution, parent = torch.class('nn.VolumetricConvolution', 'nn.Module')
 
 function VolumetricConvolution:__init(nInputPlane, nOutputPlane, kT, kW, kH, dT, dW, dH, padT, padW, padH)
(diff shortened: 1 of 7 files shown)


soumith:
this PR is in a position where I can only merge it in if there's it's cunn equivalent, or else cunn will fail.
Any plans of adding the cunn PR too?

shubhamjain0594:
Will add it, and let you know when done.

elistevens:
FYI, I've used the `.c` changes here as the starting point for this PR: https://github.com/pytorch/pytorch/pull/688 The CUDA changes are also implemented there (but obviously none of the Lua stuff).

It wasn't clear how to keep attribution, so I just copied the files and referenced this PR in the commit. It sounds like git subtree is used, and so the backend changes should show up here if my PR is accepted. I think.

Cheers!

shubhamjain0594:
I was in process of submitting a cudnn and cunn PR but if it is done then I will just submit the lua wrappers for it. If there is a better way this PR can be submitted in whole, then it would be great. 

@soumith let us know what do you think?

PR #1109, opened 2017-01-18
Using accreal instead of real in the API

pavanky:
This is done to be consistent with the changes made to cunn

### DIFF
--- a/THNN.lua
+++ b/THNN.lua
@@ -73,6 +73,13 @@ local replacements =
     }
 }
 
+-- gsub(s, 'real', 'float') changes accreal to accfloat.
+-- typedef accfloat ahead of time.
+ffi.cdef("typedef double accfloat;")
+-- gsub(s, 'real', 'double') changes accreal to accfloat.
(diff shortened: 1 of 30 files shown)


pavanky:
This PR depends on: https://github.com/torch/torch7/pull/915

THNN.lua
```diff
@@ -73,6 +73,13 @@ local replacements =
     }
 }
 
+-- gsub(s, 'real', 'float') changes accreal to accfloat.
+-- typedef accfloat ahead of time.
+ffi.cdef("typedef double accfloat;")
```
gchanan: shouldn't accfloat be float?

pavanky: It is `double` in `THGenerateFloatType.h`

gchanan: ah you are correct, didn't realize they were different on cpu vs gpu.

pavanky: But it is `float` in the CUDA version for performance reasons. I talked to soumith about it and decided to not change the way they are right now.

PR #1110, opened 2017-01-20
added c implementation of GatedLinearUnit

huihuifan:


soumith:
missing tests and docs, you must've forgot to include them in the PR.

PR #1114, opened 2017-01-24
Remove redundant zeroing (nn.Linear)

zhtvk:
Hi all,

Thank you for your work on Torch! I would like to contribute this small patch for nn.Linear. It seems that the zeroing of self.output and self.gradInput is not necessary here as it will be performed in the addmm/addmv ops.

Kind regards,
Viktor


soumith:
this is needed for certain BLAS libraries that actually do 0 * destination + 1 * mm, where any nans in the destination would break things. That's why we have it here.

soumith:
(the zeroing historically did not exist, but we added it after some bug reports)

zhtvk:
Thanks @soumith, wasn't aware of this.

PR #1124, opened 2017-02-03
Support index parameters with Cuda*Tensor.

gchanan:
These modules check the type explicitly for CudaTensor
rather than the regular expression.  There are other
modules that have this identical check, but do operations
on the resulting type which may or may not be supported
(i.e. require more checking), so I didn't change them.
These should be safe because the index type is independent
of the type being checking.

### DIFF
--- a/Max.lua
+++ b/Max.lua
@@ -21,7 +21,7 @@ end
 function Max:_lazyInit()
    self._output = self._output or self.output.new()
    if not self._indices then
-      if torch.type(self.output) == 'torch.CudaTensor' then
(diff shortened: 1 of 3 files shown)


soumith:
@gchanan wtf we see https://travis-ci.org/torch/nn/jobs/198101969 on this PR. this looks similar to this behavior we saw on cunn

soumith:
oh, looks like TemporalRowConvolution can have REALLY LARGE errors. I suspect a buffer used before zero-ing it. cc: @willfrey

willfrey:
I'll look into it!

PR #1126, opened 2017-02-04
Bugfix/rowconv

willfrey:
Travis successfully built this twice. I think I've zeroed out all the buffers that need zeroing.


soumith:
did you figure out the actual bug? I see no fix in the code, just reordering of code.

willfrey:
Seems like it was a race condition

PR #1129, opened 2017-02-08
Support Tensor constant

jnhwkim:
Now, `constant_scalar` in `nn.AddConstant` can be a 1-D tensor. It is useful when a fixed bias is needed. And, polished indents.

### DIFF
--- a/AddConstant.lua
+++ b/AddConstant.lua
@@ -3,7 +3,7 @@ local AddConstant, parent = torch.class('nn.AddConstant', 'nn.Module')
 function AddConstant:__init(constant_scalar,ip)
    parent.__init(self)
    self.constant_scalar = constant_scalar
-   if type(self.constant_scalar) == 'userdata' then
+   if torch.isTensor(self.constant_scalar) then
       self.tmp = constant_scalar.new()
    end
 
--- a/AddConstant.lua
+++ b/AddConstant.lua
@@ -1,10 +1,12 @@
 local AddConstant, parent = torch.class('nn.AddConstant', 'nn.Module')
 
 function AddConstant:__init(constant_scalar,ip)
-  parent.__init(self)
-  assert(type(constant_scalar) == 'number', 'input is not scalar!')
-  self.constant_scalar = constant_scalar
-  
+   parent.__init(self)
+   self.constant_scalar = constant_scalar
+   if type(self.constant_scalar) == 'userdata' then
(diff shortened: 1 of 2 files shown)


soumith:
two quick fixes that are needed:

1. checking if constant_scalar is a tensor is not done with type(constant_scalar) == 'userdata'. instead, use torch.isTensor

2. creating new tensors shouldn't be done via torch.Tensor(), but via x.new(). in this case, self.tmp = constant_scalar.new() where appropriate

jnhwkim:
@soumith thanks for your kind advice. could you elaborate the second note? that's what I should understand correctly.

AddConstant.lua
```diff
@@ -1,37 +1,51 @@
...
+      error('in-place flag must be boolean')
+   end
+end
+
+function AddConstant:updateOutput(input)
+   assert(type(self.constant_scalar) == 'number' or
+      (torch.isTensor(self.constant_scalar) and input:nDimension() <= 2 and
+      input:size(input:nDimension()) == self.constant_scalar:size(1)),
+      'input is not scalar or doesn\'t match with the dimension of constant!')
+   if type(self.constant_scalar) == 'userdata' and input:nDimension() == 2 then
```
soumith: you still haven't changed this

AddConstant.lua
```diff
@@ -1,37 +1,51 @@
...
+      local nOutput = self.constant_scalar:size(1)
+      self.tmp:resize(1,nOutput)
```
soumith: actually, can you get rid of self.tmp, and just allocate a tensor: `local tmp` on every forward pass. Because we have caching allocator, we dont need to pre-allocate tensors.

AddConstant.lua
jnhwkim: @soumith updated!

AddConstant.lua
jnhwkim: @soumith very nice! I updated as you mentioned. however, I have no idea about the difference between `.new()` and `torch.Tensor()`. Is creating a new tensor via `torch.Tensor()` a bad practice, since it postponed proper memory allocation, while `.new()` does allocate a memory space according to a given tensor?

soumith:
torch.Tensor creates a double tensor. .new() woll create a tensor of the same type as the original tensor.

jnhwkim:
@soumith that makes sense! I've learned from you! ;-)

PR #1131, opened 2017-02-13
adding weight norm container

huihuifan:
adding container, documentation, and tests (also tested on modules in fbnn). 

let me know if I should revert the whitespace changes in containers.md.


soumith:
thanks!

PR #1135, opened 2017-02-15
make testIO clone input and gO for inplace tests

ngimel:


Jacobian.lua
```diff
@@ -293,10 +293,12 @@ function nn.Jacobian.testIO(module,input, minval, maxval)
    minval = minval or -2
    maxval = maxval or 2
    local inrange = maxval - minval
+   inputclone = input:clone()
```
soumith: local

Jacobian.lua
```diff
@@ -293,10 +293,12 @@ function nn.Jacobian.testIO(module,input, minval, maxval)
...
 
    -- run module
    module:forward(input)
    local go = module.output:clone():copy(torch.rand(module.output:nElement()):mul(inrange):add(minval))
+   goclone = go:clone()
```
soumith: again needs local

soumith:
thanks

PR #1137, opened 2017-02-17
Using accreal instead of real in the API

pavanky:
- This reverts commit 7a07afe545b4deae5919d9dc268bfac3d37398c7.
- Includes fixes for TemporalRowConvolution

### DIFF
--- a/THNN.lua
+++ b/THNN.lua
@@ -73,6 +73,13 @@ local replacements =
     }
 }
 
+-- gsub(s, 'real', 'float') changes accreal to accfloat.
+-- typedef accfloat ahead of time.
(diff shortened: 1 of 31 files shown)


soumith:
you said it includes fixes for TemporalRowConvolution, but the unit tests are still massively failing. Any ideas on why?

pavanky:
@soumith The "fixes" just includes converting real to accreal for TemporalRowConvolution too. It was added after my original PR.

I am curious since you commented here and not cunn. Are the nn tests failing too ?

soumith:
let's try this again! thanks for the PR!

PR #1149, opened 2017-02-24
MixtureTable - variable length bugfix

FabbyD:
If batch size between two successive inputs of variable lengths was the same, self.size was not recomputed. I removed all batch size checks to make sure self.size is recomputed every time.


soumith:
thanks!

FabbyD:
I'm glad I could help 👍

PR #1150, opened 2017-02-28
THVector_(add) -> THVector_(adds)

soumith:
needed for https://github.com/torch/torch7/pull/944


fmassa:
So, you finally decided to rename the functions [the other way around](https://github.com/torch/torch7/pull/944#issuecomment-282520930)?

soumith:
adds -> addscalar. Best I could do.

fmassa:
Ok, I think I misunderstood the comment in the thread.
I thought you would be keeping the old function names (`THVector_add` for vector+vector) to avoid possible breakage in libs using TH.
But I'm ok with both 👍

PR #1152, opened 2017-03-01
Bottle handling gradInput = nil

joeyhng:
gradInput is sometimes nil for saving computation (like [this](https://github.com/facebook/fb.resnet.torch/blob/master/models/resnet.lua#L202)). This PR handles this case to avoid error in Bottle.


soumith:
thanks!

Issue #1160, opened 2017-03-06
nn.Contiguous corrupts the input tensor when it changes from contiguous to non-contiguous


richardassar:
I've a training script which alternates between training and sampling from a generative network. 

When sampling the network ends up passing a contiguous tensor through a `nn.Contiguous` layer and when training the input tensor ends up non-contiguous.

I encountered some difficulty with the network blowing up **only** when I sampled between training iterations. I tracked the problem down to `nn.Contiguous`, triggered only when the preceding layer alternates in this manner without a call to `Module:clearState`.

This line https://github.com/torch/nn/blob/master/Contiguous.lua#L7 sets `self.output` to the input tensor with a call to `set`, assigning `self.output`'s storage to `input` if the input tensor is contiguous, when the input tensor subsequently becomes contiguous it will then reach https://github.com/torch/nn/blob/master/Contiguous.lua#L5 and the calls to `resize` and `copy` will destroy the tensor.

The fix is to check if `self.output.storage() == input.storage()` and then call `self.output:set()` to break the link.

I'll submit a pull request including the fix and test case.

PR #1161, opened 2017-03-06
Fix and test case for nn.Contiguous

richardassar:
Fix for https://github.com/torch/nn/issues/1160


soumith:
thanks!

richardassar:
It was this project https://github.com/richardassar/SampleRNN_torch where I needed to alternate between training and sampling.

PR #1162, opened 2017-03-07
Adding LinearWeightNorm layer, unit test and updated documentation

richardassar:
Adding a specialised case of `nn.WeightNorm(nn.Linear(...))`, holds fewer tensors, occupies less memory and avoids repeatedly computing the weight matrix when unnecessary.

Added a unit test and updated docs.


LinearWeightNorm.lua
```diff
@@ -0,0 +1,220 @@
...
+    end   
+end
+
+function LinearWeightNorm:updateWeightMatrix()
+    if self.dirty or self.train then
+        if self.norm:dim() == 0 then self.norm:resizeAs(self.g) end
+        if self.scale:dim() == 0 then self.scale:resizeAs(self.g) end
+        if self.weight:dim() == 0 then self.weight:resizeAs(self.v) end
+
+        self.norm:norm(self.v,2,2):add(self.eps)
```
jdriesen0908: It's very possible that adding this eps is what causes the unittest to fail. You could try to replace with 
```lua
self.norm:norm(self.v,2,2):clamp(self.eps, math.huge)
```

LinearWeightNorm.lua
```diff
@@ -0,0 +1,220 @@
...
+            -- update the size of addBuffer if the input is not the same size as the one we had in last updateGradInput
+            updateAddBuffer(self, input)
+            self.gradBias:addmv(scale, gradOutput:t(), self.addBuffer)
+        end
+    end
+
+    local scale = self.scale:expandAs(self.v)
+    local norm = self.norm:expandAs(self.v)
+
+    self.weight:copy(self.gradV):cmul(self.v):cdiv(norm) -- reuse the weight tensor, avoid unnecessary alloc
```
jdriesen0908: No need for `:copy()` here (see above)

LinearWeightNorm.lua
```diff
@@ -0,0 +1,220 @@
...
+
+function LinearWeightNorm:updateWeightMatrix()
+    if self.dirty or self.train then
+        if self.norm:dim() == 0 then self.norm:resizeAs(self.g) end
+        if self.scale:dim() == 0 then self.scale:resizeAs(self.g) end
+        if self.weight:dim() == 0 then self.weight:resizeAs(self.v) end
+
+        self.norm:norm(self.v,2,2):add(self.eps)
+        self.scale:copy(self.g):cdiv(self.norm)
+        self.weight:copy(self.v):cmul(self.scale:expandAs(self.v))
```
jdriesen0908: These explicit copies are not necessary. You can just do
```lua
self.scale:cdiv(self.g, self.norm)
self.weight:cmul(self.v, self.scale:expandAs(self.v))
```

LinearWeightNorm.lua
```diff
@@ -0,0 +1,220 @@
...
+
+    local scale = self.scale:expandAs(self.v)
+    local norm = self.norm:expandAs(self.v)
+
+    self.weight:copy(self.gradV):cmul(self.v):cdiv(norm) -- reuse the weight tensor, avoid unnecessary alloc
+    self.gradG:sum(self.weight,2)
+    
+    self.gradV:cmul(scale)
+
+    self.weight:copy(self.v):cmul(scale):cdiv(norm)
```
jdriesen0908: No need for `:copy()` here (see above)

LinearWeightNorm.lua
```diff
@@ -0,0 +1,220 @@
...
+    self.weight:cmul(self.gradG:expandAs(self.weight))
+
+    self.gradV:add(-1,self.weight)
+
+    self.dirty = true
```
jdriesen0908: The way I understand the `dirty` flag is that if true it indicates that the rows of self.weight are not on the hyper-spheres with respective radiuses  self.g. Setting it to true here seems a bit premature since the module's state only becomes 'dirty' after a parameter update. 
I don't think you  really need this line. For users who manually update the parameters, i.e. without calling `updateParameters`, there is already a safety in place, since when in training mode, the module always behaves as though it's dirty (see line 100).

richardassar:
Hey Joris,

Thanks :)

I'll check out the eps issue. I do like the `clamp(eps,math.huge)` approach
better.

I noticed the tests failing but assumed since mine was passing that it was
unrelated to my changes, I'll look into it.

Removing the copy is definitely neater, I like it much better without.

You're right, that `self.dirty = true` is unnecesary in accGradParameters.

I'll make these edits soon.

Cheers,
Richard

nicholas-leonard:
@richardassar Are you still working on this?

richardassar:
Hi Nicholas. I've been quite busy but I can integrate these changes this evening.

Did you get my message about the weight-normalised version of SeqGRU/SeqLSTM/SeqLSTMP? These are located [here](https://github.com/richardassar/SampleRNN_torch).

richardassar:
If those modules are of interest I can submit a PR.

Also, `LinearWeightNorm` is usable as-is, the failing test seems to be a knock on effect of something else causing *other* tests to fail, the gradient check actually passes. 

The changes suggested by @jdriesen0908 don't seem to affect performance, but do neaten the code.

LinearWeightNorm.lua
richardassar: Strangely the test was passing when run individually, but when run as part of the entire suite other tests failed.

LinearWeightNorm.lua
richardassar: I don't agree with this last point. The module is dirty because `self.weight` is used as a temporary and does need to be recomputed at this point, even before the parameter updates. An alternative would be to use a separate temporary but at the cost of less memory efficiency.

richardassar:
@nicholas-leonard I've updated the module and the test works fine for me locally... any idea why it fails on only certain builds on travis-ci?

https://travis-ci.org/torch/nn/builds/231120571?utm_source=github_status&utm_medium=notification

nicholas-leonard:
@richardassar Fixed, rebased and merged here https://github.com/torch/nn/commit/65f42723f51d13683a431329452e5f385988e984 . Thanks for the PR!

richardassar:
Thanks!

PR #1164, opened 2017-03-07
Allow SelectTable to accept non-tensor table content.

kbullaughey:
At present, SelectTable assumes that the tables will contain only tensors. Sometimes it is convenient in NLP work to pass around tables that some tensors and some other other data, such as strings. This fix allows for that, albeit with the caveat that those components of the table will not back-progagate gradients.


nicholas-leonard:
@kbullaughey Thanks!

PR #1167, opened 2017-03-08
Fix possible math typo in MultiMarginCriterion doc

twitwi:
The `-` should mathematically be `+`, and it makes it coherent with the math block just before it.


soumith:
thanks!

PR #1171, opened 2017-03-10
MapTable: 'zeroGradParameters' and 'updateParameters' bugfix

davidemaz:
These two commits fix a bug in MapTable layer. 
zeroGradParameters only works when parameters are shared among layers. If they are not shared, the call to module:zeroGradParameters() is effective only on the first module. Instead, by calling the parent function, it is effective on the whole set of modules. The same applies for updateParameters.
Since managing a list of shared/not shared parameters is tricky and involves more lines of code, I opted for the simplest solution: a boolean variable that indicates if the whole set of parameters is shared or not. The default behavior remains the same as before.

### DIFF
--- a/MapTable.lua
+++ b/MapTable.lua
@@ -16,7 +16,7 @@ function MapTable:_extend(n)
          if shared then
            self.modules[i] = self.module:clone(table.unpack(self.sharedparams))
          else
-           self.modules[i] = self.module:clone(table.unpack(self.sharedparams))
+           self.modules[i] = self.module:clone()
          end
       end
    end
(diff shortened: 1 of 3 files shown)


MapTable.lua
```diff
@@ -2,7 +2,7 @@ local MapTable, parent = torch.class('nn.MapTable', 'nn.Container')
 
 function MapTable:__init(module, shared)
    parent.__init(self)
-   self.shared = shared or true
+   if shared ~= nil then self.shared = shared else self.shared = true end
```
pavanky: This can just be `shared = (shared == nil) and true or shared`

davidemaz: yes sure, this seems a more compact way to write it! I can change it.

PR #1181, opened 2017-03-24
Adding IndexLinear

pavanky:
This is a faster alternative to SparseLinear. This is achieved by
- using different layout for input
- using different layout for weight
- sparse weight updates 

This module also supports max normalization. The detailed description can be seen in the documentation.


pavanky:
Here is a  [benchmark](https://gist.github.com/pavanky/e899e213bee797df8ea66b9152e0590e#file-indexlinear_bench-lua) showing the performance improvements and here are the [results](https://gist.github.com/pavanky/e899e213bee797df8ea66b9152e0590e#file-results-log).

Hardware tested on:
- CPU: i7 3930K
- GPU: GTX Titan Z, CUDA 8.0

@soumith I did not find a place to put the benchmarks in nn. Do you mind if we created a new folder for benchmarks ?

soumith:
sure, put it under test/benchmarks

pavanky:
@soumith done. Let me know if there are any specific sizes you are interested in so I can include that in the comments.

PR #1183, opened 2017-03-28
Table input for nn.Bottle

danpol:
Now nn.Bottle can accept modules that require input and/or output to be a table of tensors.


nicholas-leonard:
@danpol can you add unit test and fix merge conflict?

PR #1186, opened 2017-03-29
Using temporary variables when performing transpose + addmm

### LINKED ISSUE #1185: Problems with parameter transposing and multiple threads
The following modules  have the following pattern during the forward step or backward step or both:

- Transpose weight in place
- perform addmm
- Transpose weight in place

```
ib/THNN/generic/Linear.c
lib/THNN/generic/TemporalConvolution.c
lib/THNN/generic/VolumetricConvolutionMM.c
lib/THNN/generic/SpatialConvolutionLocal.c
lib/THNN/generic/SpatialConvolutionMM.c
lib/THNN/generic/TemporalRowConvolution.c
```

------------------------

This is a problem when multiple threads are sharing the same parameter tensors. There are two solutions:

1) Try to use gemm directly with the transpose options.
2) Have a temporary tensor for storing the transpose. Since the transpose is just changing the metadata, it should not create any overhead.

I am in the process of creating a fix with (2). Let me know if (1) is the preferred solution.

pavanky:
- Fixes #1185

### DIFF
--- a/lib/THNN/generic/Linear.c
+++ b/lib/THNN/generic/Linear.c
@@ -42,9 +42,10 @@ void THNN_(Linear_updateOutput)(
       THTensor_(zero)(output);
     }
     THNN_(Linear_updateAddBuffer)(state,input,addBuffer);
-    THTensor_(transpose)(weight,weight,0,1);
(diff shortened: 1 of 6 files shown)


soumith:
thanks for the cleanup!

PR #1196, opened 2017-04-18
Add more contiguity checks to THNN

apaszke:


soumith:
as discussed, this is insufficient / wrong direction

PR #1197, opened 2017-04-19
nn.Transpose setNumInputDims() added

davidemaz:
This PR adds the setNumInputDims() function to nn.Transpose layer. The behaviour is the same as for other layers (e.g. nn.View). It allows the use of minibatches with nn.Transpose


soumith:
thanks!

PR #1200, opened 2017-04-19
Add more newContiguous calls and checks

apaszke:


fmassa:
Hey, I'm a bit late here, but I think that there are some problems with this PR.
I think we should use `freeCopyTo` when freeing the weights (if they were not contiguous), if not the weights won't be updated in the non-contiguous case.
`freeCopyTo` is extensively used in THC, for example in https://github.com/torch/cutorch/blob/master/lib/THC/generic/THCTensorMasked.cu
(Typing from the phone, sorry for the lack of details)

soumith:
freeCopyTo is only needed if we updated / modified the weights. But none of these functions actually modify the weights.

apaszke:
I only added some asserts for gradWeight which is modified + contiguous calls for read-only weights.

fmassa:
Yeah, my bad, I should have looked more closely. Sorry for the noise

PR #1204, opened 2017-05-03
Added sharedparams default values in MapTable so that it is backwards…

paucarre:
Models Serialized before commit `179133fb499483555074eaff2ea875e975508df1` no longer work as the `sharedparams` is null. 
This pull requests makes MapTable compatible with already serialized models.


soumith:
thanks!

PR #1205, opened 2017-05-04
WeightNorm: don't update weights when not in training mode

jgehring:
In evaluation mode, it's not necessary to re-compute the child module's weights
at every forward pass. Instead, do it once when switching from training to
evaluation mode. This speeds up inference.


soumith:
thanks!

PR #1208, opened 2017-05-04
Fix WeightNorm serialization for permutated weight matrices

jgehring:
For layers where the first weight dimension does not correspond to the output
dimension, self.viewOut will *not* correspond to self.weight:size(). This is
fixed by introducing another member variable that simply holds the original size
of the weight matrix.

### DIFF
--- a/WeightNorm.lua
+++ b/WeightNorm.lua
@@ -34,6 +34,7 @@ function WeightNorm:__init(module, outputDim)
 
     -- view size back to original weight
     self.viewOut = self.weight:size()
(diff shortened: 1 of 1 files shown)


jgehring:
Not sure why the CI build failed. `jacTest` fails on TemporalRowConvolution, which is tested directly after WeightNorm. The failure happens for GCC+LUAJIT21 only.

soumith:
it's a flaky module / test, no one's figured why it's like that

jgehring:
Ok thanks!

PR #1220, opened 2017-05-24
Fix backward compatibility with serialized nn.Sum modules

guillaumekln:
44a15a1 introduced the `squeeze` option for nn.Sum but did not correctly handle the case of serialized nn.Sum modules in which undefined `self.squeeze` means `self.squeeze = true`.


soumith:
thanks!

PR #1245, opened 2017-06-19
recursive functions use resize instead of resizeAs

nicholas-leonard:


utils.lua
```diff
@@ -130,15 +130,20 @@ function nn.utils.recursiveAdd(t1, val, t2)
...
       for key,_ in pairs(t2) do
-         t1[key], t2[key] = nn.utils.recursiveCopy(t1[key], t2[key])
+         t1[key], t2[key] = nn.utils.recursiveCopy(t1[key], t2[key], async)
       end
    elseif torch.isTensor(t2) then
       t1 = torch.isTensor(t1) and t1 or t2.new()
-      t1:resizeAs(t2):copy(t2)
+      t1:resize(t2:size())
+      if async then
+        t1:copyAsync(t2)
```
pavanky: May be we should check if `t1` or `t2` are cuda tensors. I don't think `copyAsync` is supported for regular tensors.

nicholas-leonard: copyAsync already checks this and generates an error.

PR #1251, opened 2017-06-23
Fix segfault in SpatialDepthWiseConvolution w/o bias

lospooky:
Calling :noBias() on nn.SpatialDepthWiseConvolution results in segfaults during both forward and backward passes. PR fixes the issue.


soumith:
thank you!

PR #1259, opened 2017-07-12
CPU implementation of L_p feature pooling

wickedfoo:
This is a naive CPU implementation of L_p feature pooling to correspond with
https://github.com/torch/cunn/pull/477

includes the original test from fbcunn that it passes.


wickedfoo:
ping... can this be merged?

Issue #1262, opened 2017-07-17
Volumetric Upsampling Modules Release Date


Dimimpel:
We’re looking to use volumetric upsampling in our network. Some files, e.g. VolumetricUpSamplingNearest.c have appeared in the torch repository last month, related it seems to a pytorch upsampling implementation. The python wrappers for these files were commited to pytorch too, but nothing for LUA here. I was wondering if LUA wrappers for these files are on the roadmap (and thus making them available when someone downloads Torch) and also roughly when is this going to happen in order to decide if it would be best for now to implement these ourselves ?

soumith:
we have no plans yet of writing Lua wrappers. It's best for now to implement these yourselves.

Dimimpel:
Thanks for the immediate answer, however when I tried to insert the wrapper myself, even if everything were compiled, the VolumetricUpSamplingNearest module didn't appear in Torch with the other nn modules.
Namely, I downloaded a Torch, installed the dependencies, inserted the wrapper (attached file),  wrote the following line in the init.lua file: require('nn.VolumetricUpSamplingNearest') and then executed the install.sh (./install.sh).
Everything were compiled but as you can see in the attached image when I entered torch and I typed
require('nn'); and then nn.(tab tab) the VolumetricUpSamplingNearest module wasn't displayed.

![screenshot from 2017-07-17 17-08-26](https://user-images.githubusercontent.com/9349855/28279431-5df3412e-6b18-11e7-9619-64ea8ba2b677.png)

[VolumetricUpSamplingNearest.zip](https://github.com/torch/nn/files/1153360/VolumetricUpSamplingNearest.zip)

soumith:
you'll also have to modify init.lua

Dimimpel:
As I said in the previous message
 **wrote the following line in the init.lua file: require('nn.VolumetricUpSamplingNearest')**
So it isn't that.

mikepound:
Thanks Soumith, we've sorted this now - apparently was an luarocks caching issue. The wrapper works fine, if we add the necessary test and documentation is this something you'd be interested in merging? Alternatively we could probably combine both into an upsample layer much like exists in pytorch, with a mode parameter.

soumith:
i'll merge either of the approaches you do. Thanks for writing the wrappers.

soumith:
and in this context, if you have interest in porting your workflow to pytorch, and you have questions you'd want to ask, let me know.

PR #1264, opened 2017-07-21
Added UpSampling module and associated tests.

mikepound:
From issue #1262, a more generic UpSampling module that provides a wrapper to the recently added volumetric upsampling code. Behaves similarly to  the pytorch upsample class, in that it infers 2D or 3D upsampling based on the input size, and can operate in either nearest neighbor or linear mode.

- This module is a little thematically different to the existing upsampling modules, in that it works on either spatial or volumetric depending on the input.
- There is cunn testing code, but I didn't include it as this is an nn commit. I wasn't sure if I could push to two repos at once.

### DIFF
--- a/UpSampling.lua
+++ b/UpSampling.lua
@@ -1,0 +1,216 @@
+require 'nn.THNN'
+local UpSampling, parent =
+   torch.class('nn.UpSampling', 'nn.Module')
+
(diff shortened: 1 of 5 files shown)


soumith:
thanks Mike!

PR #1267, opened 2017-07-27
replace long long with size_t

elikosan:
The long long construct causes the msvc compiler to crash when compiling in 32bit mode. The long long is not necessary anyway in this case since it is cast-ed later to size_t.


soumith:
thanks Eric!

PR #1271, opened 2017-08-03
BN supports batchsize=1

nicholas-leonard:
This fixes a bug where samples of batchsize 1 would introduce a NaN in BatchNormalization.running_var (see C code where `accreal unbiased_var = sum / (n - 1);`). 
Of course, ideally, we wouldn't need to support a batchsize (that is, n) of 1.
But while iterating a dataset of, say, 101 samples in batch sizes of 10 or less, the last batch will have one sample. This use case break BatchNormalization by introducing a NaN in the running_var.

### DIFF
--- a/BatchNormalization.lua
+++ b/BatchNormalization.lua
@@ -130,7 +130,7 @@ function BN:updateOutput(input)
       self.running_var:cdata(),
       self.save_mean:cdata(),
       self.save_std:cdata(),
-      self.train,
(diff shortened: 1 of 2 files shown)


pavanky:
looks good :+1:

yzhuang:
Looking at https://github.com/torch/nn/blob/master/lib/THNN/generic/BatchNormalization.c

It seems like with this fix, we would not update the running_mean and running_var at all when using a batch_size of 1.  This behavior seems broken right? (Also momentum is just ignored since we don't update the running mean and var.)

PR #1277, opened 2017-09-05
1D Bias Vector in SpatialDepthwiseConvolution

killeent:
As title. The end goal is to support passing a 1D Bias Tensor in PyTorch to the backing C code. Basically we just add some resizing to the existing code, and rename nOutputPlane to depthwiseMultiplier where appropriate.

### DIFF
--- a/SpatialDepthWiseConvolution.lua
+++ b/SpatialDepthWiseConvolution.lua
@@ -1,14 +1,17 @@
 local THNN = require 'nn.THNN'
 local SpatialDepthWiseConvolution, parent = torch.class('nn.SpatialDepthWiseConvolution', 'nn.Module')
 
-function SpatialDepthWiseConvolution:__init(nInputPlane, nOutputPlane, kW, kH, dW, dH, padW, padH)
+SpatialDepthWiseConvolution.__version = 2
+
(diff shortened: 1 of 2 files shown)


SpatialDepthWiseConvolution.lua
```diff
@@ -118,7 +127,7 @@ end
 
 function SpatialDepthWiseConvolution:__tostring__()
    local s = string.format('%s(%d -> %d, %dx%d', torch.type(self),
-         self.nInputPlane, self.nOutputPlane, self.kW, self.kH)
+         self.nInputPlane, self.depthwiseMultiplier, self.kW, self.kH)
```
soumith: add a :read method to this file similar to:
https://github.com/torch/nn/blob/master/BatchNormalization.lua#L184-L192

Add a `__version = 2` similar to https://github.com/torch/nn/blob/master/BatchNormalization.lua#L33

Then, in the `:read`, add:
```
   parent.read(self, file)
   if version < 2 then
       self.depthwiseMultiplier = self.nOutputPlane
       self.nOutputPlane = self.nInputPlane * self.depthwiseMultiplier;
   end
```

soumith (changes requested):

killeent:
Closing in favor of https://github.com/pytorch/pytorch/pull/3057

PR #1283, opened 2017-10-02
fix condition logic for coef recalculation in SpatialSubtractiveNormalization

elikosan:
Existing conditional expression always verified, resulting in overhead as coefficients were recalculated every time. Fix and simplify expression by using cached input size.


soumith:
this will fail if input size changes right? that's not correct.

elikosan:
If the input size changes, then it will recompute the coefs.
The expression: not input:isSize(self._inpsz) becomes true when the size of input and the cached size are different.

soumith:
thanks!

elikosan:
Hi Soumith,

I found out why Travis failed.
I need to clear my cache in SpatialSubtractiveNormalization:clearState()
Otherwise the jac.TestIO fails

I will make another pull request with the fix.
Sorry about that.

Eric.

